PR #22, opened 2020-09-08
[Ready for Merge] Add ubuntu16.04 into build task

ShengguangXiao:
Add Ubuntu16.04 to the develop and release build work flows, so it will be very easy to support users who is using Ubuntu 16.04.


ShengguangXiao:
This PR used wrong approach to support all Ubuntu Distro

PR #39, opened 2020-09-18
Accept both old and new magic values

monstrobishi:
This PR makes the magic header changes more robust by allowing nodes to accept both old and new magic headers


ShengguangXiao:
Should we also change for the message sending part?

monstrobishi:
@ShengguangXiao this PR is only to change receiving part, sending will stay the same

PR #45, opened 2020-09-22
Implemented utxo check

monstrobishi:
This introduces a restriction to AccountToUTXO to prevent non-DFI tokens from being transformed into UTXOs

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -445,6 +445,15 @@ Res ApplyAccountToUtxosTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
     if (msg.balances != *minted.val) {
         return Res::Err("%s: amount of minted tokens in UTXOs and metadata do not match: (%s) != (%s)", base, minted.val->ToString(), msg.balances.ToString());
     }
+
+    // block for non-DFI transactions
+    for (auto const & kv : minted.balances) {
+        DCT_ID tokenId = kv.first;
+        if(tokenId == DCT_ID{0}) {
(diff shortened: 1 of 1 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -445,6 +445,15 @@ Res ApplyAccountToUtxosTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
     if (msg.balances != *minted.val) {
         return Res::Err("%s: amount of minted tokens in UTXOs and metadata do not match: (%s) != (%s)", base, minted.val->ToString(), msg.balances.ToString());
     }
+
+    // block for non-DFI transactions
+    for (auto const & kv : minted.balances) {
+        DCT_ID tokenId = kv.first;
```
ShengguangXiao, line 451: Can use `const DCT_ID& tokenId = kv.first;` To avoid extra call of assign operator

src/masternodes/mn_checks.cpp
```diff
@@ -445,6 +445,15 @@ Res ApplyAccountToUtxosTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
...
+        if(tokenId != DCT_ID{0}) {
```
ShengguangXiao, line 452: Add blank between if and (

ShengguangXiao (approved):

uzyn:
Looks good. Does this also block on the interface side automatically? @monstrobishi

monstrobishi:
@uzyn yes it will fail the dry run execution test in the RPC call and return an error message

uzyn (approved):
Thanks for the speedy work @monstrobishi

PR #46, opened 2020-09-28
Change source code and website to defichain

ShengguangXiao:
When run defie —version, it is still displaying bitcoin information, now change to defichain website and github repo


monstrobishi:
Small and simple PR, merging in

Issue #54, opened 2020-10-01
getaccount API third parameter indexed_amounts always report error 


ShengguangXiao:
Below API call already report parameter is not a valid json value.

```
./src/defi-cli -testnet getaccount 76a91417c6c6844421f0bd738cd2971305852fc82b03c988ac "{}" true
```

monstrobishi:
@saurabh391 you managed to get this working right

shengguang-blockone:
Don't input the third parameter will work

ShengguangXiao:
Don't input the third parameter will work.

ShengguangXiao:
fixed in PR https://github.com/DeFiCh/ain/pull/55

PR #55, opened 2020-10-02
Fix getaccount API third parameter not working

ShengguangXiao:
Fix issue https://github.com/DeFiCh/ain/issues/54


monstrobishi:
Looks good, but any clue why travis is failing @ShengguangXiao

PR #58, opened 2020-10-07
Minted implementation in the gettoken

DeFiDev:
Added minted field for gettoken rpc


monstrobishi:
This looks good merging

PR #59, opened 2020-10-08
Refactored "AMK heights" validation

DeFiDev:
- all `if (height < AMK)` clauses moved from "outer Apply switch" and from RPCs into inner Apply*s. It is less readable than "outer switch", but it is a "single point of validation" now.
- all heights in RPCs adjusted to "current height+1"
- all accesses to chainstate (I mean those `*.Tip()->height" or ".Height()` in tx versions etc) SHOULD be protected by cs_main lock!

After acceptance we can promote those changes into Stage#4 for the rest txs/rpcs. To minimize conflicts - we have not touched Stage#4 yet.


monstrobishi (approved):

monstrobishi:
This looks good, merging

PR #60, opened 2020-10-09
Added gettokenbalances

DeFiDev:
* Added `gettokenbalances` which returns the sums of balances of all accounts belongs to the wallet
* For `listaccounts`, the `is_mine_only` parameter has been added, which returns balances of all accounts belongs to the wallet


monstrobishi (approved):
This looks good, merging it in

PR #62, opened 2020-10-14
Deferred masternode operator checks for minter thread

DeFiDev:
* Deferred masternode operator checks for minter thread


uzyn:
@DeFiDev @monstrobishi 

This is breaking our default behavior – where new coins are minted to owner and not operator.

1. What's the motivation behind this PR? 
2. This breaks the current behavior. We should have a unit test for it.

PR #63, opened 2020-10-14
Replace bishan to bayfront

DeFiDev:
* Replace bishan to bayfront


monstrobishi (approved):
This looks good no mention of Bishan anywhere

PR #65, opened 2020-10-14
Fix liquidity pool API parameter bug and help message

ShengguangXiao:
1. Make liquidity pool API empty parameter "[]" optional.
2. Fix `addpoolliquidity` and `removepoolliquidity` cannot run from defi-cli problem, `addpoolliquidity` the first parameter need to convert to json object, `removepoolliquidity` the second parameter no need to convert.
3. Update the API help messages.


monstrobishi (approved):
Looks good, merging in

PR #70, opened 2020-10-20
Fix call getgov API from defi-cli always fail

ShengguangXiao:
Fix call getgov API from defi-cli always fail


monstrobishi:
Looks good, merging in

PR #75, opened 2020-10-26
Fix liquidity check bug

ShengguangXiao:
Fix check liquidity pool problem when do pool swap.


ShengguangXiao:
I think now the reserve amount need at least 1000 may be too large, because 1000 is very big amount asset for bitcoin. We can add another constant like minimum reserve to swap, this value can be defined lower than 1000.

PR #76, opened 2020-10-27
Fix/poolswap restriction

DeFiDev:
Restrict initial swappable amount in tx to the pool reserved funds.

**Consensus breaking, if liquidity trade was deployed before!!!**

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -943,7 +943,13 @@ Res ApplyPoolSwapTx(CCustomCSView &mnview, const CCoinsViewCache &coins, const C
         return Res::Err("%s: can't find the poolpair!", base);
     }
 
+    /// @attention Consensus breaking, if liquidity trade was deployed!!!
     CPoolPair pp = poolPair->second;
+    const auto srcReserve = pp.idTokenA == poolSwapMsg.idTokenFrom ? pp.reserveA : pp.reserveB;
+    if (poolSwapMsg.amountFrom > srcReserve) {
+        return Res::Err("%s: %s", base, "Tradeble amount > than reserved in the pool");
+    }
+
     const auto res = pp.Swap({poolSwapMsg.idTokenFrom, poolSwapMsg.amountFrom}, poolSwapMsg.maxPrice, [&] (const CTokenAmount &tokenAmount) {
         auto resPP = mnview.SetPoolPair(poolPair->first, pp);
         if (!resPP.ok) {
--- a/test/functional/feature_poolswap.py
+++ b/test/functional/feature_poolswap.py
@@ -161,6 +161,19 @@ class PoolPairTest (DefiTestFramework):
         self.nodes[0].updatepoolpair({"pool": "GS", "status": True})
         self.nodes[0].generate(1)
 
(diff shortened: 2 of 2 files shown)


uzyn:
@DeFiDev you can change the target yourself (click "Edit")

src/masternodes/mn_checks.cpp
```diff
@@ -943,7 +943,13 @@ Res ApplyPoolSwapTx(CCustomCSView &mnview, const CCoinsViewCache &coins, const C
         return Res::Err("%s: can't find the poolpair!", base);
     }
 
+    /// @attention Consensus breaking, if liquidity trade was deployed!!!
     CPoolPair pp = poolPair->second;
+    auto srcReserve = pp.idTokenA == poolSwapMsg.idTokenFrom ? pp.reserveA : pp.reserveB;
```
ShengguangXiao, line 948: Add const in front

ShengguangXiao (approved):

src/masternodes/mn_checks.cpp
```diff
@@ -943,7 +943,13 @@ Res ApplyPoolSwapTx(CCustomCSView &mnview, const CCoinsViewCache &coins, const C
         return Res::Err("%s: can't find the poolpair!", base);
     }
 
+    /// @attention Consensus breaking, if liquidity trade was deployed!!!
     CPoolPair pp = poolPair->second;
+    const auto srcReserve = pp.idTokenA == poolSwapMsg.idTokenFrom ? pp.reserveA : pp.reserveB;
+    if (poolSwapMsg.amountFrom > srcReserve) {
```
monstrobishi, line 949: Why are we comparing amountFrom with reserveTokenFrom? The user will always put in the same from amount as they are willing to trade? I don't see what we're guarding against here?

monstrobishi: If user wants to trade 100 tokenA (as amount/token from), then 100 from the user's account will go into the pool first already.

DeFiDev: @monstrobishi  amountFrom it's amount which user send, not amount which user want to get. So that's why it's comparing with reserves - he set amount to exchange and it must be less than reserves.

uzyn: Exactly what @monstrobishi is saying, `amountFrom` can exceed reserve, correct? Don't see a point for this. 

@monstrobishi and myself feel that this PR is unnecessary, can you confirm @DeFiDev ? If you think it is needed, can you clarify further?

DeFiDev: This PR is not necessary for math correctness itself (really, `amountFrom` may exceed `reserveFrom` and all the math will be done correctly except of user's "disappointed expectations"). But this PR will solve, at least, the secondary problem - CPU load by very long loops in the case when unswapped amount _vastly_ exceeds initial reserve (ex. 1000000 COINs to swap vs 0.0001 COIN reserved). With this PR the real loop will be no more than 1000 iterations for the worst case.

monstrobishi:
This breaks consensus and also is not ideal behaviour, closing for now, will reopen if we see the need

PR #78, opened 2020-10-28
Added to minting test multiply tokens minting

DeFiDev:
Added test for mint multiple tokens in one transaction.


ShengguangXiao:
If mint multiple tokens together, and these tokens belong to the same address, it will have below error. I think it added the same vin multile times in line `rawTx.vin.insert(rawTx.vin.end(), auths.begin(), auths.end());`, `addpoolliquidity` should also have the same problem.

```
error code: -26
error message:
bad-txns-inputs-duplicate (code 16)
```

monstrobishi:
I think more investigation is needed, please observe @ShengguangXiao above comments, @DeFiDev

monstrobishi:
I will merge this anyway as it's better to have more unit tests, but the problem still exists

PR #79, opened 2020-10-30
Improve slope calculation efficiency

uzyn:
Improve slope calculation efficiency to O(1).

TODOs:
- [x] Remove `SLOPE_SWAP_RATE`.


uzyn:
Removed `SLOPE_SWAP_RATE` for calculation, leaving the rest in there for now as there's no harm.

ShengguangXiao (approved):

uzyn:
@ShengguangXiao can you take a look at the failed `wallet_balance.py` test and fix it if possible? Not sure why this would break wallet_balance functional test. It seems unrelated.

ShengguangXiao:
> @ShengguangXiao can you take a look at the failed `wallet_balance.py` test and fix it if possible? Not sure why this would break wallet_balance functional test. It seems unrelated.

OK, let me take a look.

ShengguangXiao:
The build is good, I run test locally is pass and trigger build on Travis again it passed too.

uzyn:
Thanks @ShengguangXiao 

@monstrobishi I guess we can target 1.4.x for this fix?

DeFiDev:
> @ShengguangXiao can you take a look at the failed `wallet_balance.py` test and fix it if possible? Not sure why this would break wallet_balance functional test. It seems unrelated.

`wallet_balance` (and one more, if I'm not mistaken, `wallet_backup`) are the very old (ancient) "bugs". I tried to fix them many times, but they bangs only on travis and never locally. I think they originates since native bitcoin code. Something wrong with sync/delays or so in the test code itself, but I can't catch it. I even wanted to exclude them from pack, but they are quite important and could help not to miss smth really critical later.

bvbfan:
Can we use better calculating since `swapped = poolT * (1 - (poolF / (poolF + unswappedA)))` denominator is higher which means `poolF / (poolF + unswappedA)` is floating point.

ShengguangXiao:
> Can we use better calculating since `swapped = poolT * (1 - (poolF / (poolF + unswappedA)))` denominator is higher which means `poolF / (poolF + unswappedA)` is floating point.

poolF + unswappedA is higher does not indicate the result is floating point, if all types are integer, ` poolF / (poolF + unswappedA)` will always be 0.

bvbfan:
Correct. It's floating point if it cast to float/double, it's 0 instead. So the calculation is not correct since unswappedA is not 0.

uzyn:
> Can we use better calculating since `swapped = poolT * (1 - (poolF / (poolF + unswappedA)))` denominator is higher which means `poolF / (poolF + unswappedA)` is floating point.

@bvbfan this runs into int division. I've in fact made this earlier on: https://github.com/DeFiCh/ain/pull/79/commits/d2b8d42bfaad5044cf96f1374ba69b525066df27 before it's fixed to this one.

Key to division is to make numerator as huge as possible before dividing.

bvbfan:
> > Can we use better calculating since `swapped = poolT * (1 - (poolF / (poolF + unswappedA)))` denominator is higher which means `poolF / (poolF + unswappedA)` is floating point.
> 
> @bvbfan this runs into int division. I've in fact made this earlier on: [d2b8d42](https://github.com/DeFiCh/ain/commit/d2b8d42bfaad5044cf96f1374ba69b525066df27) before it's fixed to this one.
> 
> Key to division is to make numerator as huge as possible before dividing.

Yes, that's correct calculation

bvbfan (approved):

PR #80, opened 2020-10-30
Isappliedcustomtx

DeFiDev:
Added isappliedcustomtx RPC method which indicates was custom transaction applied on chain or not.


monstrobishi:
@DeFiDev has conflicts with base branch now, and for some reason Macosx build has failed, can you investigate thanks (Not urgent)

PR #84, opened 2020-10-31
Fix tx error if mint tokens of same owner together

ShengguangXiao:
Fix the comments in PR https://github.com/DeFiCh/ain/pull/78

If mint multiple tokens together, and these tokens belong to the same address, it will have below error. I think it added the same vin multile times in line `rawTx.vin.insert(rawTx.vin.end(), auths.begin(), auths.end());`, `addpoolliquidity` should also have the same problem.

```
error code: -26
error message:
bad-txns-inputs-duplicate (code 16)
```

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -1143,7 +1143,7 @@ UniValue minttokens(const JSONRPCRequest& request) {
                 }
 
                 for (const auto& auth : auths) {
-                    if (std::find_if(rawTx.vin.begin(), rawTx.vin.end(), [&auth](const CTxIn& txIn) { return txIn == auth; }) == rawTx.vin.end()) {
+                    if (std::find(rawTx.vin.begin(), rawTx.vin.end(), auth) == rawTx.vin.end()) {
                         rawTx.vin.push_back(auth);
                     }
                 }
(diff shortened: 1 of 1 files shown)


monstrobishi (approved):
Nice, this doesn't break consensus

src/masternodes/mn_rpc.cpp
```diff
@@ -1141,7 +1141,12 @@ UniValue minttokens(const JSONRPCRequest& request) {
                         throw JSONRPCError(RPC_INVALID_REQUEST, "Incorrect Authorization");
                     gotFoundersAuth = true; // we need only one any utxo from founder
                 }
-                rawTx.vin.insert(rawTx.vin.end(), auths.begin(), auths.end());
+
+                for (const auto& auth : auths) {
+                    if (std::find_if(rawTx.vin.begin(), rawTx.vin.end(), [&auth](const CTxIn& txIn) { return txIn == auth; }) == rawTx.vin.end()) {
```
DeFiDev, line 1146: Why "find_if" with predicate here instead of just "find"??
   `if (std::find(rawTx.vin.begin(), rawTx.vin.end(), auth) == rawTx.vin.end()) {`
...

src/masternodes/mn_rpc.cpp
```diff
@@ -1737,8 +1742,13 @@ UniValue addpoolliquidity(const JSONRPCRequest& request) {
             if (!ExtractDestination(kv.first, ownerDest)) {
                 throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid owner destination");
             }
-            std::vector<CTxIn> rawIn = GetAuthInputs(pwallet, ownerDest, UniValue(UniValue::VARR));
-            rawTx.vin.insert(rawTx.vin.end(), rawIn.begin(), rawIn.end());
+
+            std::vector<CTxIn> auths = GetAuthInputs(pwallet, ownerDest, UniValue(UniValue::VARR));
+            for (const auto& auth : auths) {
+                if (std::find_if(rawTx.vin.begin(), rawTx.vin.end(), [&auth](const CTxIn& txIn) { return txIn == auth; }) == rawTx.vin.end()) {
```
DeFiDev, line 1748: same here

PR #88, opened 2020-11-03
Fix try to convert string to json in defi-cli bug

ShengguangXiao:
Fix try to convert string to json in defi-cli bug

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -242,7 +242,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
 
     { "setgov", 0, "variables" },
     { "setgov", 1, "inputs" },
-    { "isappliedcustomtx", 0, "txid" },
+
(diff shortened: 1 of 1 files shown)


src/rpc/client.cpp
```diff
@@ -242,7 +242,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
 
     { "setgov", 0, "variables" },
     { "setgov", 1, "inputs" },
-    { "isappliedcustomtx", 0, "txid" },
+
     { "isappliedcustomtx", 1, "blockHeight" },
```
monstrobishi, line 246: How about the blockheight, or do we need the conversion for string to number?

ShengguangXiao: Yes，so need to keep the block height here.

DeFiDev: yes, (1) should stay, (0) may be killed

DeFiDev (approved):

PR #89, opened 2020-11-03
Advancing `master` to v1.3.x

uzyn:
Now that 1.3.x is live.


monstrobishi:
I suggest we merge this once a v1.4.x branch is created, and we are somewhat certain there are no more incremental non-breaking updates to v1.3x

src/test/liquidity_tests.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        printf("Distributed: = %ld\n", distributed);
+        BOOST_CHECK(distributed == 9999000000); // always slightly less due to MINIMUM_LIQUIDITY & rounding
+
+        // check it
+        auto rwd25 = 25 * COIN / ProvidersCount;
+        auto rwd50 = 50 * COIN / ProvidersCount;
+        cache.ForEachPoolShare([&] (DCT_ID const & id, CScript const & owner) {
+//            if (id == RWD25/* || id == RWD50*/)
+//                printf("owner = %s: %s,\n", owner.GetHex().c_str(), cache.GetBalance(owner, DCT_ID{0}).ToString().c_str());
```
ShengguangXiao, line 338: Should we remove debugging code?

src/test/liquidity_tests.cpp
```diff
@@ -0,0 +1,377 @@
...
+    for (int i = 0; i < PoolCount; ++i) {
+        DCT_ID idA, idB, idPool;
+        std::tie(idA, idB, idPool) = CreatePoolNTokens(mnview, "A"+std::to_string(i), "B"+std::to_string(i));
+        auto optPool = mnview.GetPoolPair(idPool);
+        BOOST_REQUIRE(optPool);
+
+    }
+    // create shares
+    mnview.ForEachPoolPair([&] (DCT_ID const & idPool, CPoolPair const & pool) {
+//            printf("pool id = %s\n", idPool.ToString().c_str());
```
ShengguangXiao, line 282: Remove debug code?

ShengguangXiao (approved):

src/test/liquidity_tests.cpp
DeFiDev: Sure.. If you think it will not be needed anymore ^)

uzyn:
> I suggest we merge this once a v1.4.x branch is created, and we are somewhat certain there are no more incremental non-breaking updates to v1.3.x

I would suggest the following workflow:

- Current latest stable version is always `master`.
- Old `master` to be retired into `v1.2.x` and clearly labelled so in README or in the code.
- Delete the redundant `v1.3.x`.
- Immediately start the next version if it has not yet been established. Just to be clear that the next 1 or 2 versions can be started ahead of time.
- Our list of major branches should be:

    - v1.2.x (past, for security fixes, if required)
    - master (retiring v1.3.x)
    - v1.4.x (next major release)

The above workflow would make it consistent and easy for outside contributor, not understanding our internal workflow to make a contribution. Most contributions would be made towards `master`. 

Thoughts? @monstrobishi @bvbfan @ShengguangXiao

src/test/liquidity_tests.cpp
uzyn: Yes. But in a separate PR from this one.

src/test/liquidity_tests.cpp
uzyn: @DeFiDev 
> Sure.. If you think it will not be needed anymore ^)

Debug code should not be committed, esp. for stable version.

monstrobishi:
Merging

PR #90, opened 2020-11-03
Clean up nasty layered escaping in examples

uzyn:
Minor documentation changes.

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -199,8 +199,8 @@ UniValue createmasternode(const JSONRPCRequest& request) {
                        "\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"
                },
                RPCExamples{
-                   HelpExampleCli("createmasternode", "ownerAddress operatorAddress \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
-                   + HelpExampleRpc("createmasternode", "ownerAddress operatorAddress \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
+                   HelpExampleCli("createmasternode", "ownerAddress operatorAddress '[{\"txid\":\"id\",\"vout\":0}]'")
+                   + HelpExampleRpc("createmasternode", "ownerAddress operatorAddress '[{\"txid\":\"id\",\"vout\":0}]'")
                },
     }.Check(request);
 
@@ -288,8 +288,8 @@ UniValue resignmasternode(const JSONRPCRequest& request) {
                        "\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"
                },
                RPCExamples{
-                   HelpExampleCli("resignmasternode", "mn_id \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
-                   + HelpExampleRpc("resignmasternode", "mn_id \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
+                   HelpExampleCli("resignmasternode", "mn_id '[{\"txid\":\"id\",\"vout\":0}]'")
+                   + HelpExampleRpc("resignmasternode", "mn_id '[{\"txid\":\"id\",\"vout\":0}]'")
                },
     }.Check(request);
 
@@ -394,8 +394,8 @@ UniValue listmasternodes(const JSONRPCRequest& request) {
                        "{id:{...},...}     (array) Json object with masternodes information\n"
                },
                RPCExamples{
-                       HelpExampleCli("listmasternodes", "\"[mn_id]\" False")
-                       + HelpExampleRpc("listmasternodes", "\"[mn_id]\" False")
+                       HelpExampleCli("listmasternodes", "'[mn_id]' False")
+                       + HelpExampleRpc("listmasternodes", "'[mn_id]' False")
                },
     }.Check(request);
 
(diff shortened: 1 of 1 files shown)


src/masternodes/mn_rpc.cpp
```diff
@@ -286,8 +286,8 @@ UniValue resignmasternode(const JSONRPCRequest& request) {
                        "\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"
                },
                RPCExamples{
-                   HelpExampleCli("resignmasternode", "mn_id \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
-                   + HelpExampleRpc("resignmasternode", "mn_id \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
+                   HelpExampleCli("resignmasternode", "mn_id \"[{\"txid\":\"id\",\"vout\":0}]\"")
```
monstrobishi, line 289: I think this needs the single quotes added

src/masternodes/mn_rpc.cpp
```diff
@@ -729,10 +729,10 @@ UniValue updatetoken(const JSONRPCRequest& request) {
                        "\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"
                },
                RPCExamples{
-                       HelpExampleCli("updatetoken", "\"token {\\\"isDAT\\\":true}\" "
-                                                     "\"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
-                       + HelpExampleRpc("updatetoken", "\"token {\\\"isDAT\\\":true}\" "
-                                                       "\"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
+                       HelpExampleCli("updatetoken", "\"token {\"isDAT\":true}' "
```
monstrobishi, line 732: Single quotes needed here?

src/masternodes/mn_rpc.cpp
```diff
@@ -1062,8 +1062,8 @@ UniValue minttokens(const JSONRPCRequest& request) {
                RPCExamples{
                        HelpExampleCli("minttokens", "\"10@symbol\"")
                        + HelpExampleCli("minttokens",
-                                      "\"10@symbol\" \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
-                       + HelpExampleRpc("minttokens", "\"10@symbol\" \"[{\\\"txid\\\":\\\"id\\\",\\\"vout\\\":0}]\"")
+                                      "\"10@symbol\" \"[{\"txid\":\"id\",\"vout\":0}]\"")
```
monstrobishi, line 1065: Single quotes needed here?

src/masternodes/mn_rpc.cpp
```diff
@@ -2084,14 +2084,14 @@ UniValue accounttoutxos(const JSONRPCRequest& request) {
...
+                    },
                 },
                 RPCResult{
                        "\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"
                 },
                RPCExamples{
-                       HelpExampleCli("accounttoutxos", "sender_address \\\"{\\\"address1\\\":\\\"100@DFI\"}\\\" [] ")
-                       + HelpExampleCli("accounttoutxos", "sender_address \"{\\\"address1\\\":\\\"1.0@DFI\\\",\\\"address2\\\":[\\\"2.0@BTC\\\", \\\"3.0@ETH\\\"]}\" []")
+                       HelpExampleCli("accounttoutxos", "sender_address \"{\"address1\":\"100@DFI\"}' [] ")
+                       + HelpExampleCli("accounttoutxos", "sender_address \"{\"address1\":\"1.0@DFI\",\"address2\":[\"2.0@BTC\", \"3.0@ETH\"]}' []")
```
monstrobishi, line 2094: The single quotes are only on the ends here, need ones at the start

monstrobishi (changes requested):

src/masternodes/mn_rpc.cpp
ShengguangXiao: Yes, the quote need to keep, because parameters need to parse as JSON, otherwise the example cannot run

bvbfan:
Why it's needed single quotes? An array `[ ]`,  an object `{ }`, an array of objects `[ { }, { } ]`

monstrobishi:
@bvbfan 

The CLI commands need wrapping in double quotes and then escaping the double quotes with \ e.g.:

`defi-cli updatetoken "{\"mintable\": true}"`

@uzyn is updating the examples to wrap in single quotes instead:

`defi-cli updatetoken '{"mintable": true}'`

It's slightly tidier because we don't need to escape everything with \, but some of the above examples the quotes are not matching, so needs some clean up

bvbfan:
Yes, got'cha it miss that whole string is a single parameter.

uzyn:
Can you assist me in fixing this so we can merge it. Thanks.

monstrobishi:
Rebasing to master since v1.3.x was closed

src/masternodes/mn_rpc.cpp
monstrobishi: This line should be 

`HelpExampleCli("updatetoken", " 'token {\"isDAT\":true}' "`

instead of

`HelpExampleCli("updatetoken", "\"token {\"isDAT\":true}' "`

bvbfan:
It has some strange conflicts that i cannot reproduce, so i just fetch master and add a commit on top of. Now all strings should be fixed.

monstrobishi:
Looks good now, merging in

monstrobishi (approved):

PR #92, opened 2020-11-05
Fix wallet locking behavior

bvbfan:
- Fix build with newer boost

- Wallet should be unlocked only in when it sign a transaction

- Small refactor as well as small .yaml syntax correction

### DIFF
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -15,6 +15,7 @@
 #include <sync.h>
 #include <ui_interface.h>
 
+#include <deque>
 #include <memory>
 #include <stdio.h>
 #include <stdlib.h>
--- a/src/masternodes/accounts.h
+++ b/src/masternodes/accounts.h
@@ -14,7 +14,7 @@
 class CAccountsView : public virtual CStorageView
 {
 public:
-    void ForEachBalance(std::function<bool(CScript const & owner, CTokenAmount const & amount)> callback, BalanceKey start) const;
+    void ForEachBalance(std::function<bool(CScript const & owner, CTokenAmount const & amount)> callback, BalanceKey start = {}) const;
     CTokenAmount GetBalance(CScript const & owner, DCT_ID tokenID) const;
 
     Res SetBalance(CScript const & owner, CTokenAmount amount);
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -159,8 +159,7 @@ static CWallet* GetWallet(const JSONRPCRequest& request) {
     std::shared_ptr<CWallet> const wallet = GetWalletForJSONRPCRequest(request);
     CWallet* const pwallet = wallet.get();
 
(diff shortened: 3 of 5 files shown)


src/httpserver.cpp
```diff
@@ -15,6 +15,7 @@
 #include <sync.h>
 #include <ui_interface.h>
 
+#include <deque>
```
ShengguangXiao, line 18: Why need to add this include?

src/masternodes/mn_rpc.cpp
```diff
@@ -310,7 +312,7 @@ UniValue resignmasternode(const JSONRPCRequest& request) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("The masternode %s does not exist", nodeIdStr));
         }
         ownerDest = nodePtr->ownerType == 1 ? CTxDestination(PKHash(nodePtr->ownerAuthAddress)) : CTxDestination(WitnessV0KeyHash(nodePtr->ownerAuthAddress));
-        targetHeight = ::ChainActive().Height() + 1;
+        targetHeight = chainHeight(*pwallet->chain().lock()) + 1;
```
ShengguangXiao, line 315: What's the purpose of this change? In line 309 the chain is locked already

monstrobishi:
Can you confirm if this also allows `listtokens`, `listaccounts`, `gettoken` and `getaccount` to be called when the wallet is locked?

monstrobishi:
I have updated the title here. Please provide a more descriptive title in the next PR :)

src/httpserver.cpp
bvbfan: It's used but not included, probably someone include it indirectly but now it doesn't, so it wont compile
https://github.com/DeFiCh/ain/pull/92/files/10eff033afbd58dd9c77646803b70a6c84d787a5#diff-63c8cb9c9dd61d50d59afd5c39914e1c259f8743030b637a7896a0746c851ef1R76

src/masternodes/mn_rpc.cpp
bvbfan: You're right, revert.

bvbfan:
> Can you confirm if this also allows `listtokens`, `listaccounts`, `gettoken` and `getaccount` to be called when the wallet is locked?

I can confirm.

monstrobishi:
Looks good

src/masternodes/accounts.h
```diff
@@ -14,7 +14,7 @@
 class CAccountsView : public virtual CStorageView
 {
 public:
-    void ForEachBalance(std::function<bool(CScript const & owner, CTokenAmount const & amount)> callback, BalanceKey start) const;
+    void ForEachBalance(std::function<bool(CScript const & owner, CTokenAmount const & amount)> callback, BalanceKey start = {}) const;
```
monstrobishi, line 17: Will this affect consensus?

bvbfan: No, AFAIK.

PR #95, opened 2020-11-06
Fix updatetoken first parameter is string but need to input as json problem

ShengguangXiao:
Fix updatetoken first parameter is string but need to input as json problem, for example, now the first parameter is token symbol, need to input like this

```
defi-cli updatetoken '"DGT#133"' "{\"mintable\":false}"
```

After change, can input like this, without the string quote in single quotes

```
defi-cli updatetoken "DGT#133" "{\"mintable\":false}"
```

or without quote

```
defi-cli updatetoken DGT#133 "{\"mintable\":false}"
```

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -179,7 +179,6 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "listmasternodes", 1, "verbose" },
     { "createtoken", 0, "metadata" },
     { "createtoken", 1, "inputs"},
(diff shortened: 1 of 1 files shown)


bvbfan:
In `feature_tokens_dat` it accepts strings.

ShengguangXiao:
> In `feature_tokens_dat` it accepts strings.

Yes, in the unittest, it is actually not using defi-cli tool, it is using python to package the query and send to the chain. There should be some unit tests to cover defi-cli.

PR #96, opened 2020-11-10
Make listtokens and gettoken reply as single layer

ShengguangXiao:
Make listtokens and gettoken reply as single layer. `listtokens` reply example after change:

```
[
  {
    "id": 0,
    "symbol": "DFI",
    "symbolKey": "DFI",
    "name": "Default Defi token",
    "decimal": 8,
    "limit": 0,
    "mintable": true,
    "tradeable": true,
    "isDAT": true,
    "isLPS": false,
    "finalized": false,
    "minted": 0.00000000,
    "creationTx": "0000000000000000000000000000000000000000000000000000000000000000",
    "creationHeight": -1,
    "destructionTx": "ffffffff00000000000000000000000000000000000000000000000000000000",
    "destructionHeight": -1,
    "collateralAddress": ""
  },
  {
    "id": 1,
    "symbol": "LP128129",
    "symbolKey": "LP128129",
    "name": "-",
    "decimal": 8,
    "limit": 0,
    "mintable": false,
    "tradeable": true,
    "isDAT": true,
    "isLPS": true,
    "finalized": true,
    "minted": 0.00000000,
    "creationTx": "77d87bab82b29ef83e89cda510fbe1e0c2060e712d02fe0afbd06e41e37518ea",
    "creationHeight": 556,
    "destructionTx": "0000000000000000000000000000000000000000000000000000000000000000",
    "destructionHeight": -1,
    "collateralAddress": "undefined"
  },
  {
    "id": 2,
    "symbol": "LP129130",
    "symbolKey": "LP129130",
    "name": "-",
    "decimal": 8,
    "limit": 0,
    "mintable": false,
    "tradeable": true,
    "isDAT": true,
    "isLPS": true,
    "finalized": true,
    "minted": 0.00000000,
    "creationTx": "a10b0bead42ede793452b55b4c64139d078d9c975b4f88a760bfde3d412a8b5e",
    "creationHeight": 571,
    "destructionTx": "0000000000000000000000000000000000000000000000000000000000000000",
    "destructionHeight": -1,
    "collateralAddress": "undefined"
  }
]
```

`gettoken` reply example after change:

```
{
  "id": 131,
  "symbol": "MyToken3",
  "symbolKey": "MyToken3#131",
  "name": "",
  "decimal": 8,
  "limit": 0,
  "mintable": true,
  "tradeable": true,
  "isDAT": false,
  "isLPS": false,
  "finalized": false,
  "minted": 24006.00000000,
  "creationTx": "f52c81743d3ab2a206bfb98cbd8109a0eac12cb2d56115fc6ad3961a548bb73d",
  "creationHeight": 576,
  "destructionTx": "0000000000000000000000000000000000000000000000000000000000000000",
  "destructionHeight": -1,
  "collateralAddress": "mswsMVsyGMj1FzDMbbxw2QW3KvQAv2FKiy"
}
```

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -903,6 +903,7 @@ UniValue updatetoken(const JSONRPCRequest& request) {
 
 UniValue tokenToJSON(DCT_ID const& id, CTokenImplementation const& token, bool verbose) {
     UniValue tokenObj(UniValue::VOBJ);
+    tokenObj.pushKV("id", uint64_t(id.v));
     tokenObj.pushKV("symbol", token.symbol);
     tokenObj.pushKV("symbolKey", token.CreateSymbolKey(id));
 
@@ -928,9 +929,7 @@ UniValue tokenToJSON(DCT_ID const& id, CTokenImplementation const& token, bool v
             tokenObj.pushKV("collateralAddress", "undefined");
         }
     }
-    UniValue ret(UniValue::VOBJ);
-    ret.pushKV(id.ToString(), tokenObj);
-    return ret;
+    return tokenObj;
 }
 
 // @todo implement pagination, similar to list* calls below
@@ -994,9 +993,9 @@ UniValue listtokens(const JSONRPCRequest& request) {
 
     LOCK(cs_main);
 
-    UniValue ret(UniValue::VOBJ);
+    UniValue ret(UniValue::VARR);
     pcustomcsview->ForEachToken([&](DCT_ID const& id, CTokenImplementation const& token) {
-        ret.pushKVs(tokenToJSON(id, token, verbose));
+        ret.push_back(tokenToJSON(id, token, verbose));
 
         limit--;
         return limit != 0;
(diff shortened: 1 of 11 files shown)


bvbfan:
Can you update `RPCResult` of both function

ShengguangXiao:
> Can you update `RPCResult` of both functions

Yes, this change will handle both functions.

uzyn:
While this is good, it might break any integrations that we may have. 

I'm pro for this change, but we need to put up a major notice that this is an RPC API change.

uzyn:
I would suggest the following:
- Do not change the interface for the same minor version. Introduce `--improved` flag for the new format.
- For 1.4.x we can introduce backward-incompatible change but we need to make sure to really highlight the difference.

Please break this PR into 2, 1 to master, 1 to `v1.4.x` if the team is agreeable with my suggestion.

Also, are you also introducing tokensupply?

ShengguangXiao:
Yes, when run unittest, i found a lot of unit test cases need to update, quite trouble some, it may break the defi-app too. I suggest we can push it to 1.4.x

ShengguangXiao:
During unit test, i found `getpoolpair` and `listpoolpairs` also returns "id" : "details" style reply, i think also need a PR to change so all the API reply can keep the same format style.

ShengguangXiao:
@uzyn @monstrobishi   And in the `RPCResult`, the token id now is integer type, if we are going to change the token id to a hash, then we should change it to a string, what's your opinion?

monstrobishi:
This presents an API change, as suggested by UZyn we will be merging into `v1.4.x`, I will create a branch and change base

monstrobishi:
Closing this sorry but I don't foresee changing the structure in the near future, will reopen once we consider this more

PR #100, opened 2020-11-15
Add script to generate burn address

ShengguangXiao:
Add script to generate burn address

### DIFF
--- a/contrib/testgen/gen_burn_addr.py
+++ b/contrib/testgen/gen_burn_addr.py
@@ -124,6 +124,15 @@ def print_usage():
     print('  python3 gen_burn_addr.py 8addressForBurn')
     print('  python3 gen_burn_addr.py 7AddressForBurn')
 
+def check_start_range(fst2):
+    if fst2 >= '73' and fst2 <= '7R':
+        return True
+
+    if fst2 >= '8F' and fst2 <= '8d':
+        return True
+
+    return False
+
 if __name__ == '__main__':
     # Check our input parameters for this script
     if (len(sys.argv) < 2):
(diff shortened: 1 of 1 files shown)


monstrobishi:
This works great, will merge in.

One minor note, some addresses don't seem to generate valid address e.g.:

`8defichainburnaddressXXXXXXXYYFnKm` work but `8DeFiChainBurnAddressXXXXXXXXvVbAC` does not work, is this supposed to be the case?

monstrobishi (approved):

uzyn:
Looks good, is there a possibility of making this work for capital letters too @ShengguangXiao ?

ShengguangXiao:
@uzyn @uzyn 
The script can work for Capital characters too, but only start with some special characters are accepted by the DefiChain.
The Defi Mainnet address should start with string from 8F ~ 8d
The defi Testnet address start with string from 73 ~ 7R,
This is determined by the chain parameter 
```
base58Prefixes[PUBKEY_ADDRESS] = {0x12}; // '8' (0('1') for bitcoin)
```
and for testnet
```
base58Prefixes[PUBKEY_ADDRESS] = {0xf}; // '7' (111 ('m' or 'n') for bitcoin)
```

monstrobishi:
Oh right thanks I missed that detail, could we make the script error out if it's not within either of those ranges?

ShengguangXiao:
> Oh right thanks I missed that detail, could we make the script error out if it's not within either of those ranges? Don't need to add a new parameter, just check against both ranges

OK

ShengguangXiao:
@monstrobishi @uzyn Updated

monstrobishi:
This looks good merging

PR #103, opened 2020-11-19
Feature sendtokenstoaddress

DeFiDev:
Added new custom TX type `AnyAccountsToAccounts`: It allows you to transfer tokens from multiple sources at once.
Added RPC method `sendtokenstoaddress`: It allows you to transfer tokens with auto finding accounts with necessary balances from your wallet.
```
{
    {"from", RPCArg::Type::OBJ, RPCArg::Optional::NO, "",
        {
            {"address", RPCArg::Type::STR, RPCArg::Optional::OMITTED, "The source defi address is the key, the value is amount in amount@token format. "
                                                                        "If obj is empty (no address keys exists) - trying to autoselecting accounts from wallet "
                                                                        "with necessary balances to transfer."},
        },
    },
    {"to", RPCArg::Type::OBJ, RPCArg::Optional::NO, "",
        {
            {"address", RPCArg::Type::STR, RPCArg::Optional::NO, "The defi address is the key, the value is amount in amount@token format. "
                                                                        "If multiple tokens are to be transferred, specify an array [\"amount1@t1\", \"amount2@t2\"]"},
        },
    },
    {"selectionMode", RPCArg::Type::STR, /* default */ "pie", "If param \"from\" is empty this param indicates accounts autoselection mode.
                                                                May be once of:
                                                                "forward" - Selecting accounts without sorting, just as address list sorted."
                                                                "crumbs - Selecting accounts by ascending of sum token amounts."
                                                                It means that we will select first accounts with minimal sum of neccessary token amounts.
                                                                "pie" - Selecting accounts by descending of sum token amounts.
                                                                It means that we will select first accounts with maximal sum of neccessary token amounts."
    },
},
RPCResult{
        "hash"                  (string) The hex-encoded hash of broadcasted transaction"
}
```
Added tests for this feature.

### DIFF
--- a/src/amount.h
+++ b/src/amount.h
@@ -128,6 +128,21 @@ struct CTokenAmount { // simple std::pair is less informative
         this->nValue -= amount;
         return Res::Ok();
     }
+    CAmount SubWithRemainder(CAmount amount) {
+        // safety checks
+        if (amount < 0) {
+            Add(-amount);
+            return 0;
+        }
+        if (this->nValue < amount) {
+            CAmount remainder = amount - this->nValue;
+            this->nValue = 0;
+            return remainder;
+        }
+        // sub
+        this->nValue -= amount;
+        return 0;
+    }
 
     ADD_SERIALIZE_METHODS;
 
--- a/src/masternodes/balances.h
+++ b/src/masternodes/balances.h
@@ -187,9 +187,11 @@ struct CAccountToUtxosMessage {
     }
 };
 
+using CAccounts = std::map<CScript, CBalances>;
+
 struct CAccountToAccountMessage {
     CScript from;
-    std::map<CScript, CBalances> to; // to -> balances
+    CAccounts to; // to -> balances
 
     std::string ToString() const {
         if (to.empty()) {
@@ -212,8 +214,8 @@ struct CAccountToAccountMessage {
 };
 
 struct CAnyAccountsToAccountsMessage {
-    std::map<CScript, CBalances> from; // from -> balances
-    std::map<CScript, CBalances> to; // to -> balances
+    CAccounts from; // from -> balances
+    CAccounts to; // to -> balances
 
     std::string ToString() const {
         if (from.empty() || to.empty()) {
@@ -240,7 +242,7 @@ struct CAnyAccountsToAccountsMessage {
 };
 
 struct CUtxosToAccountMessage {
-    std::map<CScript, CBalances> to; // to -> balances
+    CAccounts to; // to -> balances
 
     std::string ToString() const {
         if (to.empty()) {
@@ -261,7 +263,7 @@ struct CUtxosToAccountMessage {
     }
 };
 
-inline CBalances SumAllTransfers(std::map<CScript, CBalances> const & to) {
+inline CBalances SumAllTransfers(CAccounts const & to) {
     CBalances sum;
     for (const auto& kv : to) {
         sum.AddBalances(kv.second.balances);
--- a/src/masternodes/balances.h
+++ b/src/masternodes/balances.h
@@ -46,6 +46,19 @@ struct CBalances
         }
         return Res::Ok();
     }
+    CTokenAmount SubWithRemainder(CTokenAmount amount) {
+        if (amount.nValue == 0) {
+            return CTokenAmount{amount.nTokenId, 0};
+        }
+        auto current = CTokenAmount{amount.nTokenId, balances[amount.nTokenId]};
+        auto remainder = current.SubWithRemainder(amount.nValue);
+        if (current.nValue == 0) {
+            balances.erase(amount.nTokenId);
+        } else {
+            balances[amount.nTokenId] = current.nValue;
+        }
+        return CTokenAmount{amount.nTokenId, remainder};
+    }
     Res SubBalances(TAmounts const & other) {
         for (const auto& kv : other) {
             auto res = Sub(CTokenAmount{kv.first, kv.second});
@@ -55,6 +68,16 @@ struct CBalances
         }
         return Res::Ok();
     }
+    CBalances SubBalancesWithRemainder(TAmounts const & other) {
+        CBalances remainderBalances;
+        for (const auto& kv : other) {
+            CTokenAmount remainder = SubWithRemainder(CTokenAmount{kv.first, kv.second});
+            // if remainder token value is zero
+            // this addition won't get any effect
+            remainderBalances.Add(remainder);
+        }
+        return remainderBalances;
+    }
     Res AddBalances(TAmounts const & other) {
         for (const auto& kv : other) {
             auto res = Add(CTokenAmount{kv.first, kv.second});
@@ -63,6 +86,14 @@ struct CBalances
             }
         }
         return Res::Ok();
+    }
+
+    CAmount GetAllTokensAmount() {
+        CAmount sum = 0;
+        for (auto& balance : balances) {
+            sum += balance.second;
+        }
+        return sum;
     }
 
     std::string ToString() const {
--- a/src/masternodes/balances.h
+++ b/src/masternodes/balances.h
@@ -180,6 +180,34 @@ struct CAccountToAccountMessage {
     }
 };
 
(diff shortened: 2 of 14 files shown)


DeFiDev:
The accounts are selected in 3 modes of selection: forward, crumbs, pie.
forward - direct bypass of wallet addresses and summation of their balances until we get the right one.
crumbs-collect crumbs, that is, account balances are collected in a selection starting with the smallest.
pie-we collect the required balance for transfers from large accounts.
But if we get a big count of accounts and their balances ("from" and "to"), we will get a huge script. And it can exceed the limits:
```
fAcceptDatacarrier = gArgs.GetBoolArg("-datacarrier", DEFAULT_ACCEPT_DATACARRIER); // this parameter is responsible for checking whether to check the size of the OP_RETURN script, by default is "true"
nMaxDatacarrierBytes = gArgs.GetArg("-datacarriersize", nMaxDatacarrierBytes); // the parameter that sets the maximum size of the OP_RETURN script, by default is 523 bytes
```
So, we have check in RPC method for exceeding of script size and if true - throw exception with message:
```
The output custom script size has exceeded the maximum OP_RETURN script size.
It could have happened because too many \"from\" accounts balances or "to".
If you used autofinding, you can try to use \"pie\" selection mode for decreasing accounts count.
```

src/masternodes/balances.h
```diff
@@ -180,6 +211,34 @@ struct CAccountToAccountMessage {
     }
 };
 
+struct CAnyAccountsToAccountsMessage {
+    std::map<CScript, CBalances> from; // from -> balances
+    std::map<CScript, CBalances> to; // to -> balances
```
ShengguangXiao, line 216: Define an custom type for `std::map<CScript, CBalances>`, it is used in a lot of places.

src/amount.h
```diff
@@ -128,6 +128,21 @@ struct CTokenAmount { // simple std::pair is less informative
         this->nValue -= amount;
         return Res::Ok();
     }
+    CAmount SubWithRemainder(CAmount amount) {
+        // safety checks
+        if (amount < 0) {
+            Add(-amount);
+            return 0;
```
ShengguangXiao, line 135: Why need to support substract negative amount, will it happen, can return error instead?

DeFiDev: As you can see `CAmount SubWithRemainder(CAmount amount);` returns CAmount type, not error type. So we need support negative amount substraction and return remainder anyway.

src/masternodes/balances.h
DeFiDev: Done.

monstrobishi:
This breaks consensus, changing base to v1.4.x

src/masternodes/mn_rpc.cpp
```diff
@@ -3086,39 +3089,326 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+    CBalances residualBalances(targetBalances);
+#ifdef DEBUG
+    LogPrintf("wallet accounts:\n");
+    for (auto &acc : accounts) {
+        LogPrintf("%s: %s\n", acc.first.GetHex(), acc.second.ToString());
+    }
+
+    const auto walletAccountsBalancesSum = SumAllTransfers(accounts);
+    LogPrintf("walletAccountsBalancesSum: %s\n", walletAccountsBalancesSum.ToString());
+#endif
```
ShengguangXiao, line 3168: Remove debug code

src/masternodes/mn_rpc.cpp
```diff
@@ -3086,39 +3089,326 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+#ifdef DEBUG
+        LogPrintf("remainder: %s\n", remainder.ToString());
+#endif
+        // calculate final balances by substraction account balances with remainder
+        // it is necessary to get rid of excess
+        CBalances finalBalances(accIt->second);
+        finalBalances.SubBalances(remainder.balances);
+#ifdef DEBUG
+        LogPrintf("selected balance: %s\n", finalBalances.ToString());
+#endif
```
ShengguangXiao, line 3217: Remove debug code, other places as well.

src/masternodes/mn_rpc.cpp
```diff
@@ -3086,39 +3089,326 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+    CWallet* const pwallet = GetWallet(request);
+
+    RPCHelpMan{"sendtokenstoaddress",
+               "\nCreates (and submits to local node and network) a transfer transaction from your accounts balances (may be picked manualy or autoselected) to the specfied accounts.\n" +
+               HelpRequiringPassphrase(pwallet) + "\n",
+               {
+                    {"from", RPCArg::Type::OBJ, RPCArg::Optional::NO, "",
+                        {
+                            {"address", RPCArg::Type::STR, RPCArg::Optional::OMITTED, "The source defi address is the key, the value is amount in amount@token format. "
+                                                                                     "If obj is empty (no address keys exists) - trying to autoselecting accounts from wallet "
```
ShengguangXiao, line 3253: Should be "then will try to auto-select accounts from wallet"

src/masternodes/mn_rpc.cpp
```diff
@@ -3086,39 +3089,326 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+    pwallet->BlockUntilSyncedToCurrentChain();
+
+    RPCTypeCheck(request.params, {UniValue::VOBJ, UniValue::VOBJ, UniValue::VSTR}, false);
+
+    CAnyAccountsToAccountsMessage msg;
+
+    msg.to = DecodeRecipients(pwallet->chain(), request.params[1].get_obj());
+    const CBalances sumTransfersTo = SumAllTransfers(msg.to);
+    if (sumTransfersTo.balances.empty()) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "zero amounts");
```
ShengguangXiao, line 3296: The error message can with more details, it is too short for user to find out what is wrong.

src/masternodes/mn_rpc.cpp
DeFiDev: Done

src/masternodes/mn_rpc.cpp
DeFiDev: Done

src/masternodes/mn_rpc.cpp
DeFiDev: Done

ShengguangXiao (approved):

src/masternodes/mn_rpc.cpp
```diff
@@ -3081,39 +3081,288 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+                                                                              "  \"pie\" - Selecting accounts by descending of sum token amounts.\n"
+                                                                              "    It means that we will select first accounts with maximal sum of neccessary token amounts."
+                    },
+                },
+                RPCResult{
+                       "\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"
+                },
+                RPCExamples{
+                        HelpExampleCli("sendtokenstoaddress", "\"{}\" "
+                                                    "\"{\\\"dstAddress1\\\":\\\"1.0@DFI\\\",\\\"dstAddress2\\\":[\\\"2.0@BTC\\\", \\\"3.0@ETH\\\"]}\" "
```
bvbfan, line 3243: Can you format helps like that
```
sendtokenstoaddress", "'{}' '{\"dstAddress1\":\"1.0@DFI\",\"dstAddress2\":[\"2.0@BTC\", \"3.0@ETH\"]}' crumbs'
```
escape parameters by `' '` then escaping is minimal.

DeFiDev: Done

bvbfan:
It looks better now. Are you OK, guys, to squash some bloat commits like 'Fix PR issues' or 'Little fix..." which are meaningless? I mean to have consistent history and every commit means exactly which is its name.

monstrobishi:
This is a duplicate of #112 , that PR already contains the changes here, closing

PR #104, opened 2020-11-19
Add ability to transfer token owner

Bushstar:
This PR adds a placeholder variable for the next hard fork as the new feature added in this PR requires a consensus change. Hard fork placeholder is CQHeight, ability to set fork height on regtest added, fork height set to max int on all chains except for regtest due to getblockchaininfo test.

enum TokenUpdateFlags added for use as addition to update token metadata to signal change of owner, CToken is serialised into metadata but contains no owner information as owner is derived from the collateralTx output, hence the requirement for an additional flag. TokenUpdateFlags uses uint8_t with bit 1 set as new owner flag, other bits available for future use.

updatetoken RPC call updated to take optional collateralAddress as new owner. Update of owner on tokens owned by foundationMembers as defined in chainparams is not allowed. After CQHeight TokenUpdateFlags added to meta data in updatetoken RPC call and only after that height does ApplyUpdateTokenAnyTx read TokenUpdateFlags from the meta data.

If metadata read in ApplyUpdateTokenAnyTx signals new owner, token is not founders token and the update TX has a new vout[1] collateral output with the correct amount then the creationTx is set to the update TX hash in UpdateToken. UpdateToken first deletes the old <CreationTx> record before creating a new one and then updates the <ID> record.

Python test has been added to test transfer of owner and ability to change asset and mint tokens after ownership transfer.

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -373,9 +373,18 @@ Res ApplyUpdateTokenAnyTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
 
     uint256 tokenTx;
     CToken newToken;
+    uint8_t updateFlags{static_cast<uint8_t>(TokenUpdateFlags::None)};
+    const uint256* newOwnerTX{nullptr};
+
     CDataStream ss(metadata, SER_NETWORK, PROTOCOL_VERSION);
     ss >> tokenTx;
     ss >> newToken;
+
+    // Read updateflags, added after CQ
+    if (static_cast<int>(height) >= consensusParams.CQHeight && !ss.empty()) {
+        ss >> updateFlags;
+    }
+
     if (!ss.empty()) {
         return Res::Err("%s: deserialization failed: excess %d bytes", base, ss.size());
     }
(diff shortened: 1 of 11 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -405,7 +414,14 @@ Res ApplyUpdateTokenAnyTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
         return Res::Err("%s: %s", base, "tx must have at least one input from token owner");
     }
 
-    auto res = mnview.UpdateToken(token.creationTx, newToken, false);
+    if (!isFoundersToken && updateFlags & static_cast<uint8_t>(TokenUpdateFlags::NewOwner)) {
```
ShengguangXiao, line 417: Can put `updateFlags & static_cast<uint8_t>(TokenUpdateFlags::NewOwner)` inside a (), it is easier to read.

Bushstar: Wil do. There's a couple of locations to update.

ShengguangXiao (approved):

ShengguangXiao:
@Bushstar Can you apply the change of PR https://github.com/DeFiCh/ain/pull/108 here to make CI build pass?

Bushstar:
@ShengguangXiao Added. Tidied the commit history by merging those minor changes into a single commit.

ShengguangXiao:
@Bushstar This PR have conflict with base branch now, can you resolve it?

Bushstar:
@ShengguangXiao resolved.

PR #105, opened 2020-11-20
Auth auto transactions

DeFiDev:
The whole of RPC's authorization mechanics were refactored. 
The main feature: all of the RPCs can create intermediate optional auth tx in the case when some "auth utxos" were missed in the regular authorization process.
This should greatly improve UX. Nothing was broken. Almost all changes touched only internal wallet/coin selector logic.

How it works:
The auth process starts as usual. But if there is no particular "auth utxo" - it is tracked and then, optional tx will be created with minimal (dust) amounts as auth txouts (and with the proper change of course). This tx will be auto funded (as usual) and signed (not sent yet!). Then, all of its txouts will be added into txins of "main" transaction - creating a kind of "chain".
Then, the "main" tx will be auto funded (as usual), with respect to locked/used coins in intermediate auth tx. Then, regular "apply*" checks will be processed. And only in the case of success, both of txs will be sent to mempool.
If some bad things happen, temporarily locked coins (by auth tx) will be unlocked by a special scoped autoobject.

This solution does not send the change back to auth address cause in the common cases this is not a "panacea": any users' activity between custom txs can (and will) grub auth utxos by standard funding process. So, sending the change back to auth addresses is almost useless.
Note that the whole spendable txouts (and the change too) of intermediate optional tx used as the next tx inputs. It was made for simplicity (of coins tracking/selecting) and for the ability to extend such a solution with additional intermediate txs in the future.

The tests in the separate commit addressed every custom tx where auth utxos used.

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -91,8 +91,7 @@ bool HasAuth(CTransaction const & tx, CCoinsViewCache const & coins, CScript con
 {
     for (auto input : tx.vin) {
         const Coin& coin = coins.AccessCoin(input.prevout);
-        assert(!coin.IsSpent());
-        if (coin.out.scriptPubKey == auth)
+        if (!coin.IsSpent() && coin.out.scriptPubKey == auth)
             return true;
     }
     return false;
@@ -108,8 +107,7 @@ bool HasFoundationAuth(CTransaction const & tx, CCoinsViewCache const & coins, C
 {
     for (auto input : tx.vin) {
         const Coin& coin = coins.AccessCoin(input.prevout);
-        assert(!coin.IsSpent());
-        if (consensusParams.foundationMembers.find(coin.out.scriptPubKey) != consensusParams.foundationMembers.end())
+        if (!coin.IsSpent() && consensusParams.foundationMembers.find(coin.out.scriptPubKey) != consensusParams.foundationMembers.end())
             return true;
     }
     return false;
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -11,6 +11,8 @@
 #include <core_io.h>
 #include <consensus/validation.h>
 #include <net.h>
+#include <policy/policy.h>
+#include <policy/settings.h>
 #include <rpc/client.h>
 #include <rpc/server.h>
 #include <rpc/protocol.h>
@@ -25,6 +27,7 @@
 
 //#ifdef ENABLE_WALLET
 #include <wallet/coincontrol.h>
+#include <wallet/fees.h>
 #include <wallet/ismine.h>
 #include <wallet/rpcwallet.h>
 #include <wallet/wallet.h>
@@ -38,9 +41,6 @@
 
 extern bool EnsureWalletIsAvailable(bool avoidException); // in rpcwallet.cpp
 extern bool DecodeHexTx(CTransaction& tx, std::string const& strHexTx); // in core_io.h
-
-extern void FundTransaction(CWallet* const pwallet, CMutableTransaction& tx, CAmount& fee_out, int& change_position,
-                            UniValue options);
 
 extern UniValue ListReceived(interfaces::Chain::Lock& locked_chain, CWallet * const pwallet, const UniValue& params, bool by_label) EXCLUSIVE_LOCKS_REQUIRED(pwallet->cs_wallet);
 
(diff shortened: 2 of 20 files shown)


monstrobishi:
@DeFiDev does this break consensus? also it's not up to date with the master, can you update it thank you

DeFiDev:
> @DeFiDev does this break consensus? also it's not up to date with the master, can you update it thank you
no, it should not break anything.
master updated

DeFiDev:
...something strange with CI/CD scripts upthere.. can you look? The same bug in other PR's

ShengguangXiao:
Some rules have been changed by GitHub, need to update a little in the work flow

monstrobishi:
@DeFiDev  CICD issue has been fixed, but this has conflicts that need to be resolved

bvbfan (commented):
Can you split commit `custom txs auth refactoring; autoauthing by extra chained tx` to consensus, mempool, rpc and wallet ones.

src/masternodes/mn_rpc.cpp
```diff
@@ -128,35 +199,128 @@ std::vector<CTxIn> GetInputs(UniValue const& inputs) {
...
+static boost::optional<CTxIn> GetAnyFoundationAuthInput(CWallet* const pwallet) {
+    CScript anyFounder;
+    for (auto const & founderScript : Params().GetConsensus().foundationMembers) {
+        if (IsMine(*pwallet, founderScript) == ISMINE_SPENDABLE) {
+            CTxDestination destination;
+            if (ExtractDestination(founderScript, destination)) {
+                anyFounder = founderScript;
+                auto auth = GetAuthInputOnly(pwallet, destination);
+                if (auth) {
+                    return auth;
```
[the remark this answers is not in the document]
DeFiDev: I never use assignment (declaration) inside ifs to avoid possible misreading or any side effects of type conversions. But.. as you wish. Done
... And we can remove extra {} in one-statement blocks, and place them on the same line. Such a great economy of lines! :)

src/masternodes/mn_rpc.cpp
```diff
@@ -128,35 +199,128 @@ std::vector<CTxIn> GetInputs(UniValue const& inputs) {
...
+    return fund(mtx, pwallet, {}, true /*lockUnspents*/), sign(mtx, pwallet, {});
+}
+
+static boost::optional<CTxIn> GetAnyFoundationAuthInput(CWallet* const pwallet) {
+    CScript anyFounder;
+    for (auto const & founderScript : Params().GetConsensus().foundationMembers) {
+        if (IsMine(*pwallet, founderScript) == ISMINE_SPENDABLE) {
+            CTxDestination destination;
+            if (ExtractDestination(founderScript, destination)) {
+                anyFounder = founderScript;
```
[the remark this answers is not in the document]
DeFiDev: ok

src/test/applytx_tests.cpp
```diff
@@ -0,0 +1,176 @@
...
+        msg.to = {
+            { CScript(0xA), CBalances{{ {DFI, 10} }} },
+            { CScript(0xB), CBalances{{ {DFI, -1} }} }
+        };
+
+        rawTx.vout[0].scriptPubKey = CreateMetaA2A(msg);
+
+        res = ApplyCustomTx(mnview, coinview, CTransaction(rawTx), amkCheated, 1, 0, false);
+        BOOST_CHECK(!res.ok);
+        BOOST_CHECK(res.msg.find("negative amount") != std::string::npos);
```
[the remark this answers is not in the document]
DeFiDev: ok

src/test/applytx_tests.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        msg.to = {
+            { CScript(0xA), CBalances{{ {DFI, -1} }} }
+        };
+
+        rawTx.vout[0].scriptPubKey = CreateMetaA2A(msg);
+
+        res = ApplyCustomTx(mnview, coinview, CTransaction(rawTx), amkCheated, 1, 0, false);
+        BOOST_CHECK(!res.ok);
+        BOOST_CHECK(res.msg.find("negative amount") != std::string::npos);
```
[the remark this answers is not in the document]
DeFiDev: ok

DeFiDev:
> Can you split commit `custom txs auth refactoring; autoauthing by extra chained tx` to consensus, mempool, rpc and wallet ones.

Done but "consensus" and "mempool" left together due to this is one logical mod, unsplittable

monstrobishi:
This looks like it breaks consensus but there are no height checks

src/masternodes/mn_checks.cpp
```diff
@@ -91,8 +91,7 @@ bool HasAuth(CTransaction const & tx, CCoinsViewCache const & coins, CScript con
 {
     for (auto input : tx.vin) {
         const Coin& coin = coins.AccessCoin(input.prevout);
-        assert(!coin.IsSpent());
-        if (coin.out.scriptPubKey == auth)
+        if (!coin.IsSpent() && coin.out.scriptPubKey == auth)
```
monstrobishi, line 94: Could this affect consensus

monstrobishi:
On closer look it looks ok but want to be sure

src/masternodes/mn_rpc.cpp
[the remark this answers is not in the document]
uzyn: @bvbfan not a good idea to declare variables inside condition. No side effects. 

Number of lines should not matter. Most sane linters would have prevented that.

test/functional/test_runner_my.py
```diff
@@ -0,0 +1,717 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2019 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Run regression test suite.
+
+This module calls down into individual test cases via subprocess. It will
+forward all unrecognized arguments onto the individual test scripts.
+
```
ShengguangXiao, line 9: This file `test_runner_my.py` looks like personal debug test file, should we commit it?

Bushstar: I do not think so, guess it slipped in by accident.

test/functional/feature_accounts_n_utxos.py
```diff
@@ -104,12 +104,12 @@ def run_test(self):
...
-            self.nodes[0].accounttoaccount(toSilver, {accountGold: "100@" + symbolSILVER}, [])
-        except JSONRPCException as e:
-            errorString = e.error['message']
-        assert("Can't find any UTXO" in errorString)
+        # missing (account exists, there are tokens, but not token 0) - autofunded now!
+        # try:
+        #     self.nodes[0].accounttoaccount(toSilver, {accountGold: "100@" + symbolSILVER}, [])
+        # except JSONRPCException as e:
+        #     errorString = e.error['message']
+        # assert("Can't find any UTXO" in errorString)
```
ShengguangXiao, line 112: So this test should pass now? Can just change the result instead of remove it?

src/masternodes/mn_checks.cpp
Bushstar: No. Either way, if `!coin.IsSpent()` is false then the function is not going to return true.

src/masternodes/mn_rpc.cpp
```diff
@@ -128,35 +199,125 @@ std::vector<CTxIn> GetInputs(UniValue const& inputs) {
...
+    }
+    return {};
+}
+
+static boost::optional<CTxIn> GetAuthInputOnly(CWallet* const pwallet, CTxDestination const& auth) {
+
+    std::vector<COutput> vecOutputs;
+    CCoinControl cctl;
+    cctl.m_avoid_address_reuse = false;
+    cctl.m_min_depth = 1;
```
Bushstar, line 215: Could we not look for unconfirmed UTXOs as an auth input? Check it is in mempool and we can trust it will be added to the chain, using unconfirmed UTXOs will also reduce issue at hand and is part of the solution here!

Bushstar, line 215: Could we not look for unconfirmed UTXOs as an auth input? Check it is in mempool and we can trust it will be added to the chain, using unconfirmed UTXOs will also reduce issue at hand and is part of the solution here!

Bushstar:
@DeFiDev the intermediate TX created to generate an auth UTXO has a dust output and a change output, this means the final TX requires that dust output for auth and another one to cover the fee.

Why not just make the auth output enough to cover the TX fee, or even scrap the change address on the intermediate TX and send everything to the auth address?

Then we only need one input on the final TX and will reduce TX size slightly and with no change address on the intermediate TX we reduce the total size of this process further still.

monstrobishi (approved):

PR #106, opened 2020-11-20
[tests] Add multiple token owner test

Bushstar:
DeFi tokens can have multiple owners using a multisig address for collateral. The test here creates a 1-of-3 multisig and using that address changes the name three times using each of the multisig addresses. This script is also a demonstration of multiple owner tokens.

### DIFF
--- a/test/functional/feature_tokens_multisig.py
+++ b/test/functional/feature_tokens_multisig.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright (c) 2014-2019 The Bitcoin Core developers
+# Copyright (c) 2014-2020 The Bitcoin Core developers
 # Copyright (c) DeFi Blockchain Developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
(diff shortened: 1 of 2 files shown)


test/functional/feature_tokens_multisig.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2019 The Bitcoin Core developers
```
ShengguangXiao, line 2: Can change to 2020

ShengguangXiao:
Can add some failure test cases too?

ShengguangXiao (changes requested):

Bushstar:
@ShengguangXiao I'll add a test for 2-of-3 and we can check that 1 single signature will fail and that 2-of-3 works, also check that a single member of the multisig cannot make changes directly.

Bushstar:
@ShengguangXiao 2-of-3 test case added with above mentioned failure cases.

ShengguangXiao (approved):

monstrobishi:
This only touches python tests, merging in

PR #107, opened 2020-11-20
Fix account value representation

bvbfan:


src/masternodes/mn_rpc.cpp
```diff
@@ -1194,7 +1194,7 @@ BalanceKey decodeBalanceKey(std::string const& str) {
 
 std::string tokenAmountString(CTokenAmount const& amount) {
     auto token = pcustomcsview->GetToken(amount.nTokenId);
-    std::string valueString = std::to_string(amount.nValue / COIN) + "." + std::to_string(amount.nValue % COIN);
+    auto valueString = strprintf("%d.%08d", amount.nValue / COIN, amount.nValue % COIN);
```
ShengguangXiao, line 1197: Can add const for these two values?

monstrobishi (approved):

ShengguangXiao (approved):

PR #112, opened 2020-11-25
Feature poolliquidity autofinding

DeFiDev:
Added the auto-selection wallet accounts feature for addpoolliquidity RPC method.
User should set single entry in the 'from' param, where "*" (star) is a key, and string/array of token amounts is a value.


src/masternodes/mn_rpc.cpp
```diff
@@ -3081,39 +3225,174 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+    // encode
+    CDataStream markedMetadata(DfTxMarker, SER_NETWORK, PROTOCOL_VERSION);
+    markedMetadata << static_cast<unsigned char>(CustomTxType::AnyAccountsToAccounts)
+                   << msg;
+    CScript scriptMeta;
+    scriptMeta << OP_RETURN << ToByteVector(markedMetadata);
+
+    if (scriptMeta.size() > nMaxDatacarrierBytes) {
+        throw JSONRPCError(RPC_VERIFY_REJECTED, "The output custom script size has exceeded the maximum OP_RETURN script size."
+                                                "It could have happened because too many \"from\" or \"to\" accounts balances."
```
ShengguangXiao, line 3309: "It could have happened" change to "It may"

src/masternodes/mn_rpc.cpp
```diff
@@ -3081,39 +3225,174 @@ UniValue isappliedcustomtx(const JSONRPCRequest& request) {
...
+                                                "If you used autofinding, you can try to use \"pie\" selection mode for decreasing accounts count.");
```
ShengguangXiao, line 3310: "used" change to "use"

DeFiDev: Which option do you think is better?

src/masternodes/mn_rpc.cpp
DeFiDev: Done

src/masternodes/mn_rpc.cpp
DeFiDev: Done

monstrobishi:
This includes both sendtokenstoaddress and add pool liquidity from any amount

PR #114, opened 2020-11-28
Filter for transactions and addLiquidity slippage protection

uzyn:


src/masternodes/poolpairs.h
```diff
@@ -130,6 +130,20 @@ class CPoolPair : public CPoolPairMessage
...
+            if (maxLiquidity256 / CPoolPair::PRECISION > std::numeric_limits<int64_t>::max()) {
+                return Res::Err( "Max liquidity +3% overflow!");
+            }
+
+            CAmount maxLiquidity = (maxLiquidity256 / arith_uint256(CPoolPair::PRECISION)).GetLow64();;
+
+            if(liquidity < maxLiquidity) {
+                return Res::Err( "Bad liquidity ratio, more than 3% difference");
+            }
+
```
uzyn, line 141: Lines 133-146 should be simplified to:

```cpp

            if ((std:max(liqA, liqB) - liquidity) / liquidity > 0.03) {
                return Res::Err( "Exceeds max ratio slippage protection of 3%");
            }
```

uzyn (commented):
@monstrobishi thanks for your quick fix.

src/masternodes/poolpairs.h
monstrobishi: Done

src/masternodes/poolpairs.h
```diff
@@ -131,7 +131,7 @@ class CPoolPair : public CPoolPairMessage
             CAmount liqB = (arith_uint256(amountB) * arith_uint256(totalLiquidity) / reserveB).GetLow64();
             liquidity = std::min(liqA, liqB);
 
-            if ((std::max(liqA, liqB) - liquidity) / liquidity * 100 > 3) {
+            if ((std::max(liqA, liqB) - liquidity) / liquidity * 100 >= 3) {
```
bvbfan, line 134: Checking liquidity that equals to 0 should be before divide by potentially 0.

bvbfan (changes requested):

src/masternodes/mn_rpc.cpp
```diff
@@ -2896,6 +2977,29 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
+                if(!hasToken) {
+                    return true; // continue
+                }
+            }
+
+            if(noRewards) {
+                if(category == static_cast<unsigned char>(CustomTxType::NonTxRewards)) {
+                    return true; // continue
+                }
+            }
```
ShengguangXiao, line 3003: It looks like the same code repeated 3 times, can put to a function?

bvbfan: I think we can refactor it, later, are you OK? I mean he want to ship fix ASAP.

src/masternodes/poolpairs.h
uzyn: liquidity == 0 is done at line 122. This has excluded liquidity == 0's case.

bvbfan: Line 122 is `totalLiquidity` here is exactly where it should be checked. Line above `liquidity` is calculated as min of 2 liquidities and it can be 0.

uzyn: Can you submit a fix please? @bvbfan

uzyn: Nope it cannot be 0. But I'm okay with a safety net fix here 


```suggestion
            if (liquidity > 0 && (std::max(liqA, liqB) - liquidity) / liquidity * 100 >= 3) {
```

src/masternodes/mn_rpc.cpp
ShengguangXiao: Sounds good

src/masternodes/poolpairs.h
bvbfan: `(std::max(liqA, liqB) - liquidity) * 100 / liquidity` should be OK

ShengguangXiao (approved):

PR #117, opened 2020-11-30
Validate signature on block index load instead of minter public key

bvbfan:
Starting time down from 1 min to 1 sec, extracting minter public key is pretty slow for reason: hash to sign should be calculated on the fly, that slowdown whole validation process. Instead signature can be safe and fast checked that it's valid.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -15,7 +15,6 @@
 #include <uint256.h>
 
 #include <vector>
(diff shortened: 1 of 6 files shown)


ShengguangXiao (approved):

monstrobishi (approved):

monstrobishi:
Just tested this is fantastic! merging in

PR #119, opened 2020-12-01
Remove use of PRECISION in reward calculations

Bushstar:
Remove use of PRECISION used in calculating pool rewards due to potential loss of rewards. It does not give the ability to create rewards at the lower values, currently the minimum possible payout is reward / 10,000. The PRECISION value always has the potential to lose rewards when the total pool liquidity is more than user liquidity times PRECISION.

### DIFF
--- a/src/masternodes/poolpairs.h
+++ b/src/masternodes/poolpairs.h
@@ -258,7 +258,7 @@ public:
     }
 
     /// @attention it throws (at least for debug), cause errors are critical!
-    CAmount DistributeRewards(CAmount yieldFarming, std::function<CTokenAmount(CScript const & owner, DCT_ID tokenID)> onGetBalance, std::function<Res(CScript const & to, CTokenAmount amount)> onTransfer) {
+    CAmount DistributeRewards(CAmount yieldFarming, std::function<CTokenAmount(CScript const & owner, DCT_ID tokenID)> onGetBalance, std::function<Res(CScript const & to, CTokenAmount amount)> onTransfer, bool newRewardCalc = false) {
 
         uint32_t const PRECISION = 10000; // (== 100%) just searching the way to avoid arith256 inflating
         CAmount totalDistributed = 0;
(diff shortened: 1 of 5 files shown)


src/masternodes/poolpairs.h
```diff
@@ -258,7 +258,7 @@ class CPoolPairView : public virtual CStorageView
     }
 
     /// @attention it throws (at least for debug), cause errors are critical!
-    CAmount DistributeRewards(CAmount yieldFarming, std::function<CTokenAmount(CScript const & owner, DCT_ID tokenID)> onGetBalance, std::function<Res(CScript const & to, CTokenAmount amount)> onTransfer) {
+    CAmount DistributeRewards(CAmount yieldFarming, std::function<CTokenAmount(CScript const & owner, DCT_ID tokenID)> onGetBalance, std::function<Res(CScript const & to, CTokenAmount amount)> onTransfer, bool newRewardCalc = false) {
```
ShengguangXiao, line 261: Didn't see anywhere pass `true` for `newRewardCalc `

ShengguangXiao:
Looks good, better to have some unittest to verify the calculation result.

src/masternodes/poolpairs.h
```diff
@@ -286,17 +286,26 @@ class CPoolPairView : public virtual CStorageView
...
                 if (pool.swapEvent) {
                     CAmount feeA = pool.blockCommissionA * liqWeight / PRECISION;       // liquidity / pool.totalLiquidity;
+                    if (newRewardCalc) {
+                        feeA = pool.blockCommissionA * liquidity / pool.totalLiquidity;
+                    }
                     distributedFeeA += feeA;
                     onTransfer(provider, {pool.idTokenA, feeA}); //can throw
 
                     CAmount feeB = pool.blockCommissionB * liqWeight / PRECISION;       // liquidity / pool.totalLiquidity;
+                    if (newRewardCalc) {
```
uzyn, line 296: I feel it would actually be better the other way round. `if (!newRewardCalc)` because the main flow should be the default flow and the condition is for the deprecated calc.

It's just minor code styling, if you can fix it it woudl be best, if not, not a blocker.

uzyn (changes requested):
Thanks for the quickfix @Bushstar 

Can you also include unit tests?

esp, one that this fix addresses?

liq = 1
totalliquidity = 1e11
reward = 1e11

expects 1 sat/fi

src/masternodes/poolpairs.h
Bushstar: Updated.

bvbfan:
Let not make code `spaghetti` i would prefer
```
if (!newRewardCalc) {
    return oldCalculation(....)
}
```
Then keep new calculation clear because we may changed again in future. I would prefer keep pld implementations out of source tree and to in separate functions.

PR #120, opened 2020-12-02
Enable filter by tx type in listaccounthistory rpc method

bvbfan:


src/masternodes/mn_rpc.cpp
```diff
@@ -2861,6 +2863,7 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
     uint32_t depth = 100;
     bool noRewards = false;
     std::string tokenFilter;
+    uint8_t txType = uint8_t(CustomTxType::None);
```
ShengguangXiao, line 2866: Can use `static_cast`

ShengguangXiao, line 2866: Can define as `const auto txType = CustomTxType::None;`

src/masternodes/mn_rpc.cpp
```diff
@@ -2886,6 +2890,13 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
         if (!optionsObj["token"].isNull()) {
             tokenFilter = optionsObj["token"].get_str();
         }
+
+        if (!optionsObj["txtype"].isNull()) {
+            const auto str = optionsObj["txtype"].get_str();
+            if (str.size() == 1) {
+                txType = uint8_t(CustomTxCodeToType(str[0]));
```
ShengguangXiao, line 2897: Avoid to cast here

src/masternodes/mn_rpc.cpp
```diff
@@ -2902,6 +2913,10 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
             if (height > startKey.blockHeight || (depth <= startKey.blockHeight && (height < startKey.blockHeight - depth)))
                 return true; // continue
 
+            if (txType && txType != category) {
```
ShengguangXiao, line 2916: Should explicitly compare with `CustomTxType::None` as `CustomTxType::None != txType`

src/masternodes/mn_rpc.cpp
```diff
@@ -2945,6 +2960,10 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
                 return true; // continue
             }
 
+            if (txType && txType != category) {
```
ShengguangXiao, line 2963: Same as above

src/masternodes/mn_rpc.cpp
bvbfan: If i get it above by `CustomTxType` i should cast here, category is unsigned char

ShengguangXiao: Yes, but at least the first element need to do comparison.

bvbfan: We know `None` is 0, it's not going to change :) I'll do the change.

bvbfan:
Reword a commit, fix grammar.

src/masternodes/mn_rpc.cpp
```diff
@@ -3032,10 +3040,26 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
+            if (str.size() == 1) {
+                txType = CustomTxCodeToType(str[0]);
+            }
+        }
+        if (!optionsObj["limit"].isNull()) {
+            limit = (uint32_t) optionsObj["limit"].get_int64();
+            depth = std::numeric_limits<decltype(depth)>::max();
+        }
+        if (limit == 0) {
+            limit = std::numeric_limits<decltype(limit)>::max();
```
ShengguangXiao, line 3055: If the `limint` is max of uint32_t, will the API call timeout?

bvbfan: I don't get what timeout you mean?

ShengguangXiao (approved):

monstrobishi (approved):

monstrobishi:
Fixed conflicts myself

PR #122, opened 2020-12-02
Hardfork for slopeswap

monstrobishi:
Added height selection for #79

### DIFF
--- a/src/masternodes/poolpairs.cpp
+++ b/src/masternodes/poolpairs.cpp
@@ -150,7 +150,7 @@ Res CPoolPair::Swap(CTokenAmount in, PoolPrice const & maxPrice, std::function<R
     return onTransfer({ forward ? idTokenB : idTokenA, result });
 }
 
-CAmount CPoolPair::slopeSwap(CAmount unswapped, CAmount &poolFrom, CAmount &poolTo, bool postBayfrontGardens) {
+CAmount CPoolPair::slopeSwap(CAmount unswapped, CAmount &poolFrom, CAmount &poolTo, bool postBayfrontGardens = false) {
     assert (unswapped >= 0);
     assert (SafeAdd(unswapped, poolFrom).ok);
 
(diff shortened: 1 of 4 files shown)


monstrobishi:
Please don't review yet, I'm still making changes

monstrobishi:
Ok review please

src/masternodes/poolpairs.h
```diff
@@ -183,10 +183,10 @@ class CPoolPair : public CPoolPairMessage
         return onReclaim(address, resAmountA, resAmountB);
     }
 
-    Res Swap(CTokenAmount in, PoolPrice const & maxPrice, std::function<Res(CTokenAmount const &)> onTransfer);
+    Res Swap(CTokenAmount in, PoolPrice const & maxPrice, std::function<Res(CTokenAmount const &)> onTransfer, bool postBayfrontGardens = false);
```
uzyn, line 186: should the default be `true` instead? It doesn't matter much, but the default should be what's more likely to be more common, which in this case would be `true`.

uzyn (commented):
Thanks @monstrobishi

uzyn (approved):

src/masternodes/poolpairs.cpp
```diff
@@ -150,17 +150,33 @@ Res CPoolPair::Swap(CTokenAmount in, PoolPrice const & maxPrice, std::function<R
...
+            CAmount stepFrom = std::min(chunk, unswapped);
+            arith_uint256 stepFrom256(stepFrom);
+            arith_uint256 stepTo = poolT * stepFrom256 / poolF;
+            poolF += stepFrom256;
+            poolT -= stepTo;
+            unswapped -= stepFrom;
+            swapped += stepTo;
+        }
+    } else {
+        arith_uint256 unswappedA = arith_uint256(unswapped);
```
ShengguangXiao, line 160: I think should put post logic in front, the old logic will always be skipped in the future

ShengguangXiao (changes requested):

src/masternodes/poolpairs.cpp
monstrobishi: we can refactor this later, it doesn't change the operation

PR #128, opened 2020-12-03
Better handling of account history

bvbfan:
This a WIP, i have plan to store rewards by pool id as well, should we have it for account history as well? Now we have owner <-> token <-> amount, but if it's not a pool account we can use token id instead?

### DIFF
--- a/src/flushablestorage.h
+++ b/src/flushablestorage.h
@@ -313,6 +313,24 @@ private:
     MapKV changed;
 };
 
+template<typename T>
+struct CLazySerialize
+{
+    CStorageKVIterator& it;
+
+    operator T()
+    {
+        return get();
+    }
+
+    T get()
+    {
+        T value;
+        BytesToDbType(it.Value(), value);
+        return value;
+    }
+};
+
 class CStorageView {
 public:
     CStorageView() = default;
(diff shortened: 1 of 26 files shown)


monstrobishi:
much nicer can you confirm if this has an impact on performance

bvbfan:
It's a bit faster, really small, what's save is serialize/deserialize to stream. I plan to make another class to handle rewards and to store rewards by pool id (in validation.cpp where rewards are redistributed)

ShengguangXiao (approved):

monstrobishi:
@bvbfan is this still a work in progress? you can change the title when you feel it's ready to merge in

bvbfan:
It's no more a WIP, rewards are two types, `Commission` and `PoolReward` they are stored separate from account history.

bvbfan:
`listaccounthistory` rpc call remains unchanged for account history, it has `txid` and `txn` the difference is there is no empty `txid` anymore, it means when it has `txid` it's account history, `Rewards` type is split to `Commission` and `PoolReward`, also rewards have `poolID`.
Account:
`owner`
`blockHeight`
`type` (`Rewards` is not valid type anymore)
`txn`
`txid`
`amounts`
Reward:
`owner`
`blockHeight`
`type` (`Commission` or `PoolReward`)
`poolID`
`amounts`

monstrobishi:
this could break the app if the Rewards type does not exist anymore, can we name "PoolReward" back to "Rewards" so that the type still exists, and split commission out

bvbfan:
Added https://github.com/cakedefi/defichain-private/issues/63

monstrobishi:
@bvbfan after merging the other PR (for including UTXO) this has quite a few conflicts

bvbfan:
Rebased.

bvbfan:
All unit tests pass locally, changes not affect wallet at all.

monstrobishi (approved):

bvbfan:
@Bushstar can you test new approach?

Bushstar (approved):

PR #133, opened 2020-12-09
RPC clearmempool

Bushstar:
Add clearmempool RPC call, useful in testing when rolling back blocks to retry a transaction without old transaction creating problems.


monstrobishi:
looks simple and doesn't touch anything, safe to merge now

PR #134, opened 2020-12-10
Fix verifying block validity

bvbfan:
We can have potentially invalid block due to test tx in isolation. It addresses yesterday stalled blockchain.


monstrobishi:
Can we confirm this solves the issue with a unit test

bvbfan:
Unit test just shows the block isn't correct and it's never minted.

PR #135, opened 2020-12-11
Fix sendtokenstoaddress

bvbfan:
Sorting is wrong, that makes auth tx to be missing.


monstrobishi:
any side effect of removing clear duplicate inputs?

bvbfan:
They are unique in `GetAuthInputsSmart` afterwards they are sorted by hash despite the n output.

PR #136, opened 2020-12-11
Prevent transactions in blocks from going below balance.

Bushstar:
Tracks changes to account when including transactions in a blocks, this prevents an account trying to spend more than it has which is an invalid state and would be rejected. After block creation some transactions left in the mempool may then be invalid due to our bespoke account layer, we iterate over the mempool and remove any transactions that fail CheckTxInputs.

### DIFF
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -5,6 +5,7 @@
 
 #include <txmempool.h>
 
+#include <chainparams.h>
 #include <consensus/consensus.h>
 #include <consensus/tx_verify.h>
 #include <consensus/validation.h>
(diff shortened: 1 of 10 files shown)


test/functional/feature_account_mining.py
```diff
@@ -0,0 +1,47 @@
...
+        blockcount = node.getblockcount()
+
+        # Generate a block
+        node.generate(1)
+
+        # Check the blockchain has extended as expected
+        assert_equal(node.getblockcount(), blockcount + 1)
+
+        # Account should now be empty
+        assert_equal(node.getaccount(account), [])
```
ShengguangXiao, line 44: Can also check the count of transactions?

Bushstar: I'll add it. We would expect five TXs in the block plus coinbase. This test will also be extended to include other account TXs.

src/txmempool.cpp
```diff
@@ -577,6 +578,49 @@ void CTxMemPool::removeForBlock(const std::vector<CTransactionRef>& vtx, unsigne
         removeConflicts(*tx);
         ClearPrioritisation(tx->GetHash());
     }
+
+    bool accountConflict{false};
+
+    // Check if any custom TXs are in mempool with conflict
+    for (indexed_transaction_set::const_iterator it = mapTx.begin(); it != mapTx.end(); ++it) {
```
ShengguangXiao, line 585: Use auto instead of indexed_transaction_set::const_iterator

src/txmempool.cpp
```diff
@@ -577,6 +578,49 @@ void CTxMemPool::removeForBlock(const std::vector<CTransactionRef>& vtx, unsigne
...
+    }
+
+    // Account conflict, check entire mempool
+    if (accountConflict) {
+        std::set<CTransactionRef> txsToRemove;
+        CCoinsViewCache mempoolDuplicate(const_cast<CCoinsViewCache*>(&::ChainstateActive().CoinsTip()));
+        CAmount txfee = 0;
+
+        // Check custom TX consensus types are now not in conflict with account layer
+        for (indexed_transaction_set::const_iterator it = mapTx.begin(); it != mapTx.end(); ++it) {
```
ShengguangXiao, line 604: Use `auto` instead of `indexed_transaction_set::const_iterator`

PR #137, opened 2020-12-14
Prevent bad tx propagation

bvbfan:
This alternative approach to address stalled blockchain issue. It tries to keep mempool consistent on node level and prevent propagation of such a bad tx to other nodes.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -561,6 +561,7 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
     {
         CCoinsView dummy;
         CCoinsViewCache view(&dummy);
+        CCustomCSView mnview(*pcustomcsview);
 
         LockPoints lp;
         CCoinsViewCache& coins_cache = ::ChainstateActive().CoinsTip();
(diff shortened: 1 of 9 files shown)


Bushstar (commented):
I like the idea. We could first check the TX is a CustomTxType before looping over all mempool TXs. Can you also add a check for accounttoaccount? I found that I can still recreate the issue in block creation with the following.

```
for _ in range(10):
    node.accounttoaccount(account, {destination: "1@DFI"})
    node.accounttoutxos(account, {destination: "1@DFI"})
```

bvbfan:
`NotAllowedToFail` does not include `accounttoaccount` if you ask me it should.

monstrobishi (approved):

src/validation.cpp
```diff
@@ -601,8 +602,21 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
...
+        const auto height = GetSpendHeight(view);
+        std::vector<unsigned char> metadata;
+
+        // check for txs in mempool
+        for (const auto& e : mempool.mapTx) {
+            const auto& tx = e.GetTx();
+            const auto txType = GuessCustomTxType(tx, metadata);
+            if (NotAllowedToFail(txType)) {
+                auto res = ApplyCustomTx(mnview, view, tx, chainparams.GetConsensus(), height, 0, false);
+                assert(res.ok || !(res.code & CustomTxErrCodes::Fatal)); // inconsistent mempool
```
monstrobishi, line 614: don't assert here, reject tx instead

monstrobishi (changes requested):

ShengguangXiao (approved):

PR #140, opened 2020-12-16
Do not force boost filesystem no deprecated

bvbfan:
Forcing BOOST_FILESYSTEM_NO_DEPRECATED makes no sense in source. It can added to configure by CCFLAGS.
Fix build with Boost > 1.74

### DIFF
--- a/src/fs.h
+++ b/src/fs.h
@@ -11,7 +11,6 @@
 #include <ext/stdio_filebuf.h>
 #endif
 
-#define BOOST_FILESYSTEM_NO_DEPRECATED
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
 


ShengguangXiao:
What's the compile error if boost version > 1.74?

Bushstar:
@ShengguangXiao here's the upstream change.

https://github.com/bitcoin/bitcoin/commit/a64e97dd476bda7c7981979d045b0d06d6f7ce47#diff-3492213c413b2d41f5e11a676db8f69ff54f71a2f5a14b2f61ee919d3a9f5122

From the above commit text.

```
wallet/walletutil.cpp:77:23: error: no member named 'level' in 'boost::filesystem::recursive_directory_iterator'
        } else if (it.level() == 0 && it->symlink_status().type() == fs::regular_file && IsBerkeleyBtree(it...
                   ~~ ^
```

@bvbfan Personally I like to see a cherry-pick of upstream changes.

bvbfan:
Also `copy_option` is deprecated and renamed to `copy_options` to match stdc++17 one.

bvbfan:
cherry-picked

ShengguangXiao:
Actually i have fixed this issue before, don't know why it comes back.
https://github.com/DeFiCh/ain/pull/26

bvbfan:
@ShengguangXiao the issue is different here
`fs::copy_file(pathSrc, pathDest, fs::copy_option::overwrite_if_exists);`
It does not compile since `copy_option` is deprecated i wrote it above.

ShengguangXiao:
@bvbfan You can fix the compile issue with new boost like below. Eventually there will be one day we also need to upgrade C++ and boost, so we should adopt new definitions.
https://github.com/Icinga/icinga2/pull/8184/files

Bushstar:
@ShengguangXiao I would rather we follow upstream solutions from Bitcoin, the fix applied here is a cherry-pick from Bitcoin.

bvbfan:
@ShengguangXiao also forcing definition in header/source file does not look right. It should be done via configure CPPFLAGS.

ShengguangXiao:
Ok, then should be good. We follow Bitcoin.

ShengguangXiao (approved):

PR #141, opened 2020-12-16
Refactor flushable storage iterator

bvbfan:
`CFlushableStorageKVIterator` is every time one ahead for no reason, prevent from unneeded storage reads.


ShengguangXiao:
Do we have a unittest to verify the change is working?

bvbfan:
All unit test pass :)

PR #142, opened 2020-12-16
Set CLIENT_VERSION to BUILD_DESC

bvbfan:
It sets CLIENT_VERSION accordingly to BUILD_DESC which is set at compilation time.

### DIFF
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -3,6 +3,7 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include <clientversion.h>
+#include <string>
 
 #include <tinyformat.h>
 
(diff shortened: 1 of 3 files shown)


monstrobishi:
will this reliably not crash depending on the string set?

bvbfan:
Will not crash, if string not match pattern (means not equals or start with) `major, minor, rev` remains unchanged say 0-es.

monstrobishi (approved):

monstrobishi:
Travis failed due to error:
```
The locale dependent function sscanf(...) appears to be used:
src/clientversion.cpp:        sscanf(BUILD_DESC, "v%d.%d.%d", &major, &minor, &rev);
Unnecessary locale dependence can cause bugs that are very
tricky to isolate and fix. Please avoid using locale dependent
functions if possible.
```

bvbfan:
There is no locale here. `v.1.3.14`

monstrobishi:
it's saying sscanf is locale dependent, maybe there is an alternative function?

bvbfan:
We can use
```
 std::setlocale(LC_ALL, "C");
```
I've not aware of other function.

bvbfan:
Use `std::stoi` a bit more ugly.

bvbfan:
I'll exclude it in `lint-locale-dependence.sh` as others.

uzyn:
@bvbfan This does not work. Running v1.3.15rc2

```
$ defi-cli -getinfo
{
  "version": 0, // it's now 0
  "protocolversion": 70016,
  "blocks": 519134,
}
```

monstrobishi:
@uzyn this won't work with the rc versions because of the non numerical text in the build version

uzyn:
@monstrobishi Got it. cool!

PR #143, opened 2020-12-17
Fix order in listaccounthistory

bvbfan:
It changes behaviour of listaccounthistory rpc call, maxBlockHeight is upper block height, depth is how many blocks to revert (default to 100) it means if maxBlockHeight is 1000, depth = 100 it reads blocks between 900 - 1000 history and rewards, returns descendent order limit by limit.


monstrobishi (approved):

bvbfan:
I'll push one more fix.

bvbfan:
The fix is when we have different owners in same block we should advance to next block, when next block is maxBlockHeight + 1 we should end searching.

PR #151, opened 2020-12-23
Search all addresses for specified address

Bushstar:
This is a quick solution for the issue discovered where no results were returned for an address from account or reward history, instead we now iterate over every entry and filter for what we require. Also owner was used by mistake when searching wallet for a specified address, account should have been used instead.


bvbfan:
Silly mistake.

ShengguangXiao (approved):

bvbfan:
I prefer to not search in all address, that's will make it slower.

ShengguangXiao:
Maybe can may search all as an option

PR #152, opened 2020-12-24
Trying to make listaccounthistory faster

bvbfan:
Keep track of `IsMine` results in map to speedup lookup history.

### DIFF
--- a/src/flushablestorage.h
+++ b/src/flushablestorage.h
@@ -6,6 +6,8 @@
 #define DEFI_FLUSHABLESTORAGE_H
 
 #include <dbwrapper.h>
+#include <optional.h>
+
 #include <boost/optional.hpp>
 #include <boost/scoped_ptr.hpp>
 #include <boost/thread.hpp>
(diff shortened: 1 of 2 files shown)


Bushstar (commented):
Doing a quick benchmark it took 24 seconds, I did count 25 seconds before this fix but this is not exact. Even in 1.3.9 we saw a result of 19 seconds. Just a thought, when search by address is restored without having to loop through all records, it might be quicker to get all addresses from the wallet and then search on each one individually.

bvbfan:
@Bushstar if you have faster wallet lookup feel free to push it.

monstrobishi:
@bvbfan I think you are more familiar with the new account history system, please continue to refine until we are able to reach reasonable speeds 🙂

Bushstar:
> @Bushstar if you have faster wallet lookup feel free to push it.

Apologies for any miscommunication, I do not think that currently there is a faster solution.

The quick fix for records not showing on specified address search in PR151 simply loops over every record which makes it as slow as searching on "mine", I think you were saying we need to migrate the storage to solve this issue. My suggestion is that, once this is done and the ability to quickly search on individual addresses is restored, we test using that to look up each wallet address in turn.

In 1.3.9 searching on an individual address is very fast as the first result returned is the address we want and also as soon as we hit an address we do not want the search is stopped, see below as I think we also want to add this.

https://github.com/DeFiCh/ain/blob/e1be1b77d74cc8df68a74ba978783397ad7c4778/src/masternodes/mn_rpc.cpp#L3129

Maybe this will be faster, maybe not, but let's test it out later when we can.

ShengguangXiao:
@bvbfan I think we should not convert to json during iterating the database, we can filter out the limited results and converted these limited results to json only, this can save memory and also speed up.

bvbfan:
The problem is we iterate over ldb in asc order, all iterators are forward ones, then in `listaccounthistory` we need result sorted in desc order, so we don't know how many records we should cache to apply limit on.

PR #153, opened 2020-12-24
Auto auth single output, single input and change to auth address

Bushstar:
PR to reduce the number of inputs and outputs generated by auto auth transactions and send change back to auth address.

Auto auth creates an auth TX with two outputs, one being a dust output to be used for authorisation and one to a change address that is not the auth address. The actual custom TX using this auto auth TX uses both outputs from the auth TX, the auth output for authorisation and the change address to cover the fee. Here you have two outputs created in the auth transaction and two outputs going to the custom TX.

There's actually a potential "Insufficient funds" issue here in auto auth, if the change output is too small or does not exist then the transaction will fail due to insufficient funds to cover the fee.

The PR here changes two things, the first is to reduce the number of auto auth inputs and outputs. If only a single auth output is required, set the output amount to be enough to cover a 1KB TX fee to avoid the above issue, after calling fund on the TX, which adds inputs and a change address, we combine the value of both auth output and change if present, then deletes that second value and set the destination of the remaining output to the auth address. This is then used in the custom TX as the single input.

This second change in this PR is to return change to specific addresses based on the call being used. Change destination behaviour is shown below.

**accounttoaccount** change to from address
**accounttoutxos** change to from address
**addpoolliquidity** on single auth change to from address
**createpoolpair** change to foundation address selected from wallet and used as auth
**createtoken** for DATs change to auth address
**minttokens** on single auth change to collateral address
**poolswap** change to from address
**removepoolliquidity** change to from address
**resignmasternode** change to owner address
**sendtokenstoaddress** on single auth change to from address
**setgov** change to foundation address selected from wallet and used as auth
**updatepoolpair** change to foundation address selected from wallet and used as auth
**updatetoken** change to collateral or foundation address used as auth

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -195,12 +195,15 @@ struct LockedCoinsScopedGuard
     }
 };
 
-static CMutableTransaction fund(CMutableTransaction & mtx, CWallet* const pwallet, CTransactionRef optAuthTx, bool lockUnspents = false) {
+static CMutableTransaction fund(CMutableTransaction & mtx, CWallet* const pwallet, CTransactionRef optAuthTx, CCoinControl* coin_control = nullptr, bool lockUnspents = false) {
     CAmount fee_out;
     int change_position = mtx.vout.size();
 
     std::string strFailReason;
     CCoinControl coinControl;
+    if (coin_control) {
+        coinControl = *coin_control;
+    }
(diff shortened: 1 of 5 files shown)


monstrobishi (approved):
This looks good, the rpc file is growing longer and longer, in a future PR would be nice to separate them out into their groupings (Account, Masternode, etc)

PR #155, opened 2020-12-29
Skip duplicated anchor rewards

ShengguangXiao:
Skip duplicated anchor rewards


monstrobishi:
looks like will cause a fork but should be ok (the old network won't actually fork it would just stall)

PR #158, opened 2021-01-05
Reduce getcustomtx length, rename JSON height and time keys

Bushstar:
Couple of updates requested by @ShengguangXiao on the getcustomtx PR to rename two JSON result keys and reduce the length of the getcustomtx function by moving out the switch statement.

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -3802,6 +3802,69 @@ UniValue sendtokenstoaddress(const JSONRPCRequest& request) {
 
 }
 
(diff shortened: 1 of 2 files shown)


ShengguangXiao (approved):

bvbfan:
I hate this strings keys, go back to my prev project i was defined them, i've not going you to do it, just i've not sure it's `blockHash` or `blockhash`, `blockTime` or `blocktime` :)

ShengguangXiao:
Actually bitcoin all using lower case `blocktime` and `blockhash` as the key, but when develop our fork some developers not following the same style, so now it is quite a mess. I suggest we can start use lower case as the key now.

PR #159, opened 2021-01-06
Keep owned account history split from the rest ones

bvbfan:
It introduces a new cli parameter `-acindex-mineonly` enabled by default, if `-acindex` is specified a warning message inform the user that full index will be used instead of mine only. In both operating modes mine history is stored separate for faster access. Now `listaccounthistory` is instant for mine, it still takes a bit more time when we search in the rest index table.

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -17,6 +17,7 @@
 
 #include <algorithm>
 #include <functional>
+#include <unordered_map>
 
 /// @attention make sure that it does not overlap with those in tokens.cpp !!!
 // Prefixes for the 'custom chainstate database' (customsc/)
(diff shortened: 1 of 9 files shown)


monstrobishi (approved):

monstrobishi:
Build is failing on CICD

`masternodes/mn_rpc.cpp:3297:71: error: reference to '_1' is ambiguous
        }, std::bind(insertEntry, _1, "sent"), std::bind(insertEntry, _1, "receive"));`

@bvbfan

bvbfan:
Yes, i keep track of what's happen since boost versions between my setup and Travis are different.

bvbfan:
Please note: Anyone using previous RC should reindex by yourself.

monstrobishi:
@bvbfan will it auto detect reindex from previous release version?

bvbfan:
Previous version yes, but not from previous RC since some key are different.

Bushstar (approved):

src/masternodes/masternodes.cpp
```diff
@@ -576,10 +595,34 @@ Res CRewardsHistoryStorage::AddBalance(CScript const & owner, DCT_ID poolID, uin
 bool CRewardsHistoryStorage::Flush()
 {
     if (acindex) {
+        auto wallets = GetWallets();
         for (const auto& diff : diffs) {
+            bool isMine = false;
             const auto& pair = diff.first;
-            SetRewardHistory({pair.first, height, pair.second}, diff.second);
+            for (auto wallet : wallets) {
```
ShengguangXiao, line 602: Can use `for (const auto& wallet : wallets)`

src/masternodes/masternodes.cpp
```diff
@@ -576,10 +595,34 @@ Res CRewardsHistoryStorage::AddBalance(CScript const & owner, DCT_ID poolID, uin
...
+                SetAllRewardHistory({pair.first, height, pair.second}, diff.second);
+            }
         }
     }
     return CCustomCSView::Flush();
 }
+
+isminetype IsMineCached(CWallet const & wallet, CScript const & script)
+{
+    static std::unordered_map<std::string, std::map<CScript, isminetype>> mineCached;
```
ShengguangXiao, line 618: Do we need to consider thread safety for this map?

src/masternodes/masternodes.cpp
bvbfan: Why that's shared_ptr?

src/masternodes/masternodes.cpp
bvbfan, line 618: `mineCached` in rpc calls (`mn_rpc.cpp`), `ApplyCustomTx` and `ConnectBlock` is guarded by `cs_main` we can marked as `GUARDED_BY` it's limited no clang only.

PR #163, opened 2021-01-08
Add custom rewards to pool pairs

Bushstar:
Add Clarke Quay hard fork logic and pool pair custom token reward feature. This allows any token to be specified as a reward to be paid from the pool owner address. Token rewards can be set by either during creation by `createpoolpair` and `updatepoolpair`. `getpoolpair` and `listpoolpair` will show this additional reward under the key `customRewards` only if it is set and the pool owner has enough balance to cover the reward. Multiple rewards can be added to a single pool, these will stop being paid out once the balance in the owner address is depleted and will continue again if topped up. `getcustomtx` has been updated to show the addition of `customRewards` to `getpoolpair` and `listpoolpair`.

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -1049,12 +1049,12 @@ Res ApplyCreatePoolPairTx(CCustomCSView &mnview, const CCoinsViewCache &coins, c
 
     auto tokenA = mnview.GetToken(poolPairMsg.idTokenA);
     if (!tokenA) {
-        return Res::Err("%s: token %s does not exist!", poolPairMsg.idTokenA.ToString());
+        return Res::Err("%s: token %s does not exist!", base, poolPairMsg.idTokenA.ToString());
     }
 
     auto tokenB = mnview.GetToken(poolPairMsg.idTokenB);
     if (!tokenB) {
-        return Res::Err("%s: token %s does not exist!", poolPairMsg.idTokenB.ToString());
+        return Res::Err("%s: token %s does not exist!", base, poolPairMsg.idTokenB.ToString());
     }
 
     if(pairSymbol.empty())
@@ -1120,9 +1120,14 @@ Res ApplyCreatePoolPairTx(CCustomCSView &mnview, const CCoinsViewCache &coins, c
     }
 
     if (!rewards.balances.empty()) {
-        // Remove empty reward amounts
+        // Check tokens exist and remove empty reward amounts
         for (auto it = rewards.balances.cbegin(), next_it = it; it != rewards.balances.cend(); it = next_it) {
             ++next_it;
+
+            auto token = pcustomcsview->GetToken(it->first);
+            if (!token) {
+                return Res::Err("%s: reward token %d does not exist!", base, it->first.v);
+            }
 
             if (it->second == 0) {
                 rewards.balances.erase(it);
@@ -1192,22 +1197,39 @@ Res ApplyUpdatePoolPairTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
         if (!rewards.balances.empty()) {
             UniValue rewardArr(UniValue::VARR);
 
-            for (const auto& reward : rewards.balances) {
-                if (reward.second > 0) {
-                    rewardArr.push_back(CTokenAmount{reward.first, reward.second}.ToString());
-                }
-            }
-
-            if (!rewardArr.empty()) {
+            // Check for special case to wipe rewards
+            if (rewards.balances.size() == 1 && rewards.balances.cbegin()->first == DCT_ID{std::numeric_limits<uint32_t>::max()}
+                    && rewards.balances.cbegin()->second == std::numeric_limits<CAmount>::max()) {
                 rpcInfo->pushKV("customRewards", rewardArr);
+            } else {
+                for (const auto& reward : rewards.balances) {
+                    if (reward.second > 0) {
+                        rewardArr.push_back(CTokenAmount{reward.first, reward.second}.ToString());
+                    }
+                }
+
+                if (!rewardArr.empty()) {
+                    rpcInfo->pushKV("customRewards", rewardArr);
+                }
             }
         }
     }
 
     if (!rewards.balances.empty()) {
-        // Remove empty reward amounts
+        // Check for special case to wipe rewards
+        if (rewards.balances.size() == 1 && rewards.balances.cbegin()->first == DCT_ID{std::numeric_limits<uint32_t>::max()}
+                && rewards.balances.cbegin()->second == std::numeric_limits<CAmount>::max()) {
+            rewards.balances.clear();
+        }
+
+        // Check if tokens exist and remove empty reward amounts
         for (auto it = rewards.balances.cbegin(), next_it = it; it != rewards.balances.cend(); it = next_it) {
             ++next_it;
+
+            auto token = pcustomcsview->GetToken(it->first);
+            if (!token) {
+                return Res::Err("%s: reward token %d does not exist!", base, it->first.v);
+            }
 
             if (it->second == 0) {
                 rewards.balances.erase(it);
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -1018,9 +1018,16 @@ Res ApplyCreatePoolPairTx(CCustomCSView &mnview, const CCoinsViewCache &coins, c
 
     CPoolPairMessage poolPairMsg;
     std::string pairSymbol;
+    CBalances rewards;
     CDataStream ss(metadata, SER_NETWORK, PROTOCOL_VERSION);
     ss >> poolPairMsg;
     ss >> pairSymbol;
+
+    // Read custom pool rewards
+    if (static_cast<int>(height) >= consensusParams.ClarkeQuayHeight && !ss.empty()) {
+        ss >> rewards;
+    }
+
     if (!ss.empty()) {
         return Res::Err("%s: deserialization failed: excess %d bytes", base,  ss.size());
     }
@@ -1079,6 +1086,20 @@ Res ApplyCreatePoolPairTx(CCustomCSView &mnview, const CCoinsViewCache &coins, c
         rpcInfo->pushKV("tradeable", token.IsTradeable());
         rpcInfo->pushKV("finalized", token.IsFinalized());
 
+        if (!rewards.balances.empty()) {
+            UniValue rewardArr(UniValue::VARR);
+
+            for (const auto& reward : rewards.balances) {
+                if (reward.second > 0) {
+                    rewardArr.push_back(CTokenAmount{reward.first, reward.second}.ToString());
+                }
+            }
+
+            if (!rewardArr.empty()) {
+                rpcInfo->pushKV("customRewards", rewardArr);
+            }
+        }
+
         return Res::Ok(base);
     }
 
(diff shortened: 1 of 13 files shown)


uzyn:
Great work @Bushstar 

Can you post some CLI usage examples and outputs from the RPC commands?

Bushstar:
@uzyn there's usage in the Python test file.

https://raw.githubusercontent.com/DeFiCh/ain/8a66b55fdd16315801132a60e960a9dd36e37b3a/test/functional/feature_custom_poolreward.py

```
poolpair_tx = self.nodes[0].createpoolpair({
    "tokenA": "GOLD#" + token_a,
    "tokenB": "SILVER#" + token_b,
    "commission": 0.001,
    "status": True,
    "ownerAddress": pool_collateral,
    "pairSymbol": "SILVGOLD",
    "customReward": ["1@" + token_a, "1@" + token_b]
})
```
```
updatepoolpair_tx = self.nodes[0].updatepoolpair({"pool": "1", "customReward": ["1@" + token_c]})
```

Output from `getcustomtx`, `listpoolpairs` and `getpoolpair` will have an additional line.
```
'customRewards': ['1.00000000@128', '1.00000000@129']
```

Bushstar:
By the way, there is no `rempoolreward`, to remove a reward funds can be moved out of the ownerAddress. Did test setting `0@token` in `updatepoolpair` to remove a token but this then fails in `DecodeAmounts` with an "Amount out of range" error.

src/validation.cpp
```diff
@@ -2304,13 +2304,19 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                 [&cache] (CScript const & owner, DCT_ID tokenID) {
                     return cache.GetBalance(owner, tokenID);
                 },
-                [&cache, &block] (CScript const & to, DCT_ID poolID, uint8_t type, CTokenAmount amount) {
+                [&cache, &block] (CScript const & to, CScript const & from, DCT_ID poolID, uint8_t type, CTokenAmount amount) {
+                    if (from != CScript()) {
+                        auto res = cache.SubBalance(from, amount);
```
bvbfan, line 2309: `SubBalance` will erase zeroes balance, is there needed to check them in `ApplyUpdatePoolPairTx`

src/masternodes/poolpairs.h
```diff
@@ -280,7 +285,28 @@ class CPoolPairView : public virtual CStorageView
...
+            if (nHeight >= Params().GetConsensus().ClarkeQuayHeight && rewards && !rewards->balances.empty()) {
+                for (auto it = rewards->balances.cbegin(), next_it = it; it != rewards->balances.cend(); it = next_it) {
+                    ++next_it;
+
+                    // Get token balance
+                    const auto balance = onGetBalance(pool.ownerAddress, it->first).nValue;
+
+                    // Make there's enough to pay reward otherwise remove it
+                    if (balance < it->second) {
+                        rewards->balances.erase(it);
```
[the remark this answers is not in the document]
Bushstar: I think this would be up to the pool owner to check and we may consider this not to actually be a problem but intentional, for example we no longer want to pay out that token so let it deplete or moved the token out of the owner address. If we do log this then we are logging for everyone, if we do log it then I would suggest adding it to a debug category to keep it out of regular logging.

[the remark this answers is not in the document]
bvbfan: You can check owner is mine for wallet, then log? I think it can notable if you want to pay but you don't know you run out of money.

src/validation.cpp
Bushstar: The `ApplyUpdatePoolPairTx` check that we do not create a `0@token` reward as `WriteBy<Reward>` in the `CPoolPairView`. `SubBalance` will subtract from the pool owner address and will not impact the `1@token` record.

src/masternodes/poolpairs.h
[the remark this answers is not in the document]
Bushstar: That's going to create at least one log entry per block, we would not want to trigger it for each reward payment for sure. The pool owner can just run the following.

`getaccount poolOwner`

My take on it is that if it ran out of tokens then it was intentional, that's how you would stop payment. Let's get someone to take the deciding vote @uzyn :slightly_smiling_face:

ShengguangXiao (approved):

src/validation.cpp
bvbfan: Got'cha. So you can remove it before wirte? I mean we don't need to keep zeroes.

Bushstar: Correct. So we do not write zero amount rewards.

src/masternodes/poolpairs.h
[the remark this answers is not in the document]
bvbfan: I'm OK with explanation. Another option is to update ppol custom reward by disable / erase that reward, then owner can enable it again when update its balance

monstrobishi (changes requested):
This is excellent work.

One minor but essential comment, after careful review I think that if someone was to pass in `customRewards: []` to updatepoolpair in an attempt to stop the rewards, this will silently do nothing (as ApplyUpdatePoolPairTx will skip it and not write anything). It may be challenging due to the way the default values work but there should be some way we can differentiate and address this scenario as I believe it would be a common usecase here, would also like to see a test added for this case.

src/masternodes/mn_checks.cpp
```diff
@@ -1141,6 +1187,39 @@ Res ApplyUpdatePoolPairTx(CCustomCSView & mnview, CCoinsViewCache const & coins,
         rpcInfo->pushKV("commission", ValueFromAmount(commission));
         rpcInfo->pushKV("status", status);
         rpcInfo->pushKV("ownerAddress", ownerAddress.GetHex());
+
+        // Add rewards here before processing them below to avoid adding current rewards
+        if (!rewards.balances.empty()) {
```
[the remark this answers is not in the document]
monstrobishi: perhaps in mn_rpc we can just automatically add a '0@0' when an empty array is passed, this will "disable" rewards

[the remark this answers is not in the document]
Bushstar: Empty array will not do anything even though the use may expect it to. On your suggestion from Slack will add feedback to this effect from the updatetoken RPC call. Currently cannot remove tokens at all, will add '0@token' in customRewards to allow token removal.

src/masternodes/poolpairs.h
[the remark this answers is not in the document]
Bushstar: Did actually have this functionality at one point to remove token, '0@token' actually removed the token from rewards, however RPC call creates errors from DecodeAmounts so I removed it from RPC and ApplyUpdatePoolPairTx, did not make sense to support a feature that could not be accessed. Will restore and work round this DecodeAmounts error.

src/validation.cpp
```diff
@@ -2321,7 +2321,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
 
             auto res = cache.SubCommunityBalance(CommunityAccountType::IncentiveFunding, distributed);
             if (!res.ok)
-                throw std::runtime_error(strprintf("Pool rewards: can't update community balance: %s. Block %ld (%s)", res.msg, block.height, block.GetHash().ToString()));
+                LogPrintf("Pool rewards: can't update community balance: %s. Block %ld (%s)", res.msg, block.height, block.GetHash().ToString());
```
bvbfan, line 2324: So we still distribute rewards without funding, is that right?

Bushstar: Check the line here. From how I see it, the min is chosen, eventually the minimum will be 0 from the community fund and will be the selected amount for block rewards. At that point no rewards will be paid out, but SubCommunityBalance will fail and without the change above would crash the network!

https://github.com/DeFiCh/ain/blob/438ecbb2ef8f368193a1a8ba6914ffaf02deb2ee/src/validation.cpp#L2298

bvbfan: If it's 0 will be good case, if it's not, rewards are distributed to all pool pairs without checking is there needed funds. See DistributeRewards, poolReward s distributed to all without checking is there funds for all.

Bushstar: It gets the balance for DistributeRewards from the community fund from the line I linked above, basically min of either reward or community fund, so when community fund gets below the reward the remainder is paid out, so the balances is checked before rewards are paid. When the community fund gets to 0 DistributeRewards will have nothing to pay.

Bushstar:
```
            CAmount poolsBlockReward = std::min(
                                           cache.GetCommunityBalance(CommunityAccountType::IncentiveFunding),
                                           var->dailyReward / (60*60*24/chainparams.GetConsensus().pos.nTargetSpacing) // 2880
                                                );
```
poolsBlockReward  is first argument to DistributeRewards as yieldFarming. The following poolReward is then calculated in DistributeRewards which will be 0 when GetCommunityBalance runs out.
```
CAmount const poolReward = yieldFarming * pool.rewardPct / COIN; 
```

bvbfan: And then in `ForEachPoolShare` you reward all pools by this reward `N * poolReward` you don't know it has enough fund for that, no?

Bushstar: We do as it uses poolReward which is calculated from above, when 0 it does not run, when less than the DFI reward it pays out whatever is left in the fund.

From ForEachPoolShare:
```
                if (poolReward) {
                    CAmount providerReward = static_cast<CAmount>((arith_uint256(poolReward) * arith_uint256(liquidity) / arith_uint256(pool.totalLiquidity)).GetLow64());
                    if (!newRewardCalc) {
                        providerReward = poolReward * liqWeight / PRECISION;
                    }
                    if (providerReward) {
                        onTransfer(provider, poolId, uint8_t(RewardType::Rewards), {DCT_ID{0}, providerReward}); //can throw
                        totalDistributed += providerReward;
                    }
                }
```

Bushstar: We do as it uses poolReward which is calculated from above, when 0 it does not run, when less than the DFI reward it pays out whatever is left in the fund.

From ForEachPoolShare:
```
                if (poolReward) {
                    CAmount providerReward = static_cast<CAmount>((arith_uint256(poolReward) * arith_uint256(liquidity) / arith_uint256(pool.totalLiquidity)).GetLow64());
                    if (!newRewardCalc) {
                        providerReward = poolReward * liqWeight / PRECISION;
                    }
                    if (providerReward) {
                        onTransfer(provider, poolId, uint8_t(RewardType::Rewards), {DCT_ID{0}, providerReward}); //can throw
                        totalDistributed += providerReward;
                    }
                }
```

Bushstar: We work out how much each pool gets with...
```
ForEachPoolPair([&] (DCT_ID const & poolId, CPoolPair pool) {
    CAmount const poolReward = yieldFarming * pool.rewardPct / COIN; 
```
Then distribute that per share by weight of liquidity.
```
ForEachPoolShare([&] (DCT_ID const & currentId, CScript const & provider) {
    if (poolReward) {
        CAmount providerReward = static_cast<CAmount>((arith_uint256(poolReward) * arith_uint256(liquidity) / arith_uint256(pool.totalLiquidity)).GetLow64());
```
I'll take a look and check for test this behaves as expected, if there is none then perhaps we should code some to make sure the community fund runs out without a fuss.

bvbfan: The problem is liquidity is just provider balance that's not guaranteed community balance has enough funds.

PR #166, opened 2021-01-13
Add ability for multiple operators per node

bvbfan:
It starts a thread per master node operator to stake.


monstrobishi:
Quick question @bvbfan will this affect consensus? I assume no but just checking because it touches some parts of code that have consensus

bvbfan:
Should not affect consensus.

src/init.cpp
```diff
@@ -1915,73 +1915,89 @@ bool AppInitMain(InitInterfaces& interfaces)
...
-                    }
-                }
-                if (!found) {
-                    LogPrintf("Error: masternode operator private key not found\n");
-                    return false;
+        std::set<std::string> operatorsSet;
+        bool atLeastOneRunningOperator = false;
+        auto const operators = gArgs.GetArgs("-masternode_operator");
+
+        for (auto const & op : operators) {
```
monstrobishi, line 1928: How does the setting of multiple operators look like in the conf now?

Can we have backwards compatibility with the current setting?

bvbfan: Backward compatibility is shown in functional test. The settings is same as cli settings you just specify multiple options
`defid -masternode_operator=<fist-one> -masternode_operator=<second-one>`

ShengguangXiao:
So the masternode_owner parameter has no use already?

bvbfan:
It should be used with `-gen` it's not been used till now in tests, probably used in `AmIOperator`

ShengguangXiao (approved):

PR #168, opened 2021-01-13
Fix logic balances less than other ones

bvbfan:
Could you guys can verify it's finally the right fix. When we compare balances when some of the right balances is higher than left one, it's mean left is less than right one.

### DIFF
--- a/src/masternodes/balances.h
+++ b/src/masternodes/balances.h
@@ -108,10 +108,10 @@ struct CBalances
         return str;
     }
 
-    static CBalances Sum(std::vector<CTokenAmount> tokens) {
-        CBalances res{};
-        for (const auto& t : tokens) {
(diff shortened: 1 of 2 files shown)


ShengguangXiao:
Can add some unittest cases to verify the comparison results?

bvbfan:
`feature_any_accounts_to_accounts` use it.

ShengguangXiao:
> `feature_any_accounts_to_accounts` use it.

It will be better if have some C++ unittest directly do the comparison.

bvbfan:
I've going to remove `GetAllTokensAmount` sum all amounts despite token is quite useless, about me. Also it can overflow int, also i don't see how it works correct when you have [`1@1`, `8@2`] and [`3@1`,  `5@2`] so sorted depend of sum1 and sum2.

PR #170, opened 2021-01-15
Fix sendtokentoaddress account selection mode

bvbfan:
It fixes the account selecting algorithm by sorting every token at one then calculations are applied.


ShengguangXiao (approved):

bvbfan:
Fix `crumbs` in asc order, `pie` in desc order.

PR #180, opened 2021-01-19
Do not allow send to non-self addresses by default

bvbfan:


monstrobishi:
oh I see, so sending to own addresses is allowed by default, so this shouldn't cause any issues with the app internally

src/masternodes/mn_rpc.cpp
```diff
@@ -486,6 +486,26 @@ static std::vector<CTxIn> GetAuthInputsSmart(CWallet* const pwallet, int32_t txV
     return result;
 }
 
+static CAccounts DecodeRecipientsDefaultInternlal(CWallet* const pwallet, UniValue const& values) {
```
monstrobishi, line 489: typo internlal

src/masternodes/masternodes.cpp
```diff
@@ -618,17 +620,34 @@ bool CRewardsHistoryStorage::Flush()
     return CCustomCSView::Flush();
 }
 
-isminetype IsMineCached(CWallet const & wallet, CScript const & script)
+struct CCacheInfo
```
monstrobishi, line 623: what are these changes for? is it an optimisation? will this speed up indexing too?

src/masternodes/mn_rpc.cpp
```diff
@@ -2208,6 +2228,7 @@ UniValue utxostoaccount(const JSONRPCRequest& request) {
                         {
                             {"address", RPCArg::Type::STR, RPCArg::Optional::NO, "The defi address is the key, the value is amount in amount@token format. "
                                                                                  "If multiple tokens are to be transferred, specify an array [\"amount1@t1\", \"amount2@t2\"]"},
+                            {"enable_external_dfi_token_tx", RPCArg::Type::BOOL, RPCArg::Optional::OMITTED, "By default we don't allow sending tokens outside dfi addresses"},
```
ShengguangXiao, line 2231: The parameter is added but never used?

src/masternodes/masternodes.cpp
ShengguangXiao: I guess the change is to dynamically update the cache if the wallet changed or unloaded.

ShengguangXiao:
Can add unit test for `enable_external_dfi_token_tx` is false?

src/masternodes/masternodes.cpp
bvbfan: Update mine cache when you insert private key, also clears cache when wallet is unloaded.

bvbfan:
> Can add unit test for `enable_external_dfi_token_tx` is false?

When it's not specified it's false, most of cases.

bvbfan:
Fix typo.

ShengguangXiao (approved):

monstrobishi:
tests passing I will merge

uzyn:
This is a bad merge, please rollback this. 

- Bad interface design: it sholdn't be part of the `to` obj.
- And why are we blocking general token send. The original issue was only for DFI token specifically.

PR #182, opened 2021-01-19
Fix gettokenbalances

bvbfan:
Now `gettokenbalances` and `getaccount` returns different token balances. It's due to balance to empty address, probably poolswap to empty address or so. `listaccounts` shows it on fist place.


ShengguangXiao (approved):

monstrobishi (approved):

uzyn:
@bvbfan @ShengguangXiao @monstrobishi 

This is clearly a missed unit test. Please include unit test along with the fix.

PR #184, opened 2021-01-21
Minimize custom tx logs

bvbfan:
It can speedup sync since writing to stdout as well in file is slow.


monstrobishi:
Can you also reduce frequency or remove logging of duplicate anchor payment, this one is flooding testnet logs

bvbfan:
@ShengguangXiao  remove duplicate anchor reward log, actually it will be good to be shown only once but  don't think it worth.

ShengguangXiao:
> @ShengguangXiao i remove duplicate anchor reward log, actually it will be good to be shown only once but don't think it worth.

I think can be removed, I added the log to verify the logic is working, now the verification is done, so safe to remove.

ShengguangXiao (approved):

monstrobishi:
travis failed saying argument conversion failed for reward address

PR #188, opened 2021-01-23
Try blocking only dfi token

monstrobishi:
Changed to block only DFI token and took out optional parameter

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -490,18 +490,12 @@ static CAccounts DecodeRecipientsDefaultInternal(CWallet* const pwallet, UniValu
     UniValue recipients(UniValue::VOBJ);
     bool allowSendingToNotOwnAddresses = false;
     for (const auto& key : values.getKeys()) {
-        if (key == "enable_external_dfi_token_tx") {
-            allowSendingToNotOwnAddresses = values[key].getBool();
-        } else {
-            recipients.pushKV(key, values[key]);
-        }
+        recipients.pushKV(key, values[key]);
     }
     auto accounts = DecodeRecipients(pwallet->chain(), recipients);
-    if (!allowSendingToNotOwnAddresses) {
-        for (const auto& account : accounts) {
-            if (IsMineCached(*pwallet, account.first) != ISMINE_SPENDABLE) {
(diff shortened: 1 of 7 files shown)


src/masternodes/mn_rpc.cpp
```diff
@@ -490,18 +490,12 @@ static CAccounts DecodeRecipientsDefaultInternal(CWallet* const pwallet, UniValu
     UniValue recipients(UniValue::VOBJ);
     bool allowSendingToNotOwnAddresses = false;
```
ShengguangXiao, line 491: This parameter can be removed

ShengguangXiao:
I will test with this PR first.

monstrobishi:
test failing will revise

ShengguangXiao:
Test with `accounttoaccount` result:
Send DFI to self                                 Can
Send DFI to other account                Cannot
Send other token to other account   Can

src/masternodes/mn_rpc.cpp
```diff
@@ -488,20 +488,13 @@ static std::vector<CTxIn> GetAuthInputsSmart(CWallet* const pwallet, int32_t txV
...
 static CAccounts DecodeRecipientsDefaultInternal(CWallet* const pwallet, UniValue const& values) {
     UniValue recipients(UniValue::VOBJ);
-    bool allowSendingToNotOwnAddresses = false;
     for (const auto& key : values.getKeys()) {
-        if (key == "enable_external_dfi_token_tx") {
-            allowSendingToNotOwnAddresses = values[key].getBool();
-        } else {
-            recipients.pushKV(key, values[key]);
-        }
+        recipients.pushKV(key, values[key]);
```
bvbfan, line 492: You can remove it, just rename values to recipients

test/functional/feature_mine_cached.py
```diff
@@ -46,19 +46,9 @@ def run_test(self):
...
-
-        self.nodes[0].sendtokenstoaddress({}, to)
-
-        self.nodes[0].unloadwallet('')
-        self.nodes[0].loadwallet('')
-
-        self.nodes[0].sendtokenstoaddress({}, to)
-
-
 if __name__ == '__main__':
```
bvbfan, line 53: Why it's removed?

monstrobishi: You can fix that later I'm just doing a quick patch for @uzyn the basic token functionality is already handled by many other tests so it's not needed here

PR #189, opened 2021-01-23
Add exclude_custom_tx option in listtransactions rpc

ShengguangXiao:
Add exclude_custom_tx option in listtransactions rpc


uzyn:
Oh I didn't expect it to be an option, I wanted it actually as part of the output.

If it's an option, this is even better, can you ensure that you can **exclude** all customTx transactions? I need exclusions rather than inclusions.

Thanks.

monstrobishi (approved):

monstrobishi:
Any chance we can get a unit test for this

ShengguangXiao:
> Any chance we can get a unit test for this

Yes, let me add a unittest for a custom tx

PR #194, opened 2021-01-25
Fix auto select accounts

bvbfan:
Fixes https://github.com/cakedefi/defichain-private/issues/300


monstrobishi (approved):

monstrobishi:
nice this is much better. can we have a unit test that compares output to list accounts

ShengguangXiao (approved):

PR #197, opened 2021-01-26
Add AutoAuthPrep meta data to auto auth TXs

Bushstar:
This PR adds meta data to the auto auth TXs so they can be filtered out on listtransactions and report their TX type in getcustomtx.

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -357,6 +357,14 @@ CTransactionRef CreateAuthTx(CWallet* const pwallet, std::set<CScript> const & a
     CMutableTransaction mtx(txVersion);
     CCoinControl coinControl;
 
+    // Encode
+    CDataStream markedMetadata(DfTxMarker, SER_NETWORK, PROTOCOL_VERSION);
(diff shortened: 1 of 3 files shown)


monstrobishi (approved):

src/masternodes/mn_rpc.cpp
```diff
@@ -4113,6 +4121,10 @@ static bool GetCustomTXInfo(const int nHeight, const CTransactionRef tx, CustomT
         case CustomTxType::AnyAccountsToAccounts:
             res = ApplyAnyAccountsToAccountsTx(mnview_dummy, ::ChainstateActive().CoinsTip(), *tx, nHeight, metadata, Params().GetConsensus(), true, &txResults);
             break;
+        case CustomTxType::AutoAuthPrep:
+            res.ok = true;
+            res.msg = "AutoAuth";
```
bvbfan, line 4126: `ToString(guess)` for consistency.

Bushstar: Makes sense. Will change next time I'm in that area :)

PR #198, opened 2021-01-26
[WIP] Reconsider blocks prior last checkpoint

bvbfan:


src/validation.cpp
```diff
@@ -2965,6 +2965,18 @@ bool CChainState::ActivateBestChainStep(CValidationState& state, const CChainPar
                         InvalidChainFound(vpindexToConnect.front());
                     }
                     state = CValidationState();
+                    if (pindexConnect == pindexMostWork) {
```
monstrobishi, line 2968: Can I ask what's this checking for?

monstrobishi: Can we trigger this off checkpoint not matching instead or is this an easier way to check

bvbfan: Just try to do it once, when we are at most work.

ShengguangXiao (approved):

PR #201, opened 2021-01-27
Prevent downgrade to genesis block on wrong chain

bvbfan:
I was able to test it on regtest, it can revert chain to genesis and regenerate it to top. If you interested to test in regtest
```
auto it = checkpoints.begin();
if (it != checkpoints.end())
```
should be condition instead of lower_bound, you should keep test data, then verify by debug.log on node0.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2924,20 +2924,25 @@ bool CChainState::ActivateBestChainStep(CValidationState& state, const CChainPar
     // Disconnect active blocks which are no longer in the best chain.
     bool fBlocksDisconnected = false;
     DisconnectedBlockTransactions disconnectpool;
-    while (m_chain.Tip() && m_chain.Tip() != pindexFork) {
-        if (!DisconnectTip(state, chainparams, &disconnectpool)) {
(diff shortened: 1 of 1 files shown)


ShengguangXiao:
> I was able to test it on regtest, it can revert chain to genesis and regenerate it to top. If you interested to test in regtest
> 
> ```
> auto it = checkpoints.begin();
> if (it != checkpoints.end())
> ```
> 
> should be condition instead of lower_bound, you should keep test data, then verify by debug.log on node0.

Can you share how you revert chain to genesis block? I also want to try.

bvbfan:
We can add a regtest only rpc function to change checkpoints (to add a checkpoint to 1 height) then we can read the log by the node, if you're interested to have an real unit test. If not 
```
auto it = checkpoints.lower_bound(pindexConnect->nHeight);
if (it != checkpoints.begin()) {
    auto index = LookupBlockIndex((--it)->second);
```
to
```
auto it = checkpoints.begin();
if (it != checkpoints.end()) {
    auto index = LookupBlockIndex(it->second);
```
`./test_runner.py p2p_segwit2 -k`

PR #204, opened 2021-01-28
[Wallet] Performance optimizations, memory usage revamp

bvbfan:
It does not contain braking changes, it redesigns containers usage to prevent memory overhead. I will plan to reindex over this branch for testing. @uzyn will notify you when all looks good to be tested on microservice node to verify it will reduce memory usage. @ShengguangXiao can also reindex over this branch on main and testnet, observe memusage as well.


bvbfan:
I bit more aggressive refactor of dead wallet code. Strip out unused code, minimize CWalletTx size.

uzyn:
@bvbfan the community found something: https://www.reddit.com/r/defiblockchain/comments/l765tb/fix_for_ram_bug_and_corrupt_blockchain/

src/wallet/wallet.cpp
```diff
@@ -100,13 +99,9 @@ static std::set<std::string> g_unloading_wallet_set;
 // Custom deleter for shared_ptr<CWallet>.
 static void ReleaseWallet(CWallet* wallet)
 {
-    // Unregister and delete the wallet right after BlockUntilSyncedToCurrentChain
-    // so that it's in sync with the current chainstate.
     const std::string name = wallet->GetName();
     wallet->WalletLogPrintf("Releasing wallet\n");
-    wallet->BlockUntilSyncedToCurrentChain();
     wallet->Flush();
-    wallet->m_chain_notifications_handler.reset();
```
ShengguangXiao, line 109: Why these two lines are removed?

src/wallet/wallet.h
```diff
@@ -679,6 +680,28 @@ struct CoinSelectionParams
...
+            bm::tag<ByHash>, bm::const_mem_fun<CWalletTx, const uint256&, &CWalletTx::GetHash>
+        >,
+        bm::ordered_non_unique<
+            bm::tag<ByOrder>, bm::member<CWalletTx, int64_t, &CWalletTx::nOrderPos>
+        >,
+        bm::ordered_non_unique<
+            bm::tag<ByTime>, bm::member<CWalletTx, unsigned int, &CWalletTx::nTimeReceived>
+        >
+    >
+> mapWalletType;
```
ShengguangXiao, line 703: The type change to `multiIndexWalletType` will be more readable.

src/wallet/walletdb.h
```diff
@@ -185,7 +185,7 @@ class WalletBatch
             return false;
         }
         m_database.IncrementUpdateCounter();
-        if (m_database.nUpdateCounter % 1000 == 0) {
+        if (m_database.nUpdateCounter % 100 == 0) {
```
ShengguangXiao, line 188: Can we define 100 as a const member variable?

src/wallet/wallet.cpp
bvbfan: It makes wallet unstable. They are in `UnloadWallet`

src/wallet/wallet.h
bvbfan: Underlying type multi_index isn't needed, about me.

src/wallet/walletdb.h
bvbfan: Will revert it. Let stay 1000.

bvbfan:
Round 2 of cleaning wallet dead code. Now it uses mineCached as well.

src/wallet/walletdb.h
ShengguangXiao: Now the value still 100

ShengguangXiao:
I test the change, reindex blocks and wallets works fine, because my wallet is small, so didn't see much ram difference as before.

monstrobishi:
@bvbfan there is a problem in travis in `rpc_fundrawtransaction.py              | ✖ Failed  | 59 s`

bvbfan:
It can be tested on systems with many addresses and/or txs, wallet should be more responsive, full sync uses same memory on my end.

src/sync.h
```diff
@@ -318,4 +319,22 @@ struct SCOPED_LOCKABLE LockAssertion
...
+    std::atomic_bool& lock;
+public:
+    CLockFreeGuard(std::atomic_bool& lock) : lock(lock)
+    {
+        bool desired = false;
+        while (!lock.compare_exchange_weak(desired, true,
+                                           std::memory_order_release,
+                                           std::memory_order_relaxed))
+            desired = false;
+    }
```
ShengguangXiao, line 333: This loop will consume a lot of CPU resources.

monstrobishi:
@bvbfan still have a failing test here: wallet_basic.py", line 367

src/sync.h
bvbfan: Force pushed a fix.

bvbfan:
Tested on full sync, it should be not a wip anymore.

PR #205, opened 2021-01-29
Do not read stream out of its bounds in FindByte

bvbfan:
It address the high memory usage on `-reindex`, It can read over and over again the file and accept same blocks which can lead to high memory usage. If that's not help i prefer to just purge old files on `-reindex`

### DIFF
--- a/src/streams.h
+++ b/src/streams.h
@@ -853,10 +853,13 @@ public:
     void FindByte(char ch) {
         while (true) {
             if (nReadPos == nSrcPos)
-                Fill();
-            if (vchBuf[nReadPos % vchBuf.size()] == ch)
-                break;
-            nReadPos++;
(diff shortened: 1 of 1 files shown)


src/streams.h
```diff
@@ -853,10 +853,13 @@ class CBufferedFile
     void FindByte(char ch) {
         while (true) {
             if (nReadPos == nSrcPos)
-                Fill();
-            if (vchBuf[nReadPos % vchBuf.size()] == ch)
-                break;
-            nReadPos++;
+                if (!Fill())
+                    throw std::ios_base::failure("CBufferedFile::FindByte: fail to read");
```
ShengguangXiao, line 857: Can you kindly use bracket here? It will be much easier to read, and also align the code.

PR #207, opened 2021-02-01
Do not search in wallet for custom tx type

bvbfan:
We don't need to search in wallet for custom tx type, it's pointless.


ShengguangXiao:
Do we need a unit test for this?

PR #211, opened 2021-02-02
Reduce masternode collateral amount

bvbfan:
It's used in POS as well.
Fixes https://github.com/cakedefi/defichain-private/issues/339

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -236,6 +236,7 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
 //
 //            uint256 masternodeID = testMasternodeKeys.begin()->first;
 //            uint32_t mintedBlocks(0);
+//            int64_t creationHeight;
 //            CKey minterKey;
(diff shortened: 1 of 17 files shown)


bvbfan:
Rebase to 1.5.x should be OK, now.

monstrobishi:
Can I confirm that existing masternodes that staked 1 million will be able to continue staking 1 million until they unstake and stake again? We can check creation height against the hardfork height

bvbfan:
It makes sense.

bvbfan:
Force push changes to keep node created before Dakota height to keep their stakes.

src/test/miner_tests.cpp
```diff
@@ -250,13 +251,14 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
 //                BOOST_CHECK(nodePtr && nodePtr->IsActive(tip->height));
 //
 //                mintedBlocks = nodePtr->mintedBlocks;
+//                creationHeight = int64_t(nodePtr->creationHeight);
 //            }
```
ShengguangXiao, line 255: Can remove all these commented code

bvbfan: Probably we can enabled it in the future?

PR #214, opened 2021-02-04
Fix swap max price calculation

bvbfan:
Fixes https://github.com/cakedefi/defichain-private/issues/404

### DIFF
--- a/src/masternodes/mn_rpc.cpp
+++ b/src/masternodes/mn_rpc.cpp
@@ -2862,33 +2862,10 @@ void CheckAndFillPoolSwapMessage(const JSONRPCRequest& request, CPoolSwapMessage
             CAmount maxPrice = AmountFromValue(metadataObj["maxPrice"]);
             poolSwapMsg.maxPrice.integer = maxPrice / COIN;
             poolSwapMsg.maxPrice.fraction = maxPrice % COIN;
-        }
-        else {
-            // This is only for maxPrice calculation
-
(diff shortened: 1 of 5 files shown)


monstrobishi (approved):

src/masternodes/mn_rpc.cpp
```diff
@@ -2877,9 +2877,18 @@ void CheckAndFillPoolSwapMessage(const JSONRPCRequest& request, CPoolSwapMessage
...
             auto resA = pool.reserveA;
             auto resB = pool.reserveB;
-            if (poolSwapMsg.idTokenFrom != pool.idTokenA) {
-                std::swap(resA, resB);
+
+            // NOTE it has a bug prior Dakota hardfork
+            if (height < Params().GetConsensus().DakotaHeight) {
+                if (poolSwapMsg.idTokenFrom != pool.idTokenA) {
+                    std::swap(resA, resB);
+                }
```
monstrobishi, line 2885: We don't need hardfork check on RPC side

monstrobishi: If maxPrice is null just set both to int MAX, then it has no protection @uzyn asked, then the app can pass in max price

bvbfan: Done.

ShengguangXiao (approved):

ShengguangXiao:
`feature_poolswap.py` failed on travis.

PR #217, opened 2021-02-07
Create poolpair need one of the token authentication

ShengguangXiao:
Create poolpair need one of the token authentication


monstrobishi:
@ShengguangXiao this only checks auth at RPC level right?

@uzyn we need this at consensus level I'm assuming?

bvbfan:
We need at consensus level check as well.

PR #218, opened 2021-02-07
Anchor refactor

Bushstar:
Change anchoring to create new auth and confirm teams every 120 blocks. Information about anchor team is embedded into the anchor field which held the anchor MN IDs previously, a single entry is now present that holds an identifiable marker, anchor creation height to find the team at that height and part of the blockhash of that creation height to make sure we are still on the same chain. For the confirm team the miner looks up the team at the current height and checks if there are any confirms by that team that meet quorum.

The rationale for changing the teams every 120 blocks is to make sure that the teams are current, with teams fixed at the point of anchor creation in the past, as those anchors got old the less likely that the MNs in the teams would still be online, this could lead to a stall in the anchoring system. MNs used in the anchor teams are selected from anchors who created blocks in the last two weeks to not select non-active MNs for use in teams.

Anchors are still chosen from a depth of at least anchoringFrequency (15 on mainnet), this defines the anchor creation height, but the anchor data now goes back until they are three hours deep into the chain, then further until they are a on a multiple of anchoringFrequency.

Confirms now include the anchor data hash and Bitcoin anchor block height. The former allows the miner and parser of the block to validate that the anchor belongs to the active chain, if not the miner will not include it and the parser would reject the block. To support this new confirm data a new anchor coinbase TX is created, this includes a new data marker to identify this TX type and a new anchor TX parsing function used when parsing block data, it is this new parsing function that makes use of the new team structure and block hash to validate the anchor.

In addition to a new TX type new P2P message and inventory types are introduced to support the extra fields in the confirm data. The old types are no longer present in the code and old messages would be ignored. Having the extra data in the P2P messages allows us to check that the anchor data height and hash relate to the active chain otherwise they fail.

MNs will now run SPV by default, non-MNs need to enable SPV manually. We expect a lot of new MNs to be set up with the reduction in coin rewards, if SPV is not enabled by default then we might expect the MNs that are chosen to be part of the teams to not be available stalling the system or at least making its functionality intermittent.

Anchoring no longer has any impact on the consensus in regards to which chain a node connects to, previously a node would switch to an anchored chain. It should be noted that the old anchoring system relied on the initial team to do the validation work and that all subsequent steps skipped further validation, to the point where quorum was not checked when reading anchor data from Bitcoin, you could actually add an anchor with no sigs at all. Combine this with anchoring being part of consensus and you could make the chain switch to an old ignore fork to reverse transactions. Team quorum is now checked when reading anchor data from Bitcoin and the anchors are fully validated on every part of the anchor journey.

Anchor auths were only validated via P2P by SPV nodes, otherwise they would not be relayed. This is no longer the case, certain validation is skipped on non-SPV nodes to allow them to relay the anchor, however anchor auth validation now includes for all nodes checks via the new ContextualValidateAnchor function that checks the height exist in chain, that the blockhash is correct for that height, that the anchor is a new type anchor, the anchor meets the time and anchoringFrequency requirements and that the creation height hash prefix matches. The SPV related checks that might be skipped on P2P are still applied when creating an anchor using the anchor auths, really an SPV node should not have anchor auths that fail SPV checks in the anchor auth index but this is a belt and braces type approach. Anchor creation in addition to checking the anchor auth again is also updated to get the team from the anchor data and find a set of anchor auths that meet a quorum.

Reading anchor data from Bitcoin no longer just adds those anchors to the list of anchors after checking sigs, note that quorum was never checked as mentioned above. Anchors are now fully validated but as this requires chain context which we may not have at the time of SPV sync, we add these anchors to the new anchor pending DB storage. On every block update the pending anchors are checked to see if the anchor is now in chain content and ContextualValidateAnchor is then used and after team quorum and sigs are checked.

The anchor confirm data embedded into the chain is now stored and can be viewed using listanchors, this allows all users to see the anchors to use in manual review, the data presented to users is the anchor height, anchor hash, anchor reward address, anchor reward TX hash, Bitcoin anchor height, Bitcoin reward TX hash and the confirm sign hash. RPC calls to show anchor data to SPV nodes has also been increased to show the new additional fields and other fields that were not shown like previous anchor field, these changes allow better manual validation.

Misc: New anchor logging category added, enable with -debug=anchoring. Anchor auth and confirm data now supports multi-operator masternodes. getmasternodeblocks RPC call added to show blocks generated by the specified masternode. getanchorteams RPC call added to show the current team or team at the specified height. Node can now shutdown is started offline with SPV enabled. spv_listanchorspending RPC call to show pending anchors. spv_listanchorsunrewarded RPC call added to show unrewarded anchors.


test/functional/feature_anchors.py
```diff
@@ -150,98 +134,5 @@ def run_test(self):
...
-        print ("Node2: Deactivating anchor, no peers")
-        disconnect_nodes(self.nodes[1], 2)
-        self.nodes[2].spv_setlastheight(0)
-        time.sleep(2)
-        # print ("Anc 2: ", self.nodes[2].spv_listanchors())
-        self.dumpheights()
-        self.dumphashes([2], 14)
-        assert_equal(self.nodes[2].getblockcount(), chain1+20)
-
-
```
monstrobishi, line 245: Any downside of removing these tests? Do we have something to replace them?

Bushstar: Those test no longer apply, they tested how anchors applied to consensus, in that the client would connect to the anchored chain. That functionality has been removed from the anchoring system.

monstrobishi (approved):
Looks good, @Bushstar ok to merge?

src/masternodes/anchors.cpp
```diff
@@ -217,17 +267,54 @@ CAnchor CAnchorAuthIndex::CreateBestAnchor(CTxDestination const & rewardDest) co
...
-                for (uint32_t i = 0; i < quorum && it0 != it1; ++i, ++it0) {
-                    LogPrintf("auths: pick up %d, %s, %s\n", it0->height, it0->blockHash.ToString(), it0->GetHash().ToString());
-                    /// @todo do we need for an extra check of the auth signature here?
-                    freshestConsensus.push_back(*it0);
+
+                uint32_t validCount{0};
+                it0Copy = it0;
+                for (uint32_t i{0}; i < quorum && it0Copy != it1; ++i, ++it0Copy) {
+                    // ValidateAuth called here performs extra checks with SPV enabled.
+                    if (ValidateAuth(*it0)) {
```
bvbfan, line 305: It should be `it0Copy` ?

bvbfan (changes requested):

Bushstar:
@monstrobishi please do not merge. Let the C++ team review it first and I want to add another test. Need to test mixed teams, on testnet the same four MNs make up each team. Will try and put this to the test on functional tests.

monstrobishi:
@Bushstar sure you let me know when it's good

src/masternodes/anchors.cpp
Bushstar: Good spot. It was originally it0 in that loop, failed to update the arg!

src/spv/spv_wrapper.cpp
```diff
@@ -489,6 +500,17 @@ void CSpvWrapper::UpdateTx(uint256 const & hash, uint32_t blockHeight, uint32_t
     }
 }
 
+uint32_t CSpvWrapper::ReadTxTimestamp(uint256 const & hash)
+{
+    std::pair<char, uint256> const key{std::make_pair(DB_SPVTXS, hash)};
+    std::pair<TBytes, std::pair<uint32_t, uint32_t> > txrec;
```
ShengguangXiao, line 506: Can use `db_tx_rec` instead of `std::pair<TBytes, std::pair<uint32_t, uint32_t> >`

Bushstar: So we can...

Bushstar: https://github.com/DeFiCh/ain/commit/9bc92060574676aae85ddab7d0dbd845d66e9b39

ShengguangXiao:
@Bushstar I think the new API `getmasternodeblocks` may need do pagination, when I try to get blocks from a master node minted more than 80K blocks, it takes long time to response.

PR #221, opened 2021-02-10
Yield a slot to other staker threads

bvbfan:
It should give a time to other working thread.


bvbfan:
Strip out dead code.

ShengguangXiao:
The CI failed in `wallet_balance.py`, but the test can pass in local

bvbfan:
Yes, it's unrelated, it was fixed for couple of time but now start failing again :)

PR #234, opened 2021-02-19
 Refactor mn_rpc file and split into different categories based on RPC types

Mixa84:
Big mn_rpc.cpp has been split into several files for rpc categories respectively: rpc_masternodes.cpp, rpc_accounts.cpp, rpc_tokens.cpp, rpc_poolpair.cpp. All inlcudes, common functions, types and rpc command declarations have been moved to mn_rpc.h.
Setgov, getgov and isappliedcustomtx stayed in mn_rpc.cpp as they belong to blockchain category and didn't want to move them into separate file.


Mixa84:
Fixes cakedefi/defichain-private#442

src/masternodes/mn_rpc.h
```diff
@@ -0,0 +1,122 @@
...
+    // it means that we select first accounts with max sum of
+    // neccessary token amounts
+    SelectionPie,
+} AccountSelectionMode;
+
+// Special guarding object. Should be created before the first use of funding (straight or by GetAuthInputsSmart())
+struct LockedCoinsScopedGuard
+    {
+        CWallet * const pwallet;
+        std::set<COutPoint> lockedCoinsBackup;
```
bvbfan, line 39: Indentation.

src/masternodes/rpc_accounts.cpp
```diff
@@ -0,0 +1,1429 @@
...
+    return valueString + "@" + token->symbol + (token->IsDAT() ? "" : "#" + amount.nTokenId.ToString());
+}
+
+UniValue AmountsToJSON(TAmounts const & diffs) {
+    UniValue obj(UniValue::VARR);
+
+    for (auto const & diff : diffs) {
+        auto token = pcustomcsview->GetToken(diff.first);
+        auto const tokenIdStr = token->CreateSymbolKey(diff.first);
+        obj.push_back(ValueFromAmount(diff.second).getValStr() + "@" + tokenIdStr);
```
bvbfan, line 15: While we are here. That's same thing, 
```
for (auto const & diff : diffs) {
    obj.push_back(tokenAmountString({diff.first, diff.second}));
}
```
in `tokenAmountString`
```
auto const tokenIdStr = token->CreateSymbolKey(amount.nTokenId);
return ValueFromAmount(diff.nValue).getValStr() + "@" + tokenIdStr;
```

bvbfan (changes requested):
Maybe we can update `clang-format` and run it when do refactor?

src/masternodes/rpc_accounts.cpp
Mixa84: Do you mean to make it like this?

`std::string tokenAmountString(CTokenAmount const& amount) {
    auto const tokenIdStr = token->CreateSymbolKey(amount.nTokenId);
    return ValueFromAmount(diff.nValue).getValStr() + "@" + tokenIdStr;
}
UniValue AmountsToJSON(TAmounts const & diffs) {
    UniValue obj(UniValue::VARR);
    for (auto const & diff : diffs) {
        obj.push_back(tokenAmountString({diff.first, diff.second}));
    }
    return obj;
}`

bvbfan: Yep, you can verify they returns same string in rpc functional test (i mean not to write one, just to make sure they are identical, in existing one that uses these functions)

src/masternodes/rpc_accounts.cpp
```diff
@@ -0,0 +1,1429 @@
...
+                                   "Execution test failed: not enough balance on owner's account, call utxostoaccount to increase it.\n" +
+                                   res.msg);
+            }
+            throw JSONRPCError(RPC_INVALID_REQUEST, "Execution test failed:\n" + res.msg);
+        }
+    }
+
+    return signsend(rawTx, pwallet, optAuthTx)->GetHash().GetHex();
+
+}
```
ShengguangXiao, line 1429: Add new line at the end

src/masternodes/rpc_accounts.cpp
Mixa84: Not exactly the same, but as I see from tests it doesn't use tokenAmountString as it uses different output if indexed_amounts true.

Mixa84:
Lets agree on clang-format so I can we can use it when modifying something and adding new files. Then I'll run clang-format to align all files that I modified now.

src/masternodes/rpc_accounts.cpp
Mixa84: Done

src/masternodes/mn_rpc.h
```diff
@@ -0,0 +1,118 @@
...
+CTransactionRef send(CTransactionRef tx, CTransactionRef optAuthTx);
+CTransactionRef signsend(CMutableTransaction& mtx, CWallet* const pwallet, CTransactionRef optAuthTx /* = {}*/);
+CWallet* GetWallet(const JSONRPCRequest& request);
+std::vector<CTxIn> GetAuthInputsSmart(CWallet* const pwallet, int32_t txVersion, std::set<CScript>& auths, bool needFounderAuth, CTransactionRef& optAuthTx, UniValue const& explicitInputs);
+std::string ScriptToString(CScript const& script);
+CAccounts GetAllMineAccounts(CWallet* const pwallet);
+CAccounts SelectAccountsByTargetBalances(const CAccounts& accounts, const CBalances& targetBalances, AccountSelectionMode selectionMode);
+
+// masternode rpcs
+UniValue createmasternode(const JSONRPCRequest& request);
```
Bushstar, line 78: You do not need RPC calls in headers.

Bushstar: In each file with RPC calls add the calls to be something like the following. This is how Bitcoin separates their RPC calls and is consistent with the rest of the code. Take a look at RPC calls in rpc/mining.cpp, rpc/misc.cpp and rpc/net.cpp for example. Each defined the calls and register them in the same files. The new register call is then added to RegisterAllCoreRPCCommands in register.h.

```
static const CRPCCommand commands[] =
{ //  category      name                  actor (function)     params
  //  ----------------- ------------------------    -----------------------     ----------
    {"blockchain",  "setgov",                &setgov,                {"variables", "inputs"}},
    {"blockchain",  "getgov",                &getgov,                {"name"}},
    {"blockchain",  "isappliedcustomtx",     &isappliedcustomtx,     {"txid", "blockHeight"}},
};

void RegisterDefiBlockchainRPCCommands(CRPCTable& tableRPC) {
    for (unsigned int vcidx = 0; vcidx < ARRAYLEN(commands); vcidx++)
        tableRPC.appendCommand(commands[vcidx].name, &commands[vcidx]);
}
```

monstrobishi: Yes agree with this, please register in the same file

monstrobishi: To be more concise, in the same file they are defined (so multiple files)

ShengguangXiao (approved):

src/masternodes/mn_rpc.h
Mixa84: Done

monstrobishi (approved):

monstrobishi (approved):

Bushstar (approved):

bvbfan (approved):

PR #236, opened 2021-02-20
Reject protocol versions less than 70016

Bushstar:
Every hard fork we can raise the PROTOCOL_VERSION and after the hard fork raise the minimum accepted PROTOCOL_VERSION, this prevents old clients connecting to the network when they are no longer valid.


bvbfan:
Good catch! We can use `PROTOCOL_VERSION` and increase it on fork?

monstrobishi:
Any reason why we don't set get headers version here

Bushstar:
@bvbfan increase PROTOCOL_VERSION pre-fork and in the release after the fork height we increment min version.

@monstrobishi GETHEADERS is a historical protocol change in Bitcoin and is set to the minimum version there. For us the minimum protocol version is 70016. We could do things like:

```
static const int PROTOCOL_VERSION = 70017;
static const int DAKOTA_VERSION = 70017;
static const int MIN_PEER_PROTO_VERSION = 70016;
```

Then in the post fork release do the following.

```
static const int PROTOCOL_VERSION = 70017;
static const int DAKOTA_VERSION = 70017;
static const int MIN_PEER_PROTO_VERSION = DAKOTA_VERSION;
```

bvbfan:
> @bvbfan increase PROTOCOL_VERSION pre-fork and in the release after the fork height we increment min version.
> 
> ```
> static const int PROTOCOL_VERSION = 70017;
> static const int DAKOTA_VERSION = 70017;
> static const int MIN_PEER_PROTO_VERSION = 70016;
> ```

Yeah, i mean
```
static const int MIN_PEER_PROTO_VERSION = PROTOCOL_VERSION - 1;
```
or
```
static const int MIN_PEER_PROTO_VERSION = std::max(70016, PROTOCOL_VERSION - 1);
```

monstrobishi (approved):

PR #244, opened 2021-02-22
Migrate to new network message

bvbfan:
There is no need to hardfork message anymore, just use new one.


bvbfan:
It seems not work, i will investigate.

bvbfan:
The cannot be done in one step, due to `PushMessage` uses the old (Bitcoin) start message, so for now we can run some nodes with this patch in another commit we can remove https://github.com/DeFiCh/ain/pull/244/commits/41d467824d102ac02c45631c0890d8791504c87a#diff-00021eed586a482abdb09d6cdada1d90115abe988a91421851960e26658bed02R578

monstrobishi (approved):

ShengguangXiao (approved):

PR #245, opened 2021-02-22
License file naming update

uzyn:
Renaming `COPYING` to `LICENSE`


LICENSE
```diff
@@ -2,7 +2,7 @@ The MIT License (MIT)
 
 Copyright (c) 2009-2019 The Bitcoin Core developers
 Copyright (c) 2009-2019 Bitcoin Developers
-Copyright (c) 2020 DeFi Blockchain Developers
+Copyright (c) DeFi Blockchain Developers
```
fuxingloh, line 5: `Copyright (c) DeFiChain Developers`?

uzyn: Officially we are DeFi Blockchain, at least that's where we started at, but over time DeFiChain has taken over for general use.

Don't mind changing it, but it involves a more thorough changing. The key thing for this PR is for us to rename `COPYING` to `LICENSE` so it implies coverage on usage license rather than just code copying license.

PR #254, opened 2021-02-25
Wipe confs/revote after new team. New getactivemasternodes call.

Bushstar:
Anchor confirms should be wiped on team change and then a revote is taken. New call added to view number of active masternodes in the last X number of blocks, defaults to 20,160 blocks, the number of blocks the anchor teams use to generate anchor teams. Reject conf from old team members rather than rely on GetQuorumFor to filter them out.

### DIFF
--- a/src/masternodes/anchors.cpp
+++ b/src/masternodes/anchors.cpp
@@ -989,6 +989,18 @@ bool CAnchorAwaitingConfirms::Validate(CAnchorConfirmMessage const &confirmMessa
         return false;
     }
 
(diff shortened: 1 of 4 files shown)


ShengguangXiao:
active master node means the nodes minted valid blocks? I tested, it works good.

ShengguangXiao (approved):

Bushstar:
@ShengguangXiao that's correct. I renamed it to getactivemasternodecount to be more in line with other calls that return a count of something.

monstrobishi (approved):

PR #255, opened 2021-02-26
Flush when it's needed in reindex chainstate

bvbfan:
/kind fix

In `-reindex-chainstate` flush mode periodic isn't called

Fixes https://github.com/cakedefi/defichain-private/issues/423

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2496,7 +2496,7 @@ bool CChainState::FlushStateToDisk(
             nLastWrite = nNow;
         }
         static const size_t memoryCacheSizeMax = gArgs.GetBoolArg("-acindex", DEFAULT_ACINDEX) ? (nDefaultDbCache << 16) : (nDefaultDbCache << 10);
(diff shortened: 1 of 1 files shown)


src/validation.cpp
```diff
@@ -2496,7 +2496,7 @@ bool CChainState::FlushStateToDisk(
             nLastWrite = nNow;
         }
         static const size_t memoryCacheSizeMax = gArgs.GetBoolArg("-acindex", DEFAULT_ACINDEX) ? (nDefaultDbCache << 16) : (nDefaultDbCache << 10);
-        bool fMemoryCacheLarge = fDoFullFlush || (mode == FlushStateMode::PERIODIC && pcustomcsview->SizeEstimate() > memoryCacheSizeMax);
+        bool fMemoryCacheLarge = fDoFullFlush || (mode == FlushStateMode::IF_NEEDED && pcustomcsview->SizeEstimate() > memoryCacheSizeMax);
```
ShengguangXiao, line 2499: Use const for fMemoryCacheLarge.

ShengguangXiao (approved):

uzyn:
Tested on a node, confirmed that this PR addresses the issue. Thanks @bvbfan

PR #261, opened 2021-03-04
Invalid tx should not be included in block

bvbfan:
/kind fix

<s>It has 2 issues about me:</s>

1. Failed tx never be part of a block, i try to ensure it
<s>2. We use a view to ensure txs validation in a block not in isolation, but we fail to discard invalid view.</s>

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -588,8 +588,8 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
         if (!customTxPassed) {
             if (fUsingModified) {
                 mapModifiedTx.get<ancestor_score>().erase(modit);
-                failedTx.insert(iter);
             }
+            failedTx.insert(iter);
             continue;
         }
 


Bushstar:
Note that it is native Bitcoin behaviour to have `failedTx.insert(iter)` in a `fUsingModified` conditional. 

https://github.com/bitcoin/bitcoin/blob/0.19/src/miner.cpp#L373

src/txmempool.cpp
```diff
@@ -594,10 +594,13 @@ void CTxMemPool::removeForBlock(const std::vector<CTransactionRef>& vtx, unsigne
...
                 continue;
             }
-            auto res = ApplyCustomTx(viewDuplicate, mempoolDuplicate, tx, Params().GetConsensus(), nBlockHeight, 0, false);
+            CCustomCSView workView(viewDuplicate); // we don't need invalid tx to change the view
+            auto res = ApplyCustomTx(workView, mempoolDuplicate, tx, Params().GetConsensus(), nBlockHeight, 0, false);
             if (!res.ok && (res.code & CustomTxErrCodes::Fatal)) {
                 LogPrintf("%s: Remove conflicting custom TX: %s\n", __func__, tx.GetHash().GetHex());
                 staged.insert(mapTx.project<0>(it));
+            } else {
+                workView.Flush(); // apply changes when they are checked
```
ShengguangXiao, line 603: Inside `ApplyCustomTx` if it is success, it already flushed, why here need to flush again?

bvbfan: Yeah, i forgot that `ApplyCustomTx` do it for us, when tx is invalid it does not flush.

Bushstar:
The last entry is bespoke to us and 

```
        // Failed, let's move on!
        if (!customTxPassed) {
            if (fUsingModified) {
                mapModifiedTx.get<ancestor_score>().erase(modit);
            }
            failedTx.insert(iter);
            continue;
        }
```

ShengguangXiao (approved):

PR #267, opened 2021-03-04
Prefer entry time over hash in tx ordering

bvbfan:
/kind fix

We must have right order when applying tx in account system or tx may be invalid.

### DIFF
--- a/src/miner.h
+++ b/src/miner.h
@@ -89,7 +89,7 @@ struct CompareTxIterByAncestorCount {
     {
         if (a->GetCountWithAncestors() != b->GetCountWithAncestors())
             return a->GetCountWithAncestors() < b->GetCountWithAncestors();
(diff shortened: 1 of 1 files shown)


src/miner.h
```diff
@@ -89,7 +89,7 @@ struct CompareTxIterByAncestorCount {
     {
         if (a->GetCountWithAncestors() != b->GetCountWithAncestors())
             return a->GetCountWithAncestors() < b->GetCountWithAncestors();
-        return CTxMemPool::CompareIteratorByHash()(a, b);
+        return CompareTxMemPoolEntryByEntryTime()(*a, *b);
```
ShengguangXiao, line 92: Is the transaction entry time is the same for different nodes? Does it cause different nodes have different order problem?

monstrobishi: It's only in miner though so doesn't make a difference

monstrobishi (approved):

PR #275, opened 2021-03-08
Check pending anchors in same ordrer as top anchor selection

Bushstar:
When checking pending anchors make sure we check them in the same order that top anchors are chosen. Otherwise we can end up with a different anchor chain than expected. After restarting the client the anchor chain is re-evaluated and correctly set.

Also reject anchors that do not meet time requirement rather than leave them in pending.

### DIFF
--- a/src/masternodes/anchors.cpp
+++ b/src/masternodes/anchors.cpp
@@ -526,9 +526,26 @@ void CAnchorIndex::CheckPendingAnchors()
 {
     AssertLockHeld(cs_main);
 
-    std::set<uint256> deletePending;
+    // Get pending anchors and organise by Bitcoin height, anchor height or TX hash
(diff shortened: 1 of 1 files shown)


monstrobishi (approved):

src/masternodes/anchors.cpp
```diff
@@ -526,9 +526,26 @@ void CAnchorIndex::CheckPendingAnchors()
...
+            if (a.anchor.height == b.anchor.height) {
+                return a.txHash < b.txHash;
+            }
+
+            // Higher DeFi height wins
+            return a.anchor.height > b.anchor.height;
+        }
+
+        return a.btcHeight < b.btcHeight;
+    };
```
ShengguangXiao, line 541: Can we some unittest for this function?

Bushstar: @ShengguangXiao We totally can. Will add one tomorrow.

PR #282, opened 2021-03-11
Bitcoin support via SPV

Bushstar:
Add the ability to send and receive Bitcoins directly in the DeFi wallet via SPV. This PR adds the following commands.

spv_dumpprivkey ADDRESS
Return the private key that relates to the address provided.

spv_getbalance
Displays the Bitcoin balance.

spv_getnewaddress
Returns a new Bitcoin address generated from a pubkey held in the DeFiChain wallet.

spv_getrawtransaction TXID
Get raw transaction that relates to the provided TXID if found.

spv_listtransactions
Returns an array of all user related TXIDs in the wallet.

spv_sendtoaddress ADDRESS AMOUNT
Send Bitcoin.

spv_fundaddress (For function test framework)
Hidden command used in testing to fund the Bitcoin wallet via a dummy transaction.


src/spv/spv_rpc.cpp
```diff
@@ -866,6 +848,213 @@ UniValue spv_setlastheight(const JSONRPCRequest& request)
...
+static UniValue spv_getnewaddress(const JSONRPCRequest& request)
+{
+    CWallet* const pwallet = GetWallet(request);
+
+    RPCHelpMan{"spv_getnewaddress",
+        "\nCreates and adds a Bitcoin address to the SPV wallet\n",
+        {
+        },
+        RPCResult{
+            "\"none\"                  Returns nothing\n"
```
ShengguangXiao, line 888: Here should returns the new address right?

src/spv/bitcoin/BRWallet.h
```diff
@@ -56,11 +58,19 @@ inline static int BRUTXOEq(const void *utxo, const void *otherUtxo)
...
+// Convert SPV UInt256 to Bitcoin uint256
+uint256 to_uint256(UInt256 const & i);
+
 typedef struct BRWalletStruct BRWallet;
 
 // allocates and populates a BRWallet struct that must be freed by calling BRWalletFree()
 // forkId is 0 for bitcoin, 0x40 for b-cash
-BRWallet *BRWalletNew(BRTransaction *transactions[], size_t txCount, BRMasterPubKey mpk, int forkId);
+BRWallet *BRWalletNew(BRTransaction *transactions[], size_t txCount, BRMasterPubKey mpk, int forkId,
+                      std::set<UInt160, decltype(&UInt160Compare)>&& userAddresses);
```
ShengguangXiao, line 73: Can define a type for `std::set<UInt160, decltype(&UInt160Compare)>`? I see it used in quite a lot places.

Bushstar: It's a Lambda comparator.

src/spv/spv_rpc.cpp
Bushstar: I think so :)

src/spv/bitcoin/BRWallet.h
Bushstar: I'll update with a struct.

Bushstar: @ShengguangXiao https://github.com/DeFiCh/ain/pull/282/commits/eaf75eff44be33b9139bf99a2a2f25dfc6108746

ShengguangXiao: Actually what i mean is make std::set<UInt160, UInt160Compare> as SetOfAddress like below, but i think now is quite simple already. Thanks.
```
using SetOfAddress = std::set<UInt160, UInt160Compare>
```

Bushstar: Oh right, I thought you did not like the somewhat cumbersome decltype(&UInt160Compare) as the comparator, I guess a using would have helped there. I often personally prefer to see what it is and not a using declaration in its place, with the decltype included a using declaration would have been better but I think it is okay now.

ShengguangXiao:
@Bushstar I tested `spv_sendtoaddress` on testnet, if it still in syncing status, then this RPC call will hangup and timeout. After fully synced, it works ok.

Bushstar:
@ShengguangXiao good spot. Will look into and remedy this on Monday.

Bushstar:
@ShengguangXiao added a commit. Sending when SPV is not connected or less than the last days worth of blocks from the tip is not allowed.

ShengguangXiao (approved):

ShengguangXiao:
> @ShengguangXiao added a commit. Sending when SPV is not connected or less than the last days worth of blocks from the tip is not allowed.

Now it works good.

monstrobishi (approved):

monstrobishi:
@Bushstar how is this handling the change address?

PR #284, opened 2021-03-15
Scale difficulty target and limit future block time

bvbfan:
/kind fix

Fixes https://github.com/cakedefi/defichain-private/issues/544

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -128,7 +128,7 @@ public:
         consensus.pos.nTargetTimespan = 5 * 60; // 5 min == 10 blocks
         consensus.pos.nTargetSpacing = 30; // seconds
         consensus.pos.nStakeMinAge = 0;
-        consensus.pos.nStakeMaxAge = 7 * 24 * 60 * 60; // One week
+        consensus.pos.nStakeMaxAge = 14 * 24 * 60 * 60; // Two weeks
         consensus.pos.fAllowMinDifficultyBlocks = false; // only for regtest
         consensus.pos.fNoRetargeting = false; // only for regtest
 
@@ -311,7 +311,7 @@ public:
         consensus.pos.nTargetTimespan = 5 * 60; // 5 min == 10 blocks
         consensus.pos.nTargetSpacing = 30;
         consensus.pos.nStakeMinAge = 0;
-        consensus.pos.nStakeMaxAge = 7 * 24 * 60 * 60; // One week
+        consensus.pos.nStakeMaxAge = 14 * 24 * 60 * 60; // Two weeks
         consensus.pos.fAllowMinDifficultyBlocks = false;
         consensus.pos.fNoRetargeting = false; // only for regtest
 
@@ -458,7 +458,7 @@ public:
         consensus.pos.nTargetTimespan = 5 * 60; // 5 min == 10 blocks
         consensus.pos.nTargetSpacing = 30;
         consensus.pos.nStakeMinAge = 0;
-        consensus.pos.nStakeMaxAge = 7 * 24 * 60 * 60; // One week
+        consensus.pos.nStakeMaxAge = 14 * 24 * 60 * 60; // Two weeks
         consensus.pos.fAllowMinDifficultyBlocks = false;
         consensus.pos.fNoRetargeting = false; // only for regtest
 
@@ -599,7 +599,7 @@ public:
         consensus.pos.nTargetTimespan = 14 * 24 * 60 * 60; // two weeks
         consensus.pos.nTargetSpacing = 10 * 60; // 10 minutes
         consensus.pos.nStakeMinAge = 0;
-        consensus.pos.nStakeMaxAge = 7 * 24 * 60 * 60; // One week
+        consensus.pos.nStakeMaxAge = 14 * 24 * 60 * 60; // Two weeks
         consensus.pos.fAllowMinDifficultyBlocks = true; // only for regtest
         consensus.pos.fNoRetargeting = true; // only for regtest
 
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -120,7 +120,7 @@ public:
         consensus.BayfrontGardensHeight = 488300;
         consensus.ClarkeQuayHeight = 595738;
         consensus.DakotaHeight = 678000; // 1st March 2021
-        consensus.DakotaCrescentHeight = std::numeric_limits<int>::max();
+        consensus.DakotaCrescentHeight = 733000; // 25th March 2021
 
         consensus.pos.diffLimit = uint256S("00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
 //        consensus.pos.nTargetTimespan = 14 * 24 * 60 * 60; // two weeks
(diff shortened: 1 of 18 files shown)


Bushstar:
@bvbfan 5 seconds is crazy tight, we do use decentralised time, we will see a lot of rejected blocks I think. On Feathercoin we used 15 minutes and I know another coin that uses 5 minutes. Also I think we should stick with the Bitcoin design and use nAdjustedTime.

https://github.com/FeatherCoin/Feathercoin/commit/5f208deac78cb8ad59baa52df59ac6bd73ad0473

ShengguangXiao:
> @bvbfan 5 seconds is crazy tight, we do use decentralised time, we will see a lot of rejected blocks I think. On Feathercoin we used 15 minutes and I know another coin that uses 5 minutes. Also I think we should stick with the Bitcoin design and use nAdjustedTime.
> 
> [FeatherCoin/Feathercoin@5f208de](https://github.com/FeatherCoin/Feathercoin/commit/5f208deac78cb8ad59baa52df59ac6bd73ad0473)

Yes, in the discuss we agreed to use 5 minutes

src/chainparams.cpp
```diff
@@ -446,10 +452,13 @@ class CDevNetParams : public CChainParams {
         consensus.BayfrontGardensHeight = 300;
         consensus.ClarkeQuayHeight = 500;
         consensus.DakotaHeight = std::numeric_limits<int>::max();
+        consensus.DakotaCrescentHeight = std::numeric_limits<int>::max();
 
         consensus.pos.diffLimit = uint256S("00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.pos.nTargetTimespan = 5 * 60; // 5 min == 10 blocks
         consensus.pos.nTargetSpacing = 30;
+        consensus.pos.nStakeMinAge = 24 * 60 * 60; // One day
```
uzyn, line 460:
```suggestion
        consensus.pos.nStakeMinAge = 0; // You are eligible to mine immediately
```

src/pos_kernel.cpp
```diff
@@ -14,15 +17,69 @@ namespace pos {
...
+                }
+                else
+                {
+                    // Did not find staker block within max age so set to max age.
+                    nTimeTx = params.pos.nStakeMaxAge;
+                }
+            }
+
+            // Calculate coinDayWeight, at min this is 1 with no impact on difficulty.
+            arith_uint256 coinDayWeight = nTimeTx / (24 * 60 * 60);
```
uzyn, line 75: Suggesting to lower the period from a day to 6 hours. 

```suggestion
            arith_uint256 period = 6 * 60 * 60; // 6 hours
            arith_uint256 coinWeight = (nTimeTx + period) / (period);
```

uzyn (changes requested):

src/validation.cpp
```diff
@@ -3722,8 +3722,13 @@ static bool ContextualCheckBlockHeader(const CBlockHeader& block, CValidationSta
         return state.Invalid(ValidationInvalidReason::BLOCK_INVALID_HEADER, false, REJECT_INVALID, "time-too-old", "block's timestamp is too early");
 
     // Check timestamp
-    if (block.GetBlockTime() > nAdjustedTime + MAX_FUTURE_BLOCK_TIME)
-        return state.Invalid(ValidationInvalidReason::BLOCK_TIME_FUTURE, false, REJECT_INVALID, "time-too-new", "block timestamp too far in the future");
+    if (block.height < consensusParams.DakotaCrescentHeight) {
+        if (block.GetBlockTime() > nAdjustedTime + MAX_FUTURE_BLOCK_TIME)
+            return state.Invalid(ValidationInvalidReason::BLOCK_TIME_FUTURE, false, REJECT_INVALID, "time-too-new", "block timestamp too far in the future");
+    } else {
```
uzyn, line 3728: We don't need `else` here.

uzyn (commented):
Thanks for the fix @Bushstar

src/pos_kernel.cpp
```diff
@@ -14,15 +17,70 @@ namespace pos {
...
+                    {
+                        auto id = pcustomcsview->GetMasternodeIdByOperator(staker);
+                        if (id && id == masternodeID)
+                        {
+                            // Found the last staked block
+                            stakerBlock = tip;
+                            break;
+                        }
+                    }
+                }
```
ShengguangXiao, line 60: How is the performance of this section of code? If consider the lookup for nStakeMaxAge which is 2 weeks, then roughly need to search for 32,000 blocks and each time need to call `ExtractMinterKey` and `GetMasternodeIdByOperator`, it may take some time. Maybe can consider save the staker information in cache, using lookup will be much faster.

Bushstar: The performance is very bad, something like 2.7 seconds. Working on a DB look up instead.

Bushstar: Also will save staker block in an in memory cache after loading from DB to save repeat lookups.

muirglacier:
Hi U-Zyn! Dan here.

I am just curious about one thing. The coin age thing is implemented in a way, that nodes that never have found a block have a coin weight of one month right?

Does that mean that every time I find a block its smarter for me to resign a masternode and set it up from scratch, so it has a fresh id without any staked blocks in its history?

src/validation.cpp
```diff
@@ -1723,7 +1723,32 @@ DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockI
...
+            {
+                auto id = pcustomcsview->GetMasternodeIdByOperator(staker);
+                if (id && id == nodeId)
+                {
+                    // Found the last staked block
+                    blockTime = tip->GetBlockTime();
+                    break;
+                }
+            }
+        }
```
ShengguangXiao, line 1749: Why here don't use `GetMasternodeLastBlockTime` ?

Bushstar: GetMasternodeLastBlockTime only stores time against a masternode ID so gets overwritten, the entry in GetMasternodeLastBlockTime is the one we are overwriting here.

ShengguangXiao (approved):

ShengguangXiao:
Looks good to me now, only need to set the `DakotaCrescentHeight`.

src/validation.cpp
bvbfan: I'm still concern about, disconnect block time will increase.

ShengguangXiao: Yes, maybe can consider cache the results.

Bushstar: @bvbfan your suggestion was to use <masternode, height>, won't we then be iterating through all records each time to find the latest height on sync for each block, rather than just get the latest time stored against the masternode?

Or do we store latest time against masternode as is with the addition of another <masternode, height> record to iterate over when running DisconnectBlock?

We may still end up iterating over a lot of records to find the latest height, max time for the loop above is 1 second if it has to go all the way back through 7 days of blocks on my computer.

bvbfan: @Bushstar yep, that's right. We can get masternode id then you can search by partial key `{masternodeid, 0}` you can find last one.

src/pos_kernel.cpp
Bushstar: We now store the masternode and time for every single block so the above code is now obsolete.

src/validation.cpp
Bushstar: We now store the masternode and time for every single block so the above code is now obsolete.

PR #294, opened 2021-03-25
Limit db writes in normal node operation

bvbfan:
/kind fix

It increase a bit memory usage to limit db writes frequency.

Fixes https://github.com/cakedefi/defichain-private/issues/487


ShengguangXiao (approved):

src/init.cpp
```diff
@@ -1476,6 +1476,7 @@ bool AppInitMain(InitInterfaces& interfaces)
     int64_t nTotalCache = (gArgs.GetArg("-dbcache", nDefaultDbCache) << 20);
     nTotalCache = std::max(nTotalCache, nMinDbCache << 20); // total cache cannot be less than nMinDbCache
     nTotalCache = std::min(nTotalCache, nMaxDbCache << 20); // total cache cannot be greater than nMaxDbcache
+    auto nCustomCacheSize = nTotalCache; // used for criminals and customs
```
ShengguangXiao, line 1479: Can log this value into logfile so can track the value?

monstrobishi (approved):

PR #296, opened 2021-03-26
Show block reward in listaccounthistory

bvbfan:
/kind feature

It shows minter block reward


monstrobishi (approved):

ShengguangXiao:
This PR works good. But i suggest to change the output "txn" to "txcount" to align with rpc `getwalletinfo` and `getchaintxstats`.

ShengguangXiao (approved):

monstrobishi:
@ShengguangXiao This RPC call is already in use so I suggest we don't change that part

monstrobishi (approved):

PR #297, opened 2021-03-30
Feature/getmininginfo

surangap:

/kind feature

Fixes cakedefi/defichain-private#522

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -270,27 +270,28 @@ static UniValue getmintinginfo(const JSONRPCRequest& request)
 //Returns the mining information of all local masternodes
 static UniValue getmininginfo(const JSONRPCRequest& request)
 {
-            RPCHelpMan{"getmininginfo",
-                "\nReturns a json object containing mining-related information for all local masternodes",
-                {},
-                RPCResult{
-                    "{\n"
-                    "  \"blocks\": nnn,             (numeric) The current block\n"
-                    "  \"currentblockweight\": nnn, (numeric, optional) The block weight of the last assembled block (only present if a block was ever assembled)\n"
-                    "  \"currentblocktx\": nnn,     (numeric, optional) The number of block transactions of the last assembled block (only present if a block was ever assembled)\n"
-                    "  \"generate\": true|false     (boolean) If the generation is on or off (see getgenerate or setgenerate calls)\n"
-                    "  \"difficulty\": xxx.xxxxx    (numeric) The current difficulty\n"
-                    "  \"networkhashps\": nnn,      (numeric) The network hashes per second\n"
-                    "  \"pooledtx\": n              (numeric) The size of the mempool\n"
-                    "  \"chain\": \"xxxx\",           (string) current network name as defined in BIP70 (main, test, regtest)\n"
-                    "  \"warnings\": \"...\"          (string) any network and blockchain warnings\n"
-                    "}\n"
-                },
-                RPCExamples{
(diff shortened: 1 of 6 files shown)


src/rpc/mining.cpp
```diff
@@ -320,9 +320,13 @@ static UniValue getmininginfo(const JSONRPCRequest& request)
         subObj.pushKV("generate", node.IsActive() && gArgs.GetBoolArg("-gen", DEFAULT_GENERATE));
         subObj.pushKV("mintedblocks", (uint64_t)node.mintedBlocks);
 
-        auto lastBlockCreationAttemptTs = node.lastBlockCreationAttemptTs;
-        subObj.pushKV("lastblockcreationattempt", FormatISO8601DateTime(node.lastBlockCreationAttemptTs));
-
+        //get the last block creation attempt by the master node
+        {
```
bvbfan, line 324: We can add check here, if node does not generate we does not need to `mapMNLastBlockCreationAttemptTs` since it does not work. Can you and test for that?

src/rpc/mining.cpp
```diff
@@ -267,6 +267,73 @@ static UniValue getmintinginfo(const JSONRPCRequest& request)
...
 
+// Returns the mining information of all local masternodes
+static UniValue getmininginfo(const JSONRPCRequest& request)
+{
+    RPCHelpMan{"getmininginfo",
+        "\nReturns a json object containing mining-related information for all local masternodes",
+        {},
+        RPCResult{
+            "{\n"
+            "  \"blocks\": nnn,             (numeric) The current block\n"
```
ShengguangXiao, line 278: If it is a single block number, should be `block`

src/rpc/mining.cpp
```diff
@@ -267,6 +267,73 @@ static UniValue getmintinginfo(const JSONRPCRequest& request)
...
+            "  \"currentblockweight\": nnn, (numeric, optional) The block weight of the last assembled block (only present if a block was ever assembled)\n"
+            "  \"currentblocktx\": nnn,     (numeric, optional) The number of block transactions of the last assembled block (only present if a block was ever assembled)\n"
```
ShengguangXiao, line 280: transaction count should use `currentblocktxcount`

src/rpc/mining.cpp
surangap: Code updated. thanks.

ShengguangXiao (approved):

src/rpc/mining.cpp
```diff
@@ -267,6 +268,76 @@ static UniValue getmintinginfo(const JSONRPCRequest& request)
...
+    UniValue mnArr(UniValue::UniValue::VARR); // masternodes array
+    for (const auto& mnId : mnIds) {
+        UniValue subObj(UniValue::VOBJ);
+
+        subObj.pushKV("masternodeid", mnId.second.GetHex());
+        CMasternode const & node = *pcustomcsview->GetMasternode(mnId.second);
+        auto state = node.GetState();
+        subObj.pushKV("masternodeoperator", node.operatorAuthAddress.GetHex());// NOTE(sp) : Should this also be encoded? not the HEX
+        subObj.pushKV("masternodestate", CMasternode::GetHumanReadableState(state));
+        auto generate = node.IsActive() && gArgs.GetBoolArg("-gen", DEFAULT_GENERATE);
```
ShengguangXiao, line 321: Can move `gArgs.GetBoolArg("-gen", DEFAULT_GENERATE);` move outside the loop and assign to a variable.

PR #279, opened 2021-03-10
Lazy rewards redistribution

bvbfan:
/kind feature

I plan to run it in real sync, i have another idea using undos, probably will give a try in another repo. For now we have pool commissions updates on every block, we can fork out by next hard fork. Account rewards are updated before every addition/subtract, optionally we can do it before subtraction only. 

Fixes https://github.com/cakedefi/defichain-private/issues/347

PR #305, opened 2021-04-06
Feature/decodecustomtx

surangap:

/kind feature

Fixes cakedefi/defichain-private#297


ShengguangXiao:
`iswitness` need to be defined in `vRPCConvertParams` in `client.cpp`

src/masternodes/rpc_tokens.cpp
```diff
@@ -825,6 +974,7 @@ static const CRPCCommand commands[] =
     {"tokens",      "gettoken",              &gettoken,              {"key" }},
     {"tokens",      "getcustomtx",           &getcustomtx,           {"txid", "blockhash"}},
     {"tokens",      "minttokens",            &minttokens,            {"amounts", "inputs"}},
+    {"tokens",      "decodecustomtx",        &decodecustomtx,        {"hexstring","iswitness"}},
```
ShengguangXiao, line 977: Add blank between strings.

src/test/rpc_tests.cpp
```diff
@@ -79,6 +79,17 @@ BOOST_AUTO_TEST_CASE(rpc_rawparams)
...
+    BOOST_CHECK_NO_THROW(r = CallRPC(std::string("decodecustomtx ")+txHexString));
+    //we can only check for type here. Because even if the relevant tx is still not in the system, the type can be returned
+    //with an error.
+    BOOST_CHECK_EQUAL(find_value(r.get_obj(), "type").get_str(), "PoolSwap");
```
ShengguangXiao, line 90: Need to check other results here too, like block height and block hash.

surangap: I think can not check for the other results always. the response depends on the block data on the running machine.
therefore can only add consistent fields.
I will try to add another functional test that covers more cases.

monstrobishi (changes requested):
This is good and the code is clean but I would like to request a functionality change. We should be able to still return the metadata if the tx is not found on chain. Will leave a comment on the appropriate line.

src/masternodes/rpc_tokens.cpp
```diff
@@ -815,6 +815,155 @@ UniValue minttokens(const JSONRPCRequest& request) {
...
+        if ((tx->IsCoinBase() && nHeight > 0)) {
+            return "Coinbase transaction. Not a custom transaction.";
+        }
+
+        if (!GetCustomTXInfo(nHeight, tx, guess, res, txResults)) {
+            return "Not a custom transaction";
+        }
+
+    } else {
+        // Should not get here
```
monstrobishi, line 928: I think here we don't have to error out, we can still return the metadata, see GuessCustomTxType, we don't need any on-chain info to get info such as the parameters etc, only for the result

src/masternodes/rpc_tokens.cpp
```diff
@@ -815,6 +815,155 @@ UniValue minttokens(const JSONRPCRequest& request) {
...
+    bool actualHeight{false};
+    CustomTxType guess;
+    UniValue txResults(UniValue::VOBJ);
+    Res res{};
+    uint256 hashBlock;
+    CBlockIndex* blockindex{nullptr};
+    CTransactionRef tx = MakeTransactionRef(std::move(mtx));
+    std::string warnings;
+
+    if (tx)
```
monstrobishi, line 871: Actually also considering suggesting removing this section, and making this function only return the parameters of the tx, not the result (As if we have tx hash we can just call getcustomtx for that anyway).

monstrobishi:
Just leaving a note here, let's revisit this after this PR is merged in:

https://github.com/DeFiCh/ain/pull/279/commits/c7b5d65095486301c82e2f144218f1119c05f77f

src/validation.cpp
```diff
@@ -1005,10 +1005,50 @@ bool GetTransaction(const uint256& hash, CTransactionRef& txOut, const Consensus
     return false;
 }
 
+/**
+ * If the transaction hash is found inside a block, the block hash is placed in hashBlock.
+ * If blockIndex is provided, the transaction is fetched from the corresponding block and its hash is placed in hashBlock.
+ */
+bool GetBlockHash(const uint256& hash, const Consensus::Params& consensusParams, uint256& hashBlock, const CBlockIndex* const blockIndex)
```
ShengguangXiao, line 1012: I feel like this function name should change to something like `GetBlockHashFromTxHash` to distinguish from the existing `GetBlockHash` class member function. And the first input parameter name can change to `hashTx` to easy understand.

ningfeiguang:
Tested on testnet, the functionality works good.

ShengguangXiao:
Tested on testnet, the functionality works good.

surangap:
Update after https://github.com/DeFiCh/ain/pull/279 is merged

PR #307, opened 2021-04-07
Updated the getmintinginfo and getmininginfo to fix referencing on rval.

surangap:

/kind fix

Fixes #


bvbfan (approved):

bvbfan:
One more place
https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L2025

surangap:
> One more place
> https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L2025

How can I miss that? I did a project-wide search. thanks!

surangap:
> > One more place
> > https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L2025
> 
> How can I miss that? I did a project-wide search. thanks!

oh yeah, the text is different.

src/validation.cpp
```diff
@@ -2022,18 +2022,24 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
         auto nodeId = mnview.GetMasternodeIdByOperator(minterKey);
         assert(nodeId);
-        auto const & node = *mnview.GetMasternode(*nodeId);
-        if (node.mintedBlocks + 1 != block.mintedBlocks)
+        auto nodePtr = mnview.GetMasternode(*nodeId);
+        if (!nodePtr) {
+            // NOTE(sp): Add branch prediction optimization here?
+            return state.Invalid(ValidationInvalidReason::CONSENSUS, error("ConnectBlock(): Failed to find masternode %s in the database",
+                                                                           nodeId->ToString()), REJECT_INVALID, "bad-masternode-id");
+        }
```
bvbfan, line 2030: Just assert, if data is missing that means storage corruption.

surangap: `\#if defined(NDEBUG)
\# error "Defi cannot be compiled without assertions."
\#endif`

Usually, release builds switch off assertions. but here we can use it. will update.

ShengguangXiao (approved):

ShengguangXiao:
@surangap @bvbfan After some study, I found that this PR maybe isn’t necessary because the temporary object returned by the function is bound to a const lvalue reference. See first exception to temporary object lifetime rules noted [here](https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime).
In fact, since C++11 it would even be fine if the temporary object was bound to an rvalue reference. See example code [here](https://en.cppreference.com/w/cpp/language/reference#Rvalue_references).

And i also writed some code to verify, it works good.
```
static std::optional<int> getValue() {
    std::optional<int> value{1};
    return value;
}


void TestGetOptionalValue() {
    const auto& value = *getValue();
    std::cout << "Get value result " << value << std::endl;
}
```

bvbfan:
@ShengguangXiao no it's a bug, never test with type which size is <= sizeof(void*) so replace int with bigger object and it will not work (e.g. string). You can extend optional object by taking const ref to it but you cannot underlying object.
https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary
```
a temporary bound to a return value of a function in a return statement is not extended: it is destroyed immediately at the end of the return expression. Such return statement always returns a dangling reference.
```

ShengguangXiao:
> optional object by taking const ref to it but you cannot underlying object.
> https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary

Yes, it is bug, the problem is we should not directly use * to get the result of optional value, if assign to an const reference, then get the value is good, see the below example.

```
struct TestStruct{
    int a;
    float b;
    std::string c;
};

static std::optional<TestStruct> getValue() {
    std::optional<TestStruct> value{
        TestStruct{
        1, 1.0, "abc"}
    };
    return value;
}


void TestGetOptionalValue() {
    const auto& valueOption = getValue();       // Good, assign to an const reference
    auto& valueResult = *valueOption;           // Good, get value from the valid reference

    const auto& value = *getValue();            // Not good, pointer to a dangling reference
    std::cout << "Get value result " << value.a << value.c << std::endl;
}
```

surangap:
> > optional object by taking const ref to it but you cannot underlying object.
> > https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary
> 
> Yes, it is bug, the problem is we should not directly use * to get the result of optional value, if assign to an const reference, then get the value is good, see the below example.
> 
> ```
> struct TestStruct{
>     int a;
>     float b;
>     std::string c;
> };
> 
> static std::optional<TestStruct> getValue() {
>     std::optional<TestStruct> value{
>         TestStruct{
>         1, 1.0, "abc"}
>     };
>     return value;
> }
> 
> 
> void TestGetOptionalValue() {
>     const auto& valueOption = getValue();       // Good, assign to an const reference
>     auto& valueResult = *valueOption;           // Good, get value from the valid reference
> 
>     const auto& value = *getValue();            // Not good, pointer to a dangling reference
>     std::cout << "Get value result " << value.a << value.c << std::endl;
> }
> ```

Could change to avoid extra optional copy. But we have several places in the codebase to change. Probably should do it under a separate ticket. But for now, the bug is fixed.

PR #310, opened 2021-04-08
SPV: Skip pending anchor until block confirmed

Bushstar:
Timestamp can be set to system time and not blocktime before the transaction is actually confirmed. We should consider an anchor TX with a blockheight of max int32 still pending even if a timestamp is set.

Correction also made in this PR to the newly created ReadTxBlockHeight by returning the correct unconfirmed block value of max int32.

### DIFF
--- a/src/masternodes/anchors.cpp
+++ b/src/masternodes/anchors.cpp
@@ -547,15 +547,16 @@ void CAnchorIndex::CheckPendingAnchors()
         }
 
         uint32_t timestamp = spv::pspv->ReadTxTimestamp(rec.txHash);
+        auto blockHeight = spv::pspv->ReadTxBlockHeight(rec.txHash);
 
-        // Do not delete, TX time still pending.
-        if (timestamp == 0 || timestamp == std::numeric_limits<int32_t>::max()) {
+        // Do not delete, TX time still pending. If block height is set to max we cannot trust the timestamp.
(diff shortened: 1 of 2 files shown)


monstrobishi (approved):

src/masternodes/anchors.cpp
```diff
@@ -547,15 +547,16 @@ void CAnchorIndex::CheckPendingAnchors()
         }
 
         uint32_t timestamp = spv::pspv->ReadTxTimestamp(rec.txHash);
+        auto blockHeight = spv::pspv->ReadTxBlockHeight(rec.txHash);
```
ShengguangXiao, line 550: Add const for blockHeight.

monstrobishi:
would this cause some inconsistency over the Anchor "view" between different nodes?

Bushstar:
@monstrobishi currently there is an inconsistent view between nodes that are online when accepting anchors and those that read them afterwards in a rare case. We assumed that time was only set when we had the block time, turns out time could be set to local system time so the pending anchor gets added before it is actually confirmed with local time. not blocktime. The problem arises when the anchor is deemed valid but afterwards when the actual block time comes in it is early then expected and the anchor fails on the three hours deep into the chain rule. Live nodes will accept this anchor on system and nodes reading Bitcoin anchors after confirmation will reject it on block time.

PR #314, opened 2021-04-12
Randomise winning anchor on blockhash

### LINKED ISSUE #306: Re-Design of DFI anchor system in a fair way

I am proposing a new re-design of the DFI anchor system to make it fair and transparent as it should be. 
At the moment a lot of users a not anchoring anymore due to some technical disadvantages on the user side.

DFI anchoring was designed to set anchor on the Bitcoin chain in a regular time intervals and at the convenient of an anchor-user. 
At the moment a lot of technical requirement is needed to protect your anchor against stealing anchor attempt. Stealing anchors is successfully performed at the moment on DFI chain. 

Anchoring was described in the Whitepaper https://defichain.com/white-paper/ and it got funded/activated at [DFIP#1](https://github.com/DeFiCh/dfips/issues/1) & [DFIP#2](https://github.com/DeFiCh/dfips/issues/2) and also documented here https://github.com/DeFiCh/ain/wiki/What-is-an-anchor%3F

Latest changed to the anchoring where done in https://github.com/DeFiCh/ain/pull/277. 

Currently we see the following anchoring happening on DFI & BTC blockchain: 

1. User#1 sends an valid anchor 
2. Another User#2 sends a bit later (between 2-10min) also an anchor which has a lower transaction hash value 
3. User#2 gets the anchor although he did send the anchor later then User#1. 

Based on public documentation about DFI anchoring, the first user should get the anchor independently of anyone sending another anchor later if User#1 gets confirmed first or at the same time as User#2. 

But., as we can see here https://github.com/DeFiCh/ain/blob/b4175b8041abdced250f7e98580effe9a6e248b5/src/masternodes/anchors.cpp#L659 its not following these `first-come-first serve`rules. 

#### Why is this needed:

To restore the balance in anchoring and have some more competing users and let them anchor whenever they think its appropriate.

Bushstar:
Randomise reward winner when paying out on transaction hash, resolves https://github.com/DeFiCh/ain/issues/306.

In case Bitcoin height and DeFi height are the same instead of rewarding whichever TX hash is lower first hash the TX hash with the Bitcoin block hash which cannot be determined at the point of anchor creation.

Also create AnchorToUniv for consistent printing of anchor information and reduce duplication of code.

### DIFF
--- a/src/masternodes/anchors.cpp
+++ b/src/masternodes/anchors.cpp
@@ -26,6 +26,7 @@ std::unique_ptr<CAnchorAwaitingConfirms> panchorAwaitingConfirms;
 
 static const char DB_ANCHORS = 'A';
 static const char DB_PENDING = 'p';
+static const char DB_BITCOININDEX = 'Z';  // Bitcoin height to blockhash table
 
 template <typename TContainer>
 bool CheckSigs(uint256 const & sigHash, TContainer const & sigs, std::set<CKeyID> const & keys)
@@ -526,7 +527,7 @@ void CAnchorIndex::CheckPendingAnchors()
 {
     AssertLockHeld(cs_main);
 
-    std::set<AnchorRec, decltype(OrderPendingAnchors)> anchorsPending(OrderPendingAnchors);
+    spv::PendingSet anchorsPending(spv::PendingOrder);
     ForEachPending([&anchorsPending](uint256 const &, AnchorRec & rec) {
         anchorsPending.insert(rec);
     });
@@ -546,11 +547,12 @@ void CAnchorIndex::CheckPendingAnchors()
             continue;
         }
 
-        uint32_t timestamp = spv::pspv->ReadTxTimestamp(rec.txHash);
-        auto blockHeight = spv::pspv->ReadTxBlockHeight(rec.txHash);
+        const auto timestamp = spv::pspv->ReadTxTimestamp(rec.txHash);
+        const auto blockHeight = spv::pspv->ReadTxBlockHeight(rec.txHash);
+        const auto blockHash = panchors->ReadBlockHash(rec.btcHeight);
 
         // Do not delete, TX time still pending. If block height is set to max we cannot trust the timestamp.
-        if (timestamp == 0 || blockHeight == std::numeric_limits<int32_t>::max()) {
+        if (timestamp == 0 || blockHeight == std::numeric_limits<int32_t>::max() || blockHash == uint256()) {
             continue;
         }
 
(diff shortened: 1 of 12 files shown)


src/masternodes/anchors.cpp
```diff
@@ -354,7 +355,30 @@ void CAnchorAuthIndex::PruneOlderThan(THeight height)
 }
 
 CAnchorIndex::CAnchorIndex(size_t nCacheSize, bool fMemory, bool fWipe)
-    : db(new CDBWrapper(GetDataDir() / "anchors", nCacheSize, fMemory, fWipe))
+    : db(new CDBWrapper(GetDataDir() / "anchors", nCacheSize, fMemory, fWipe)),
+      PendingOrder([](const CAnchorIndex::AnchorRec& a, const CAnchorIndex::AnchorRec& b)
```
bvbfan, line 359: It can be a free function

src/masternodes/anchors.cpp
```diff
@@ -652,13 +677,20 @@ CAnchorIndex::AnchorRec const * BestOfTwo(CAnchorIndex::AnchorRec const * a1, CA
...
+        const auto blockHash = panchors->ReadBlockHash(a1->btcHeight);
+        auto a1Hash = Hash(a1->txHash.begin(), a1->txHash.end(), blockHash.begin(), blockHash.end());
+        auto a2Hash = Hash(a2->txHash.begin(), a2->txHash.end(), blockHash.begin(), blockHash.end());
+        if (a1Hash < a2Hash) {
+            return a1;
+        }
+    } else if (a1->txHash < a2->txHash) {
         return a1;
+    }
     return a2;
```
bvbfan, line 694: `
return pendingAnchor(*a1, *a2) ? a1 : a2;
`

bvbfan (changes requested):

src/masternodes/anchors.cpp
Bushstar: Good idea. https://github.com/DeFiCh/ain/pull/314/commits/ec77e3d48055d71f716d3e111199f510505e1e05

src/masternodes/anchors.cpp
Bushstar: Thought that was not going to work, but it does. Check it out.

https://github.com/DeFiCh/ain/pull/314/commits/ec77e3d48055d71f716d3e111199f510505e1e05

bvbfan (approved):

ShengguangXiao (approved):

monstrobishi (approved):

PR #317, opened 2021-04-13
Minimize undo storage data

bvbfan:
/kind fix

It aims to minimize undo storage data by preventing entire pool pair data structure serialize

Fixes https://github.com/cakedefi/defichain-private/issues/603


monstrobishi:
Can we also add the ability to remove undos up to checkpoint in this PR?

monstrobishi (approved):

monstrobishi:
very nice

PR #323, opened 2021-04-14
Make account history own database

bvbfan:
/kind feature

Account history is optional feature it should not be part of blockchain directory, it will limit memory usage since it's flushed on every block.

Fixes https://github.com/cakedefi/defichain-private/issues/604

### DIFF
--- a/src/masternodes/accountshistory.cpp
+++ b/src/masternodes/accountshistory.cpp
@@ -7,9 +7,7 @@
 #include <masternodes/masternodes.h>
 #include <key_io.h>
 
-#include <limits>
-
-/// @attention make sure that it does not overlap with those in masternodes.cpp/tokens.cpp/undos.cpp/accounts.cpp !!!
+/// @Note it's in own database
 const unsigned char CAccountsHistoryView::ByAccountHistoryKey::prefix = 'h';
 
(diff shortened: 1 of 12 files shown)


monstrobishi (approved):

src/masternodes/accountshistory.cpp
```diff
@@ -7,18 +7,93 @@
...
+CAccountsHistoryWriter::CAccountsHistoryWriter(CCustomCSView & storage, uint32_t height, uint32_t txn, const uint256& txid, uint8_t type, CAccountsHistoryView* historyView)
+    : CStorageView(new CFlushableStorageKV(storage.GetRaw())), height(height), txn(txn), txid(txid), type(type), historyView(historyView)
+{
+}
+
+Res CAccountsHistoryWriter::AddBalance(CScript const & owner, CTokenAmount amount)
+{
+    auto res = CCustomCSView::AddBalance(owner, amount);
+    if (historyView && res.ok && amount.nValue != 0) {
+        diffs[owner][amount.nTokenId] += amount.nValue;
```
ShengguangXiao, line 44: Does it need to use SafeAdd?

bvbfan: `CCustomCSView::AddBalance` do `SafeAdd` so res will be not ok.

ShengguangXiao (approved):

PR #332, opened 2021-04-21
Split pool reward records

bvbfan:
/kind fix

Pool and custom rewards to be distinct history items

Fixes https://github.com/cakedefi/defichain-private/issues/393


monstrobishi:
Will this return 2 types of account history for custom rewards and regular rewards? Ideally they should both be "Rewards" from the user's perspective

bvbfan:
Yes it'll be distinct, so it should be 2 records but as same `Rewards` type?

monstrobishi:
Yes the format in listaccounthistory should show `Reward` in both cases, the user doesn't care how the reward was set on the backend

bvbfan:
Split records not types.

monstrobishi (approved):

PR #335, opened 2021-04-21
ICX Orderbook

### LINKED ISSUE #345: DeFiChain side orderbook for Interchain Exchange

Refer to pinkpaper for implementation details: https://github.com/DeFiCh/pinkpaper/tree/main/interchain-exchange

#### Why is this needed:

Interchain Exchange requires a robust set of custom transactions and database operations in order to fasciliate cross chain swap.

### LINKED ISSUE #350: Implement ICX incentive & swap fees

See pinkpaper for implementation details: https://github.com/DeFiCh/pinkpaper/blob/main/interchain-exchange/fees.md

#### Why is this needed:

In order to provide a fair and stable interchain exchange, careful use of fees and order of operations is important. The fees are in place to prevent abuse of the interchain swap system.

### LINKED ISSUE #381: Disclaimer for atomic swap and ICX usage

On any and all of the ICX and atomic swap endpoints, the following disclaimer should be shown:

> WARNING: ICX and Atomic Swap are experimental features. You might end up losing your funds. USE IT AT YOUR OWN RISK. 

#### Why is this needed:

ICX and Atomic Swap are introduced on Eunos as experimental features. Users should be made clearly aware of the risks.

Mixa84:

/kind feature

Adding ICX feature to Defichain. It is on chain orderbook and atomic swap of DFC tokens with HTLC with submitting info about external HTLC.

#### Additional comments?:
This is Work In Progress and it is not ready for merge (Lint, Tests missing)


src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
+#include <masternodes/icxorder.h>
+
+/// @attention make sure that it does not overlap with other views !!!
+const unsigned char CICXOrderView::ICXOrderCreationTx           ::prefix = '1';
+const unsigned char CICXOrderView::ICXOrderKey                  ::prefix = '2';
+const unsigned char CICXOrderView::ICXOrderStatus               ::prefix = 'B';
```
bvbfan, line 6: Duplicate, you should recheck with existing prefixes.

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+const unsigned char CICXOrderView::ICXMakeOfferCreationTx       ::prefix = '3';
+const unsigned char CICXOrderView::ICXMakeOfferKey              ::prefix = '4';
+const unsigned char CICXOrderView::ICXSubmitDFCHTLCCreationTx   ::prefix = '5';
+const unsigned char CICXOrderView::ICXSubmitDFCHTLCKey          ::prefix = '6';
+const unsigned char CICXOrderView::ICXSubmitDFCHTLCStatus       ::prefix = 'C';
```
bvbfan, line 11: Duplicate

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+const unsigned char CICXOrderView::ICXSubmitEXTHTLCCreationTx   ::prefix = '7';
+const unsigned char CICXOrderView::ICXSubmitEXTHTLCKey          ::prefix = '8';
+const unsigned char CICXOrderView::ICXClaimDFCHTLCCreationTx    ::prefix = '9';
+const unsigned char CICXOrderView::ICXClaimDFCHTLCKey           ::prefix = '0';
+const unsigned char CICXOrderView::ICXCloseOrderCreationTx      ::prefix = 'A';
```
bvbfan, line 16: Duplicate

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+const int CICXMakeOffer::STATUS_OPEN = 0;
+
+const int CICXSubmitDFCHTLC::DEFAULT_TIMEOUT = 100;
+const int CICXSubmitDFCHTLC::STATUS_OPEN = 0;
+const int CICXSubmitDFCHTLC::STATUS_CLAIMED = 1;
+const int CICXSubmitDFCHTLC::STATUS_REFUNDED = 2;
+
+const int CICXSubmitEXTHTLC::STATUS_OPEN = 0;
+const int CICXSubmitEXTHTLC::STATUS_EXPIRED = 1;
```
bvbfan, line 33: Can you use `uint8_t`

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+    OrderKey key({CICXOrder::STATUS_OPEN,{order.idToken,order.chain}},order.creationTx);
+    EraseBy<ICXOrderKey>(key);
+    key.first.first = status;
+    WriteBy<ICXOrderKey>(key, 1);
+    WriteBy<ICXOrderCreationTx>(order.creationTx, order);
+    
+    return {order.creationTx, Res::Ok()};
+}
+
+void CICXOrderView::ForEachICXOrder(std::function<bool (OrderKey const &, uint8_t)> callback, StatusAsset const & pair)
```
bvbfan, line 70: Seems unused, `ICXOrderKey` is needed?

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+{
+    //this should not happen, but for sure
+    if (GetICXOrderByCreationTx(order.creationTx)) {
+        return Res::Err("order with creation tx %s already exists!", order.creationTx.GetHex());
+    }
+
+    OrderKey key({CICXOrder::STATUS_OPEN,{order.idToken,order.chain}}, order.creationTx);
+    WriteBy<ICXOrderCreationTx>(order.creationTx, order);
+    WriteBy<ICXOrderKey>(key, 1);
+    WriteBy<ICXOrderStatus>(StatusKey(order.creationHeight+order.expiry,order.creationTx),CICXOrder::STATUS_EXPIRED);
```
bvbfan, line 54: Why we have status expired on creation?

src/masternodes/icxorder.h
```diff
@@ -0,0 +1,490 @@
...
+    typedef std::pair<DCT_ID,std::string> AssetPair;
+    typedef std::pair<uint8_t,AssetPair> StatusAsset;
+    typedef std::pair<StatusAsset,uint256> OrderKey;
+
+    typedef std::pair<uint256,uint256> TxidPairKey;
+
+    typedef std::pair<uint32_t,uint256> StatusTxid;
+    typedef std::pair<StatusTxid,uint256> StatusTxidKey;
+
+    typedef std::pair<uint32_t,uint256> StatusKey;
```
bvbfan, line 432: We should rethink these keys, probably we can optimize them.

bvbfan (changes requested):

src/masternodes/icxorder.cpp
Mixa84: This key in db is basically on which block height the order expires. I could save `1` in value but I wanted to be semantically correct so I set expired status which will become on that height.

src/masternodes/icxorder.h
Mixa84: Maybe there could be optimization as I refactored them several times while developing ICX. But most of them is necessary.

src/masternodes/icxorder.cpp
Mixa84: I think I had some warning in comparison with some unsigned ints, but will use uint8_t as it is one byte!

src/masternodes/icxorder.cpp
Mixa84: Didn't check overlap after rebase with 1.7.x. I will change them.

src/masternodes/icxorder.cpp
Mixa84: Didn't check overlap after rebase with 1.7.x. I will change them.

src/masternodes/icxorder.cpp
Mixa84: Didn't check overlap after rebase with 1.7.x. I will change them.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1079,406 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            mnview.ForEachICXSubmitDFCHTLC([&](CICXOrderView::StatusTxidKey const & key, uint8_t) {
+                if (key.first.second == submitexthtlc.offerTx)
+                {
+                    dfchtlc = mnview.GetICXSubmitDFCHTLCByCreationTx(key.second);
+                    return (false);
+                }
+                return false;
+            },{CICXSubmitDFCHTLC::STATUS_OPEN,submitexthtlc.offerTx});
+            if (!dfchtlc)
+                return Res::Err("offer %s needs to have dfc htlc submitted first, but no dfc htlc found!", submitexthtlc.offerTx.GetHex());
```
monstrobishi, line 1337: This error message is in both branches, is the flow for `TYPE_INTERNAL` supposed to be different?

Mixa84: For `TYPE_INTERNAL` orders (DFC/BTC) you must first submit DFC HTLC, and for `TYPE_EXTERNAL` you must first submit EXT HTLC.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1079,406 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            mnview.ForEachICXSubmitDFCHTLC([&found,&submitexthtlc](CICXOrderView::StatusTxidKey const & key, uint8_t) {
+                if (key.first.second == submitexthtlc.offerTx)
+                {
+                    found = true;
+                    return (false);
+                }
+                return true;
+            },{CICXSubmitDFCHTLC::STATUS_OPEN,offer->creationTx});
+            if (found)
+                return Res::Err("offer %s needs to have dfc htlc submitted first, but external htlc already submitted!", submitexthtlc.offerTx.GetHex());
```
monstrobishi, line 1367: @Mixa84 ok must be copy-paste error here then

Mixa84: You can see that two texts are different.
`offer %s needs to have dfc htlc submitted first, but external htlc already submitted!`
`offer %s needs to have dfc htlc submitted first, but no dfc htlc found!`
One (on line 1367) says it needs DFC HTLC first bout found EXT first, and the other (on line 1337) it needs DFC HTLC first but no DFC found!

monstrobishi: I need more clarity on this one then:

`offer %s needs to have dfc htlc submitted first, but external htlc already submitted!`

How can external already be submitted when it needs dfc first?

ShengguangXiao (dismissed):

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+{
+    //this should not happen, but for sure
+    if (GetICXClaimDFCHTLCByCreationTx(claimdfchtlc.creationTx)) {
+        return Res::Err("claimdfchtlc with creation tx %s already exists!", claimdfchtlc.creationTx.GetHex());
+    }
+
+    WriteBy<ICXClaimDFCHTLCCreationTx>(claimdfchtlc.creationTx, claimdfchtlc);
+    
+    EraseBy<ICXSubmitDFCHTLCKey>(StatusTxidKey({CICXSubmitDFCHTLC::STATUS_OPEN,dfchtlc.offerTx},dfchtlc.creationTx));
+    WriteBy<ICXSubmitDFCHTLCKey>(StatusTxidKey({CICXSubmitDFCHTLC::STATUS_CLAIMED,dfchtlc.offerTx},dfchtlc.creationTx),1);
```
ShengguangXiao, line 217: Add a blank after `,`

src/masternodes/icxorder.cpp
```diff
@@ -0,0 +1,248 @@
...
+const int CICXOrder::DEFAULT_EXPIRY = 2880;
+const int CICXOrder::TYPE_INTERNAL = 1;
+const int CICXOrder::TYPE_EXTERNAL = 2;
+const int CICXOrder::STATUS_OPEN = 0;
+const int CICXOrder::STATUS_CLOSED = 1;
+const int CICXOrder::STATUS_EXPIRED = 2;
+
+const int CICXMakeOffer::STATUS_OPEN = 0;
+
+const int CICXSubmitDFCHTLC::DEFAULT_TIMEOUT = 100;
```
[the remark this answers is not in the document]
Mixa84: Timeout in blocks. Will add comment.

monstrobishi:
Resolves #345

Mixa84:
Finally added complete refactor code.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1442 @@
...
+    if (!metaObj["expiry"].isNull()) makeoffer.expiry = metaObj["expiry"].get_int();
+
+    int targetHeight;
+    {
+        LOCK(cs_main);
+        auto order = pcustomcsview->GetICXOrderByCreationTx(makeoffer.orderTx);
+        if (!order)
+            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("orderTx (%s) does not exist",makeoffer.orderTx.GetHex()));
+
+        if (order->amountToFill * order->orderPrice / COIN < makeoffer.amount)
```
ShengguangXiao, line 383: `order->amountToFill * order->orderPrice` may overflow, need to convert to 256bits integer to calculate.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1442 @@
...
+        auto offer = pcustomcsview->GetICXMakeOfferByCreationTx(submitdfchtlc.offerTx);
+        if (!offer)
+            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("offerTx (%s) does not exist",submitdfchtlc.offerTx.GetHex()));
+        auto order = pcustomcsview->GetICXOrderByCreationTx(offer->orderTx);
+        if (!order)
+            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("orderTx (%s) does not exist",offer->orderTx.GetHex()));
+
+        if (order->orderType == CICXOrder::TYPE_INTERNAL)
+        {
+            if (submitdfchtlc.amount * order->orderPrice / COIN != offer->amount)
```
ShengguangXiao, line 540: `submitdfchtlc.amount * order->orderPrice` may overflow, need to convert to 256bits integer to calculate.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1442 @@
...
+            if (!dfchtlc)
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("offer (%s) needs to have dfc htlc submitted first, but no dfc htlc found!",
+                        submitexthtlc.offerTx.GetHex()));
+            if (submitexthtlc.hash != dfchtlc->hash)
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Invalid hash, external htlc hash is different than dfc htlc hash! - %s != %s!",
+                        submitexthtlc.hash.GetHex(),dfchtlc->hash.GetHex()));
+        }
+        else if (order->orderType == CICXOrder::TYPE_EXTERNAL)
+        {
+            if (submitexthtlc.amount * order->orderPrice / COIN != offer->amount )
```
ShengguangXiao, line 776: `submitdfchtlc.amount * order->orderPrice` may overflow

src/masternodes/rpc_icxorderbook.cpp
Mixa84: You are correct, will convert it!

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1443 @@
...
+
+    int targetHeight;
+    {
+        LOCK(cs_main);
+        auto order = pcustomcsview->GetICXOrderByCreationTx(makeoffer.orderTx);
+        if (!order)
+            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("orderTx (%s) does not exist",makeoffer.orderTx.GetHex()));
+
+        // if (order->amountToFill * order->orderPrice / COIN < makeoffer.amount)
+        //     throw JSONRPCError(RPC_INVALID_PARAMETER, "cannot make offer with that amount, order (" + order->creationTx.GetHex() + ") has less amount to fill!");
```
ShengguangXiao, line 385: Why remove this check?

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1115,422 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        
+        CAmount calcedAmount(static_cast<CAmount>((arith_uint256(order->amountToFill) * arith_uint256(order->orderPrice) / arith_uint256(COIN)).GetLow64()));
+        if (calcedAmount < makeoffer.amount)
+            return Res::Err("%s: %s", __func__, "cannot fill order with that amount, order (" + order->creationTx.GetHex() + ") has less amount to fill!");
+
+        CScript txidAddr(makeoffer.creationTx.begin(),makeoffer.creationTx.end());
+        
+        //calculating DFI per BTC
+        auto DFIBTCpair = mnview.GetPoolPair({5});
+        CAmount DFIperBTC = (arith_uint256(DFIBTCpair->reserveB) * arith_uint256(COIN) / DFIBTCpair->reserveA).GetLow64();
```
ShengguangXiao, line 1193: Get hardcoded poolpair 5? It crashes the node on my regtest, and the `DFIBTCpair` should be checked first?

src/masternodes/rpc_icxorderbook.cpp
Mixa84: I noticed that doing double checks in both rpc and mn_checks can lead to some bugs, because if the check is slightly different you can miss something. For example if you check x < 4 in rpc and x <= 4 in mn checks you won't be able to send tx with x == 4. This is not a big issue because if mn_checks is last check point and we can change RPC without consensus change. But I realized that I can miss some checks if I do it on both sides. So basically whatever is checked in mn_checks can be removed from RPC...

src/masternodes/mn_checks.cpp
Mixa84: Yeah, I haven't tested this yet. Just did the implementation first. I need to check for mainnet, testnet, devnet and regtest because pool pairs are different and on some they do not exist!

src/masternodes/icxorder.h
```diff
@@ -28,8 +29,8 @@ class CICXOrder
     CScript ownerAddress; //address of token asset in case of DFC/BTC order
     CAmount amountFrom; // amount of asset that is selled
     CAmount amountToFill; // how much is left to fill the order
```
bvbfan, line 34: `amountToFill` is `orderPrice - amountFrom` do we need separate field?

src/masternodes/icxorder.h
```diff
@@ -239,13 +259,13 @@ struct CICXSubmitDFCHTLCMessage : public CICXSubmitDFCHTLC {
...
-    static const int STATUS_EXPIRED;
+    static const uint8_t STATUS_OPEN;
 
     // This tx is acceptance of the offer and evidence of HTLC on external chain in the same time. It is a CustomTx on DFC chain
     //! basic properties
     uint256 offerTx; // txid for which offer is this HTLC
     CAmount amount;
+    CScript receiveAddress;
     uint256 hash; 
     std::string htlcscriptAddress;
```
bvbfan, line 284: Unused ?

src/masternodes/icxorder.h
```diff
@@ -239,13 +259,13 @@ struct CICXSubmitDFCHTLCMessage : public CICXSubmitDFCHTLC {
...
     CPubKey ownerPubkey;
```
bvbfan, line 285: Unused?

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -94,7 +126,7 @@ UniValue icxcreateorder(const JSONRPCRequest& request) {
                         {
                             {"tokenFrom|chainFrom", RPCArg::Type::STR, RPCArg::Optional::OMITTED, "Symbol or id of selling token/chain"},
```
bvbfan, line 143: I don't see reason to store chain name, it does not store any kind of value.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -146,33 +178,37 @@ UniValue icxcreateorder(const JSONRPCRequest& request) {
...
 
-        if (!metaObj["ownerAddress"].isNull()) order.ownerAddress = DecodeScript(metaObj["ownerAddress"].getValStr());
-        else throw JSONRPCError(RPC_INVALID_PARAMETER,"Invalid parameters, argument \"ownerAddress\" must be non-null if \"tokenFrom\" specified");
+        if (!metaObj["ownerAddress"].isNull())
+            order.ownerAddress = DecodeScript(metaObj["ownerAddress"].getValStr());
+        else
+            throw JSONRPCError(RPC_INVALID_PARAMETER,"Invalid parameters, argument \"ownerAddress\" must be non-null if \"tokenFrom\" specified");
        
-        if (!::IsMine(*pwallet, DecodeDestination(metaObj["ownerAddress"].getValStr()))) {
+        if (!::IsMine(*pwallet, DecodeDestination(metaObj["ownerAddress"].getValStr())))
```
bvbfan, line 208: Use `order.ownerAddress` in other places as well, you don't need to decode it second time.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -436,68 +479,124 @@ UniValue icxsubmitdfchtlc(const JSONRPCRequest& request) {
...
+
+            bool found = false;
+            pcustomcsview->ForEachICXSubmitDFCHTLCOpen([&found, &submitdfchtlc](CICXOrderView::TxidPairKey const & key, uint8_t i) {
+                if (key.first != submitdfchtlc.offerTx) 
+                    return false;
+                found = true;
+                return false;
+            }, submitdfchtlc.offerTx);
+            if (found)
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "dfc htlc already submitted!");
```
bvbfan, line 578: Make a simple function in the view `bool HasICXSubmitDFCHTLCOpen` in other places as well.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -436,68 +479,124 @@ UniValue icxsubmitdfchtlc(const JSONRPCRequest& request) {
...
+            found = false;
+            pcustomcsview->ForEachICXSubmitEXTHTLC([&found, &submitdfchtlc](CICXOrderView::TxidPairKey const & key, uint8_t i) {
+                if (key.first == submitdfchtlc.offerTx)
+                {
+                    found = true;
+                    return (false);
+                }
+                return false;
+            }, submitdfchtlc.offerTx);
+            if (found)
```
bvbfan: `HasICXSubmitEXTHTLC`

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -556,10 +654,11 @@ UniValue icxsubmitexthtlc(const JSONRPCRequest& request) {
                         {
                             {"offerTx", RPCArg::Type::STR, RPCArg::Optional::NO, "txid of offer tx for which the htlc is"},
                             {"amount", RPCArg::Type::NUM, RPCArg::Optional::NO, "amount in htlc"},
+                            {"receiveAddress", RPCArg::Type::NUM, RPCArg::Optional::OMITTED, "address for receiving DFC tokens in case of EXT/BTC order type"},
                             {"htlcScriptAddress", RPCArg::Type::STR, RPCArg::Optional::NO, "script address of external htlc"},
                             {"hash", RPCArg::Type::STR, RPCArg::Optional::NO, "hash of seed used for the hash lock part"},
                             {"ownerPubkey", RPCArg::Type::STR, RPCArg::Optional::NO, "pubkey of the owner to which the funds are refunded if HTLC timeouts"},
```
bvbfan: What's the point of `htlcScriptAddress` and `ownerPubkey` since they are unused?

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -578,62 +677,131 @@ UniValue icxsubmitexthtlc(const JSONRPCRequest& request) {
...
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("cannot make ext htlc with that amount, different amount necessary for offer (%s) - %s != %s!",
+                                offer->creationTx.GetHex(), ValueFromAmount(submitexthtlc.amount).getValStr(), ValueFromAmount(offer->amount).getValStr()));
+
+            bool found = false;
+            pcustomcsview->ForEachICXSubmitEXTHTLC([&found, &submitexthtlc](CICXOrderView::TxidPairKey const & key, uint8_t i) {
+                if (key.first != submitexthtlc.offerTx)
+                    return false;
+                found = true;
+                return false;
+            }, submitexthtlc.offerTx);
```
bvbfan: `HasICXSubmitEXTHTLC`

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -537,9 +538,9 @@ UniValue icxsubmitdfchtlc(const JSONRPCRequest& request) {
 
         if (order->orderType == CICXOrder::TYPE_INTERNAL)
         {
-            if (submitdfchtlc.amount * order->orderPrice / COIN != offer->amount)
-                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("cannot make dfc htlc with that amount, different amount necessary for offer (%s) - %s != %s!",
-                                offer->creationTx.GetHex(), ValueFromAmount(submitdfchtlc.amount * order->orderPrice / COIN).getValStr(), ValueFromAmount(offer->amount).getValStr()));
+            // if (submitdfchtlc.amount * order->orderPrice / COIN != offer->amount)
+            //     throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("cannot make dfc htlc with that amount, different amount necessary for offer (%s) - %s != %s!",
+            //                     offer->creationTx.GetHex(), ValueFromAmount(submitdfchtlc.amount * order->orderPrice / COIN).getValStr(), ValueFromAmount(offer->amount).getValStr()));
```
bvbfan, line 560: Commented code?

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -773,9 +774,9 @@ UniValue icxsubmitexthtlc(const JSONRPCRequest& request) {
         }
         else if (order->orderType == CICXOrder::TYPE_EXTERNAL)
         {
-            if (submitexthtlc.amount * order->orderPrice / COIN != offer->amount )
-                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("cannot make ext htlc with that amount, different amount necessary for offer (%s) - %s != %s!",
-                                offer->creationTx.GetHex(), ValueFromAmount(submitexthtlc.amount * order->orderPrice / COIN).getValStr(), ValueFromAmount(offer->amount).getValStr()));
+            // if (submitexthtlc.amount * order->orderPrice / COIN != offer->amount )
+            //     throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("cannot make ext htlc with that amount, different amount necessary for offer (%s) - %s != %s!",
+            //                     offer->creationTx.GetHex(), ValueFromAmount(submitexthtlc.amount * order->orderPrice / COIN).getValStr(), ValueFromAmount(offer->amount).getValStr()));
```
bvbfan, line 814: Commented code?

bvbfan (changes requested):
Basically, validate parameters on rpc but not check balances there (`GetBalance` should be in nm_checks)
Also tune you editor, you have a lot of space characters added where it shouldn't.
`OrderKey` has chain name which is never used, why we need to waste space for that?
`TxidPairKey` is pair of 2 txs second one is never used
`StatusKey` is unoptimized
Just make a key, can be a tuple or struct, but pair are unwanted, You have more records, that means more undos and db goes to high for nothing.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1079,406 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+
+        if (order->orderType == CICXOrder::TYPE_INTERNAL)
+        {
+            if (makeoffer.receiveDestination.empty())
+                return Res::Err("receiveAddress must not be null!");
+        }
+        else
+        {
+            if (!CPubKey(makeoffer.receiveDestination).IsFullyValid())
+                return Res::Err("%s: %s", __func__, "Invalid receivePubKey, (" + HexStr(makeoffer.receiveDestination) + ") receivePubkey is not a valid pubkey!");
```
bvbfan, line 1217: Remove all reference to `__func__`. It's not needed anymore, just return a string it will be constructed for concrete function.

bvbfan (changes requested):

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1497 @@
...
+             << submitdfchtlc;
+
+    CScript scriptMeta;
+    scriptMeta << OP_RETURN << ToByteVector(metadata);
+
+    const auto txVersion = GetTransactionVersion(targetHeight);
+    CMutableTransaction rawTx(txVersion);
+
+    CTransactionRef optAuthTx;
+    std::set<CScript> auths;
```
Bushstar, line 650: Do we need this an the subsequent auths.size() check? If not can be removed, auths in arg to GetAuthInputsSmart can be {} and auths.size() in conditional below can be removed.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1497 @@
...
+    CTransactionRef optAuthTx;
+    std::set<CScript> auths{authScript};
+    rawTx.vin = GetAuthInputsSmart(pwallet, rawTx.nVersion, auths, false, optAuthTx, txInputs);
+
+    rawTx.vout.push_back(CTxOut(0, scriptMeta));
+
+    CCoinControl coinControl;
+
+    // Return change to auth address
+    if (auths.size() == 1) {
```
Bushstar, line 448: Conditional may be redundant as auth always one?

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1497 @@
...
+             << submitexthtlc;
+
+    CScript scriptMeta;
+    scriptMeta << OP_RETURN << ToByteVector(metadata);
+
+    const auto txVersion = GetTransactionVersion(targetHeight);
+    CMutableTransaction rawTx(txVersion);
+
+    CTransactionRef optAuthTx;
+    std::set<CScript> auths;
```
Bushstar, line 861: Do we need this an the subsequent auths.size() check? If not can be removed, auths in arg to GetAuthInputsSmart can be {} and auths.size() in conditional below can be removed.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1497 @@
...
+             << claimdfchtlc;
+
+    CScript scriptMeta;
+    scriptMeta << OP_RETURN << ToByteVector(metadata);
+
+    const auto txVersion = GetTransactionVersion(targetHeight);
+    CMutableTransaction rawTx(txVersion);
+
+    CTransactionRef optAuthTx;
+    std::set<CScript> auths;
```
Bushstar, line 981: Do we need this an the subsequent auths.size() check? If not can be removed, auths in arg to GetAuthInputsSmart can be {} and auths.size() in conditional below can be removed.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1497 @@
...
+             << order;
+
+    CScript scriptMeta;
+    scriptMeta << OP_RETURN << ToByteVector(metadata);
+
+    const auto txVersion = GetTransactionVersion(targetHeight);
+    CMutableTransaction rawTx(txVersion);
+
+    CTransactionRef optAuthTx;
+    std::set<CScript> auths;
```
Bushstar, line 283: No auth set here for the order, is that correct? I'd expect an address to be set as auth here to allow the auth address to be able to close the order with closeorder with an input that corresponds to it.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1497 @@
...
+             << closeorder;
+
+    CScript scriptMeta;
+    scriptMeta << OP_RETURN << ToByteVector(metadata);
+
+    const auto txVersion = GetTransactionVersion(targetHeight);
+    CMutableTransaction rawTx(txVersion);
+
+    CTransactionRef optAuthTx;
+    std::set<CScript> auths;
```
Bushstar, line 1080: No auth set, how do we authorise a close order TX? Do we need to set an auth on open order and make a corresponding auth here.

Bushstar (commented):
We need to check auths, I think to remove unused code in certain places and add auth to open and close order.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1114,443 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            return Res::Err("order amountToFill does not equal to amountFrom!");
+        if (order.orderPrice == 0)
+            return Res::Err("order price must be greater than 0!");
+        if (order.expiry == 0)
+            return Res::Err("order expiry must be greater than 0!");
+
+        return mnview.ICXCreateOrder(order);
+    }
+
+    Res operator()(const CICXMakeOfferMessage& obj) const {
```
Bushstar, line 1169: Since CloseOffer requires auth of create offer TX vout 1, we should probably make sure this TX has the correct auth input with HasAuth set to receive or owner address.

test/functional/feature_icx_orderbook.py
```diff
@@ -0,0 +1,342 @@
...
+# file LICENSE or http://www.opensource.org/licenses/mit-license.php.
+"""Test ICX Orderbook."""
+
+from test_framework.test_framework import DefiTestFramework
+
+from test_framework.util import assert_equal
+
+from decimal import Decimal
+
+class ICXOrderbookTest (DefiTestFramework):
```
Bushstar, line 14: Needs tests for icx_closeorder, icx_closeoffer and icx_getorder. Probably best to have a TX between the close tests, first time I tested close order it worked as the change used on make order which became of the Auth was then used as the input for the close order by chance and appeared to work.

src/masternodes/icxorder.h
Mixa84: amountToFill is a value how much is left to fill the order, and is updated as order is partially filled. If we don't want to keep what was the amount of the order when it was opened, then I can subtract from amountFrom directly!

src/masternodes/icxorder.h
Mixa84: Used in case of external type of order - BTC order comes first.

But, I realized recently that only one address is used for all tokens and I wanted to refactor the flow so that in order will be token address and optional btc pubkey for BTC/DFI order and in offer to have also token address and optional BTC pubkey for DFI/BTC order. Then there is no need for addresses in HTLCs!

src/masternodes/rpc_icxorderbook.cpp
Mixa84: I have introduced chain name as I made this not only for BTC, but it can be any BTC protocol chain.
If we are going to do only BTC and it will be necessary to make hardfork in order to introduce new chain then I can remove chain name and bind it only to BTC!

src/masternodes/rpc_icxorderbook.cpp
Mixa84: They are not unused, but they are not used in Defichain. Those data is necessary so that the other party can create BTC HTLC on his side in order to monitor it and claim it. So he needs: htlc script address, owner pubkey, his pubkey (he already has his own) and timeout values to recreate it.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1114,443 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        if (calcedAmount < makeoffer.amount)
+            makeoffer.amount = calcedAmount;
+
+        CScript txidAddr(makeoffer.creationTx.begin(),makeoffer.creationTx.end());
+        
+        //calculating DFI per BTC
+        auto id = mnview.ICXGetDFIBTCPoolPairId(height);
+        auto DFIBTCPoolPair = mnview.GetPoolPair(id);
+        CAmount DFIperBTC = CICXOrderView::DEFAULT_DFI_BTC_PRICE;
+        if (DFIBTCPoolPair)
```
ShengguangXiao, line 1194: If didn't get the poolpair, do we still need to continue? I think we should just return an error and stop here.

src/masternodes/icxorder.h
Mixa84: Same as above.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1114,443 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+
+            srcAddr = offer->ownerAddress;
+        }
+
+        // subtract the balance from order/offer txidaddr and dedicate them for the dfc htlc
+        CScript htlcTxidAddr(submitdfchtlc.creationTx.begin(), submitdfchtlc.creationTx.end());
+        res = ModifyTokenBalance(order->idToken, submitdfchtlc.amount, srcAddr, htlcTxidAddr);
+            if (!res.ok)
+                return Res::Err("%s: %s", __func__, res.msg);
+
```
ShengguangXiao: Please align the code.

src/masternodes/icxorder.h
bvbfan: I don't get what you mean directly? You have price and the amount you have till now, amountToFill is always price - amount

src/validation.cpp
```diff
@@ -2603,6 +2608,180 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
+                if (status == CICXSubmitEXTHTLC::STATUS_EXPIRED && order->orderType == CICXOrder::TYPE_INTERNAL)
+                {
+                    std::unique_ptr<CICXSubmitDFCHTLCImplemetation> dfchtlc;
+                    cache.ForEachICXSubmitDFCHTLCOpen([&](CICXOrderView::TxidPairKey const & key, uint8_t i) {
+                        if (key.first == exthtlc->offerTx)
+                        {
+                            dfchtlc = pcustomcsview->GetICXSubmitDFCHTLCByCreationTx(key.second);
+                            return (false);
+                        }
+                        return false;
```
ShengguangXiao, line 2763: This lambda function always return false, so only search one item?

src/masternodes/mn_checks.cpp
Mixa84: We have a default price for `DFIperBTC` for this case. But there should be gov variable that will set pool pair id.

src/masternodes/mn_checks.cpp
```diff
@@ -1029,6 +1114,443 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        CAmount DFIperBTC = CICXOrderView::DEFAULT_DFI_BTC_PRICE;
+        if (DFIBTCPoolPair)
+            DFIperBTC = (arith_uint256(DFIBTCPoolPair->reserveA) * arith_uint256(COIN) / DFIBTCPoolPair->reserveB).GetLow64();
+        if (order->orderType == CICXOrder::TYPE_INTERNAL)
+        {
+            if (makeoffer.receiveDestination.empty())
+                return Res::Err("receiveAddress must not be null!");
+            
+            // calculating and locking takerFee in offer txidaddr
+            if (DFIperBTC)
```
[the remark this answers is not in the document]
Mixa84: This is left over from initial implementation where I set default price 0 and if it does not get the price then it was not taking the takerFee at all. Will remove it!

src/validation.cpp
Mixa84: Yes, this is intended behavior. It should check if DFC or in below cased EXT HTLC exists. So if it doesn't exists it returns maker deposit to the maker because taker didn't create his htlc in period of 100 DFC blocks.

src/masternodes/mn_checks.cpp
Mixa84: Already in mn_checks.cpp

src/masternodes/rpc_icxorderbook.cpp
Mixa84: No, claim does not need auth as the other side is claiming htlc. Removed!

src/masternodes/icxorder.h
Mixa84: amountFrom is amount of asset selling - 100 DFI
orderPrice is price of 1 asset in opposite asset - 1 DFI = 0.00015 BTC
amountToFill is how much is left to fill - 100 DFI initially, when offer fills 35 DFI then amountToFill will be 65 DFI.

I can remove redundant data if we don't need to save initial amount of order (100 DFI) and then I can sub from amountFrom when successful trade is done. Also I could store orderPrice as price * amount so I have already calced amount how much it is in opposite asset and not calculating it every time.

src/masternodes/mn_checks.cpp
```diff
@@ -398,6 +447,54 @@ class CCustomTxVisitor : public boost::static_visitor<Res>
...
+            return true;
+        }, DCT_ID{1});
+        assert(res.v != 0);
+        return res;
+    }
+
+    CPoolPair GetDFIBTCPoolPair() const {
+        auto BTC = FindTokenByPartialSymbolName(CICXOrder::TOKEN_BTC);
+        auto pair = mnview.GetPoolPair(DCT_ID{0}, BTC);
+        assert(pair);
```
ShengguangXiao, line 494: Can return some fail here? assert is not compiled in release mode

src/masternodes/mn_checks.cpp
```diff
@@ -398,6 +447,54 @@ class CCustomTxVisitor : public boost::static_visitor<Res>
...
 
+    Res CheckICXTx() const {
+        if (tx.vout.size() != 2) {
+            return Res::Err("malformed tx vouts ((wrong number of vouts)");
+        }
+        return Res::Ok();
+    }
+
+    Res ICXTransfer(DCT_ID id, CAmount amount, CScript const & from, CScript const & to) const {
+        assert(!from.empty() || !to.empty());
```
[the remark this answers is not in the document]
bvbfan: It's internal function, the caller takes the care to pass valid arguments and it should.

src/masternodes/mn_checks.cpp
```diff
@@ -398,6 +447,54 @@ class CCustomTxVisitor : public boost::static_visitor<Res>
...
+    DCT_ID FindTokenByPartialSymbolName(const std::string& symbol) const {
+        DCT_ID res{0};
+        mnview.ForEachToken([&](DCT_ID id, CTokenImplementation token) {
+            if (token.symbol.find(symbol) == 0) {
+                res = id;
+                return false;
+            }
+            return true;
+        }, DCT_ID{1});
+        assert(res.v != 0);
```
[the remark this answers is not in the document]
bvbfan: Same as above.

src/masternodes/mn_checks.cpp
Mixa84: BTC does not exist on regtest, but we should have tests create BTC in the beginning. But if someone trys to do something without creating BTC token it will fail.

bvbfan: Sure, if BTC does not exist it's an unusual env. Assert enforces correct env.

src/validation.cpp
```diff
@@ -2705,7 +2715,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                     CScript ownerAddress;
                     amount = {order->idToken, dfchtlc->amount};
                     if (order->orderType == CICXOrder::TYPE_INTERNAL)
-                        ownerAddress = order->ownerAddress;
+                        ownerAddress = CScript(order->creationTx.begin(),order->creationTx.end());
```
ShengguangXiao, line 2718: Add blank after `,`

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -775,6 +778,9 @@ UniValue icxsubmitexthtlc(const JSONRPCRequest& request) {
 
         targetHeight = ::ChainActive().Height() + 1;
 
+        if (order->creationHeight + order->expiry < targetHeight + (submitexthtlc.timeout * 6 / 100))
```
ShengguangXiao, line 781: Can define 6 and 100 as const variables? I see it used somewhere else too.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1479 @@
...
+        LOCK(cs_main);
+        CCoinsViewCache coinview(&::ChainstateActive().CoinsTip());
+        if (optAuthTx)
+            AddCoins(coinview, *optAuthTx, targetHeight);
+        const auto metadata = ToByteVector(CDataStream{SER_NETWORK, PROTOCOL_VERSION, order});
+        execTestTx(CTransaction(rawTx), targetHeight, metadata, CICXCreateOrderMessage{}, coinview);
+    }
+    UniValue ret(UniValue::VOBJ);
+    ret.pushKV("WARNING", "ICX and Atomic Swap are experimental features. You might end up losing your funds. USE IT AT YOUR OWN RISK.");
+    ret.pushKV("result",signsend(rawTx, pwallet, optAuthTx)->GetHash().GetHex());
```
monstrobishi, line 355: Since this is already contained in an object called "result", can you rename this property to "txid"? Same for the other RPC calls

monstrobishi (approved):

monstrobishi (changes requested):
Comment above

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1479 @@
...
+            if (submitdfchtlc.amount != calcedAmount)
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("cannot make dfc htlc with that amount, amount must be equal to calculated exthtlc amount - %s != %s",
+                                ValueFromAmount(submitdfchtlc.amount).getValStr(), ValueFromAmount(calcedAmount).getValStr()));
+
+            if (submitdfchtlc.hash != exthtlc->hash)
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Invalid hash, dfc htlc hash is different than extarnal htlc hash - %s != %s",
+                        submitdfchtlc.hash.GetHex(),exthtlc->hash.GetHex()));
+            if (submitdfchtlc.timeout < CICXSubmitDFCHTLC::MINIMUM_2ND_TIMEOUT)
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Invalid parameters, argument \"timeout\" must be greater than %d", CICXSubmitEXTHTLC::MINIMUM_2ND_TIMEOUT - 1));
+            if (submitdfchtlc.timeout >= (exthtlc->creationHeight + (exthtlc->timeout * 100 / 6 )) - targetHeight)
```
monstrobishi, line 638: I'm curious why we are using `100 / 6` here, DFC block is every 30 seconds, and bitcoin block is every 10 minutes, shouldn't it be a 20 <-> 1 conversion ratio?

Note that we are altering the difficulty adjustment rate in this release too, so we are expecting the DFC blocktime to stabilise at 30 seconds instead of the current 37 seconds.

Mixa84: I was using this as per notes we had from two weeks ago. I also checked with U-Zyn should I use 100 blocks per hour for DFI.
I can change this just let me know which is the number we are targeting. This is only theoretically as we cannot know how much blocks exactly will be in an hour.

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1479 @@
...
+    RPCHelpMan{"icx_submitdfchtlc",
+                "\nEXPERIMENTAL warning: ICX and Atomic Swap are experimental features. You might end up losing your funds. USE IT AT YOUR OWN RISK.\n\nCreates (and submits to local node and network) a dfc htlc transaction.\n" +
+                HelpRequiringPassphrase(pwallet) + "\n",
+                {
+                    {"htlc", RPCArg::Type::OBJ, RPCArg::Optional::OMITTED, "",
+                        {
+                            {"offerTx", RPCArg::Type::STR, RPCArg::Optional::NO, "txid of offer tx for which the htlc is"},
+                            {"amount", RPCArg::Type::NUM, RPCArg::Optional::NO, "amount in htlc"},
+                            {"hash", RPCArg::Type::STR, RPCArg::Optional::NO, "hash of seed used for the hash lock part"},
+                            {"timeout", RPCArg::Type::NUM, RPCArg::Optional::OMITTED, "timeout (absolute in blocks) for expiration of htlc"},
```
ShengguangXiao, line 520: Can add descrition the timeout is DFI blocks or BTC blocks?

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1479 @@
...
+                HelpRequiringPassphrase(pwallet) + "\n",
+                {
+                    {"htlc", RPCArg::Type::OBJ, RPCArg::Optional::OMITTED, "",
+                        {
+                            {"offerTx", RPCArg::Type::STR, RPCArg::Optional::NO, "txid of offer tx for which the htlc is"},
+                            {"amount", RPCArg::Type::NUM, RPCArg::Optional::NO, "amount in htlc"},
+                            {"htlcScriptAddress", RPCArg::Type::STR, RPCArg::Optional::NO, "script address of external htlc"},
+                            {"hash", RPCArg::Type::STR, RPCArg::Optional::NO, "hash of seed used for the hash lock part"},
+                            {"ownerPubkey", RPCArg::Type::STR, RPCArg::Optional::NO, "pubkey of the owner to which the funds are refunded if HTLC timeouts"},
+                            {"timeout", RPCArg::Type::NUM, RPCArg::Optional::NO, "timeout (absolute in block) for expiration of external htlc"},
```
ShengguangXiao, line 700: Can add descrition the timeout is DFI blocks or BTC blocks?

src/masternodes/rpc_icxorderbook.cpp
```diff
@@ -0,0 +1,1479 @@
...
+        CCoinsViewCache coinview(&::ChainstateActive().CoinsTip());
+        if (optAuthTx)
+            AddCoins(coinview, *optAuthTx, targetHeight);
+        auto metadata = ToByteVector(CDataStream{SER_NETWORK, PROTOCOL_VERSION, submitexthtlc});
+        execTestTx(CTransaction(rawTx), targetHeight, metadata, CICXSubmitEXTHTLCMessage{}, coinview);
+    }
+
+    UniValue ret(UniValue::VOBJ);
+    ret.pushKV("WARNING", "ICX and Atomic Swap are experimental features. You might end up losing your funds. USE IT AT YOUR OWN RISK.");
+    ret.pushKV("txid",signsend(rawTx, pwallet, optAuthTx)->GetHash().GetHex());
```
ShengguangXiao, line 869: Add blank after `,`

src/masternodes/rpc_icxorderbook.cpp
monstrobishi: @Mixa84 this isn't resolved

ShengguangXiao: @monstrobishi It is resolved, please pull the latest commits.

bvbfan (dismissed):

monstrobishi (dismissed):

Bushstar (approved):

PR #337, opened 2021-04-22
Extend block merkle root by hash of account changes

### LINKED ISSUE #349: Include account system changes in block merkle root

Include account system changes in block merkle root, by calculating the entire account state, hashing it and including it as part of the merkle root.

#### Why is this needed:

Account system state is not currently enforced strongly by the information in the block, encoding the account state as part of the merkle root ensures that account changes are reflected and strongly enforced at every block.

bvbfan:
/kind feature

It ensures changes made by txs in account system layer are validated. It gains performance improvement static_cast over dynamic_cast, we know at compile time the storage time, so we don't need to use RTTI.

It's WIP since i can make better testing, unit / functional / (manual ?)

Fixes #349


monstrobishi:
Resolves #349

src/masternodes/masternodes.cpp
```diff
@@ -768,6 +769,19 @@ bool CCustomCSView::CalculateOwnerRewards(CScript const & owner, uint32_t target
     return UpdateBalancesHeight(owner, targetHeight);
 }
 
+uint256 CCustomCSView::MerkleRoot() {
+    auto& rawMap = GetStorage().GetRaw();
```
ShengguangXiao, line 773: Use `const auto&`

PR #366, opened 2021-05-05
[WIP] Feature/operator

### LINKED ISSUE #348: Implement Operator model: createoperator and updateoperator

Please see for implementation details: https://github.com/DeFiCh/pinkpaper/tree/main/operator

Operators will take over tasks that require foundation authorization

RPC calls need to be modified to take operator argument

#### Why is this needed:

As the blockchain becomes more decentralised, operators will need to take over the role of managing the blockchain.

surangap:

/kind feature

Fixes #348

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -125,6 +125,8 @@ public:
         consensus.DakotaHeight = 678000; // 1st March 2021
         consensus.DakotaCrescentHeight = 733000; // 25th March 2021
         consensus.EunosHeight = std::numeric_limits<int>::max();
+        consensus.FUpgradeHeight = std::numeric_limits<int>::max();
+
 
         consensus.pos.diffLimit = uint256S("00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
 //        consensus.pos.nTargetTimespan = 14 * 24 * 60 * 60; // two weeks
@@ -342,6 +344,7 @@ public:
         consensus.DakotaHeight = 220680;
         consensus.DakotaCrescentHeight = 287700;
         consensus.EunosHeight = std::numeric_limits<int>::max();
+        consensus.FUpgradeHeight = std::numeric_limits<int>::max();
 
         consensus.pos.diffLimit = uint256S("00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
 //        consensus.pos.nTargetTimespan = 14 * 24 * 60 * 60; // two weeks
@@ -520,6 +523,7 @@ public:
         consensus.DakotaHeight = 10;
         consensus.DakotaCrescentHeight = 10;
         consensus.EunosHeight = 10;
+        consensus.FUpgradeHeight = 10;
 
         consensus.pos.diffLimit = uint256S("00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.pos.nTargetTimespan = 5 * 60; // 5 min == 10 blocks
@@ -690,6 +694,7 @@ public:
         consensus.DakotaHeight = 10000000;
         consensus.DakotaCrescentHeight = 10000000;
         consensus.EunosHeight = 10000000;
+        consensus.FUpgradeHeight = 10000000;
 
         consensus.pos.diffLimit = uint256S("00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.pos.nTargetTimespan = 14 * 24 * 60 * 60; // two weeks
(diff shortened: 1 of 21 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -329,6 +334,18 @@ class CCustomMetadataParseVisitor : public boost::static_visitor<Res>
         return !res ? res : serialize(obj);
     }
 
+    Res operator()(CCreateOperatorMessage& obj) const {
```
monstrobishi, line 337: Add hardfork height check here

src/masternodes/mn_checks.cpp
```diff
@@ -329,6 +334,18 @@ class CCustomMetadataParseVisitor : public boost::static_visitor<Res>
...
+        //auto res = isPostEunosFork();
+        auto res = Res::Ok();
+        return !res ? res : serialize(obj);
+    }
+
+    Res operator()(CUpdateOperatorMessage& obj) const {
```
monstrobishi, line 343: And here

src/chainparams.cpp
```diff
@@ -472,6 +477,11 @@ class CTestNetParams : public CChainParams {
             /* nTxCount */ 178351,
             /* dTxRate  */ 0.03842042178237066
         };
+
```
monstrobishi, line 480: Add a new hardfork height to chain params, we won't use Eunos for this, call it FUpgradeHeight for now

ShengguangXiao:
I think still need to add `listoperators` and `getoperator` rpc.

src/masternodes/operators.h
```diff
@@ -30,7 +30,16 @@ struct CCreateOperatorMessage {
     CScript operatorAddress;
     std::string operatorName;
     std::string operatorURL;
-    // OperatorState operatorState;
+    uint8_t operatorState{0};
+
+    CCreateOperatorMessage(CScript operatorAddress_, std::string operatorName_, std::string operatorURL_, OperatorState operatorState_)
+                            :  operatorAddress(operatorAddress_)
+                            ,  operatorName(operatorName_)
+                            ,  operatorURL(operatorURL_)
```
bvbfan, line 38: Use `std::move` for these

src/masternodes/operators.h
```diff
@@ -40,14 +49,20 @@ struct CCreateOperatorMessage {
...
 };
 
 struct CUpdateOperatorMessage {
     COperatorId operatorId;
     CCreateOperatorMessage newOperator;
 
+    CUpdateOperatorMessage() = default;
+
+    CUpdateOperatorMessage(COperatorId operatorId_, CCreateOperatorMessage newOperator_) 
+                            : operatorId(operatorId_), newOperator(newOperator_)
```
bvbfan, line 63: `std::move`

bvbfan (changes requested):

surangap:
> I think still need to add `listoperators` and `getoperator` rpc.

yes. initial work is to add `createoperator` and `updateoperator`. But Since i have already written underlying functions for retrieval, I should add those. @monstrobishi shall i add those today before starting next task. it would be quick. and then i can complete the functional test also.

src/masternodes/operators.h
surangap: was using `std::forward` for this. But seems it is not perfect for this situation(when lval is passed).
will change to `std::move`with the next commit, better.

src/masternodes/operators.cpp
```diff
@@ -33,10 +33,11 @@ Res COperatorView::UpdateOperator(const COperatorId& operatorId, const COperator
...
     }
 
+    //move everithing one by one(if possible) for the clarity.
     opertr.operatorAddress = std::move(newOperator.operatorAddress);
-    opertr.operatorName = newOperator.operatorName;
-    opertr.operatorURL = newOperator.operatorURL;
-    opertr.operatorState = newOperator.operatorState;
+    opertr.operatorName = std::move(newOperator.operatorName);
+    opertr.operatorURL = std::move(newOperator.operatorURL);
+    opertr.operatorState = std::move(newOperator.operatorState);
```
bvbfan: Let me explain a bit. It was introduced by oracle PR but moving from const object can not be done. You have `const COperator` so moving its members are unaffected. You should declare `newOperator` non const.

src/masternodes/operators.h
```diff
@@ -33,9 +33,9 @@ struct CCreateOperatorMessage {
     uint8_t operatorState{0};
 
     CCreateOperatorMessage(CScript operatorAddress_, std::string operatorName_, std::string operatorURL_, OperatorState operatorState_)
-                            :  operatorAddress(operatorAddress_)
-                            ,  operatorName(operatorName_)
-                            ,  operatorURL(operatorURL_)
+                            :  operatorAddress(std::forward<CScript>(operatorAddress_))
+                            ,  operatorName(std::forward<std::string>(operatorName_))
+                            ,  operatorURL(std::forward<std::string>(operatorURL_))
```
bvbfan, line 38: It should be `move`. `forward` was introduced to handle a situation when you can have const and non-const ref say
```
temlpate <typename T>
void wrapper(T&& t)
```
In this context `t` can be type rvalue or lvalue, `forward` ensures no copies.
https://en.cppreference.com/w/cpp/utility/forward

bvbfan (changes requested):

src/masternodes/operators.cpp
surangap: Aha it's a const ref. thanks for pointing out. make sense. But i think we should keep it as const ref. otherwise, it means you can change the incoming object properties inside the function.

src/masternodes/operators.h
surangap: https://github.com/DeFiCh/ain/pull/366#discussion_r629113157

PR #369, opened 2021-05-05
Do not allow checkpoint height to be reverted

bvbfan:
/kind fix

It's a functional test to ensure longer chain that overlap last checkpoint is rejected, including the checkpoint height


ShengguangXiao (approved):

src/validation.cpp
```diff
@@ -3755,7 +3755,7 @@ static bool ContextualCheckBlockHeader(const CBlockHeader& block, CValidationSta
         // GetLastCheckpoint finds the last checkpoint in MapCheckpoints that's in our
         // g_blockman.m_block_index.
         CBlockIndex* pcheckpoint = GetLastCheckpoint(params.Checkpoints());
-        if (pcheckpoint && nHeight < pcheckpoint->nHeight)
+        if (pcheckpoint && nHeight <= pcheckpoint->nHeight)
```
monstrobishi, line 3758: Could this have any effects?

ShengguangXiao: In some rare case, it may cause a fork?

bvbfan: Longest chain that overlap checkpoint (the equal is to prevent overlap exactly checkpoint height, since it's verified by us) so node provide that chain will fork since its chain isn't accepted. See https://github.com/DeFiCh/ain/pull/369/commits/3c30b56b09d951d592247823f8b992702300eb3a#diff-fe7777e94729859264eb9950e085807449c5cadc091498284a92966960e37e30R51

ShengguangXiao (dismissed):

monstrobishi (dismissed):

monstrobishi (approved):

PR #371, opened 2021-05-07
Defer chain lock to caller to allow atomic lock

### LINKED ISSUE #368: Ensure using deadlock avoidance mechanism on mutex acquiring
#### What happened:

Currently acquire more than one recursive mutex is ordered, that means acquiring same mutexes in other thread in different / reverse order cause a deadlock 

No deadlock on reverse order mutex acquiring

```
Time |      Thread A      |     Thread B
t1   | Attempt to lock L1 | Attempt to lock L2
t2   |    Gets lock L1    |    Gets lock L2
t3   |      In cs L1      |      In cs L2 
t4   | Attempt to lock L2 | Attempt to lock L1
t5   |     Block on L2    |     Block on L1
<waits forever: deadlock> | <waits forever: deadlock>
```

You can read more on ABBA deadlock

bvbfan:
/kind fix

It ensures atomic lock on multiple mutexes

Fixes #368 #458

### DIFF
--- a/src/test/sync_tests.cpp
+++ b/src/test/sync_tests.cpp
@@ -5,48 +5,49 @@
 #include <sync.h>
 #include <test/setup_common.h>
 
+#include <future>
+
 #include <boost/test/unit_test.hpp>
 
-namespace {
-template <typename MutexType>
-void TestPotentialDeadLockDetected(MutexType& mutex1, MutexType& mutex2)
+template<typename MutexType>
+void TryPotentialDeadlock(MutexType& mutex1, MutexType& mutex2)
 {
+    std::promise<void> threadFinished;
+    auto futureFinished = threadFinished.get_future();
     {
         LOCK2(mutex1, mutex2);
+        std::promise<void> threadStarted;
+        // start new thread
+        std::thread([&]() {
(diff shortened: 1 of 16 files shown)


src/test/sync_tests.cpp
```diff
@@ -5,48 +5,49 @@
...
+        TryPotentialDeadlock(rmutex1, rmutex2);
+    }
 
-    #ifdef DEBUG_LOCKORDER
-    g_debug_lockorder_abort = prev;
-    #endif
+    {
+        Mutex mutex1, mutex2;
+        TryPotentialDeadlock(mutex1, mutex2);
+    }
```
monstrobishi, line 50: Does this test reproduce the issue this fix is trying to solve

bvbfan: Yes, that's the issue,

uzyn:
Need reviewers

uzyn:
@bvbfan let's rebase this to master and not 1.7

bvbfan:
Let's keep open till 1.7.x is merged to master, then i will rebase it to master.

monstrobishi (approved):

PR #376, opened 2021-05-10
Added listaccounts RPC pagination test

### LINKED ISSUE #365: Add python test to simulate failing test on jellyfish
#### What happened:

This test on the jellyfish side is failing:
```
it('should listAccounts with pagination start and including_start', async () => {
      const accounts = await waitForListingAccounts()

      const pagination = {
        start: accounts[accounts.length - 1].key,
        including_start: true
      }

      const lastAccounts = await client.account.listAccounts(pagination)
      expect(lastAccounts.length).toBe(1)
    })
```

Please add a Python test in the ain repo to simulate this same test, so we can ensure correct behaviour at the Python test level.

The test should pass.

#### What are your environment parameters:

CICD docker container

surangap:

/kind chore

Fixes #365

#### Additional comments?:
Refer my Note in the code when giving the pagination parameters

### DIFF
--- a/test/functional/feature_listaccounts_pagination.py
+++ b/test/functional/feature_listaccounts_pagination.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2019 The Bitcoin Core developers
+# Copyright (c) DeFi Blockchain Developers
+# Distributed under the MIT software license, see the accompanying
+# file LICENSE or http://www.opensource.org/licenses/mit-license.php.
+"""Test listaccounts RPC pagination"""
+
+from test_framework.test_framework import DefiTestFramework
+from test_framework.util import assert_equal
+
+class AccountsValidatingTest(DefiTestFramework):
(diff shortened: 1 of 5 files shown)


test/functional/feature_listaccounts_pagination.py
```diff
@@ -0,0 +1,73 @@
...
+        node.utxostoaccount({account2: "10@0"})
+        node.generate(1)
+        self.sync_all()
+
+        accounts = node1.listaccounts()
+        #make sure we have two account in the system
+        assert_equal(len(accounts), 2)
+
+        pagination = {
+                "start": accounts[0]["key"], #NOTE(sp) : As per the current implementation, Has to give the first element in the array, not the last.
```
fuxingloh, line 62: @bvbfan has the sort order changed? This behaviour is weird.

For these test conditions, I would expect it to return 2 instead of one, since `including_start: True`.

```py
assert_equal(len(result2), 1)
```

PR #380, opened 2021-05-12
Fix chain split

Bushstar:
When a masternode creates a block the last block time is stored, this is used to calculate the coin-age and essentially resets the coin-age multiplier to zero. When the same masternode is staking on two nodes, a mempool difference causes the block hash to be different and the coin-age has been used to change the target difficulty a chain split will occur. The split nodes cannot accept each others blocks as the last entry in the masternode block time is there own which reset coin-age removing the target multiplier that was used to create the block, the result is that the block on the other chain is rejected as having a high hash. Invalidate the split block on one chain and it removes the last block time entry and allows it to accept the other chain.

This commit here uses the current height minus one when getting the last block time, we do not want to get the current or max block time in case of chain split.

### DIFF
--- a/src/masternodes/rpc_masternodes.cpp
+++ b/src/masternodes/rpc_masternodes.cpp
@@ -25,7 +25,12 @@ UniValue mnToJSON(uint256 const & nodeId, CMasternode const& node, bool verbose)
         obj.pushKV("banTx", node.banTx.GetHex());
         obj.pushKV("state", CMasternode::GetHumanReadableState(node.GetState()));
         obj.pushKV("mintedBlocks", (uint64_t) node.mintedBlocks);
-        obj.pushKV("targetMultiplier", pos::CalcCoinDayWeight(Params().GetConsensus(), node, GetTime()).getdouble());
+        int height;
+        {
+            LOCK(cs_main);
+            height = ChainActive().Height();
+        }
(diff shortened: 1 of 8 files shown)


bvbfan (dismissed):

monstrobishi (dismissed):

src/masternodes/rpc_masternodes.cpp
```diff
@@ -25,7 +25,10 @@ UniValue mnToJSON(uint256 const & nodeId, CMasternode const& node, bool verbose)
         obj.pushKV("banTx", node.banTx.GetHex());
         obj.pushKV("state", CMasternode::GetHumanReadableState(node.GetState()));
         obj.pushKV("mintedBlocks", (uint64_t) node.mintedBlocks);
-        obj.pushKV("targetMultiplier", pos::CalcCoinDayWeight(Params().GetConsensus(), node, GetTime()).getdouble());
+        {
+            LOCK(cs_main);
```
ShengguangXiao, line 29: This LOCK is for `ChainActive().Height()` ? Can just lock to get the height, no need lock for the whole calculation function.
```
int height;
{
    LOCK(cs_main);
    height = ChainActive().Height();
}
.... Use the height here.
```

Bushstar: Sure thing, now updated.

ShengguangXiao (approved):

PR #390, opened 2021-05-15
Change daily LP rewards automatically on subsidy reduction

Bushstar:
Change the daily LP rewards on the hard fork and each subsequent subsidy reduction. Prevent manual changing of daily LP rewards after the hard fork.

Also subtract from the burn community balance on chain rollback.

### DIFF
--- a/src/masternodes/govvariables/lp_daily_dfi_reward.cpp
+++ b/src/masternodes/govvariables/lp_daily_dfi_reward.cpp
@@ -18,8 +18,12 @@ UniValue LP_DAILY_DFI_REWARD::Export() const {
     return ValueFromAmount(dailyReward);
 }
 
-Res LP_DAILY_DFI_REWARD::Validate(const CCustomCSView &) const
+Res LP_DAILY_DFI_REWARD::Validate(const CCustomCSView & view) const
 {
+    if (view.GetLastHeight() >= static_cast<uint32_t>(Params().GetConsensus().EunosHeight)) {
+        return Res::Err("Cannot be set manually after Eunos hard fork");
+    }
+
     // nothing to do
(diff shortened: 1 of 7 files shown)


monstrobishi (dismissed):

src/masternodes/govvariables/lp_daily_dfi_reward.cpp
```diff
@@ -18,8 +18,12 @@ UniValue LP_DAILY_DFI_REWARD::Export() const {
     return ValueFromAmount(dailyReward);
 }
 
-Res LP_DAILY_DFI_REWARD::Validate(const CCustomCSView &) const
+Res LP_DAILY_DFI_REWARD::Validate(const CCustomCSView & view) const
 {
+    if (view.GetLastHeight() >= static_cast<uint32_t>(Params().GetConsensus().EunosHeight)) {
```
bvbfan, line 23: We set last height just before `ConnectBlock` ends, so in Eunos height `GetLastHeight` will point to previous one, do you think? It can be used by -1 at end.

uzyn (dismissed):
Great tests. Looks good.

src/masternodes/govvariables/lp_daily_dfi_reward.cpp
Bushstar: An off by one in this circumstance does not make a difference, we do not expect to set LP_DAILY_DFI_REWARD manually and especially not on the hard fork block. The automatic changing of the LP_DAILY_DFI_REWARD comes after parsing of the setgov custom TX.

bvbfan (approved):

monstrobishi (approved):

PR #400, opened 2021-05-18
Skip old coinbase block post Eunos

bvbfan:
/kind fix

Post Eunos coinbase spent less than pre-fork, so situation is old masternode can cause a bad block which will case an updated node to revert blocks till last checkpoint / anchor height. It can happen at any time post Eunos.


src/validation.cpp
```diff
@@ -2730,9 +2724,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                     if (!cache.HasICXSubmitEXTHTLCOpen(dfchtlc->offerTx))
                     {
                         CTokenAmount makerDeposit{DCT_ID{0}, offer->takerFee};
-                        auto res = cache.AddBalance(order->ownerAddress, makerDeposit);
-                        if (!res)
-                            LogPrintf("Can't refund makerDeposit back to owner: %s\n", res.msg);
+                        cache.AddBalance(order->ownerAddress, makerDeposit);
```
ShengguangXiao, line 2727: Why skip the result check?

src/validation.cpp
```diff
@@ -3541,6 +3523,16 @@ bool CChainState::ActivateBestChainStep(CValidationState& state, const CChainPar
                         // but it can be produced by outdated/malicious masternode
                         // so we should not shoutdown entire network, let's skip it
                         continue;
+                    } else if (reason == ValidationInvalidReason::CONSENSUS) {
+                        if (pindexConnect->nHeight >= chainparams.GetConsensus().EunosHeight) {
+                            auto strReason = state.GetRejectReason();
+                            // we have situaltion when old masternode will generate a block
```
ShengguangXiao, line 3529: change situaltion -> situation

src/validation.cpp
bvbfan: Since we can subtract we can add, it's safe. If add fails that's mean corrupt storage.

src/validation.cpp
```diff
@@ -3523,6 +3523,16 @@ bool CChainState::ActivateBestChainStep(CValidationState& state, const CChainPar
                         // but it can be produced by outdated/malicious masternode
                         // so we should not shoutdown entire network, let's skip it
                         continue;
+                    } else if (reason == ValidationInvalidReason::CONSENSUS) {
+                        if (pindexConnect->nHeight >= chainparams.GetConsensus().EunosHeight) {
```
monstrobishi, line 3527: Change this to `==`

src/validation.cpp
Mixa84: I think we shouldn't do close order, offer, dfchtlc, exthtlc if Sub/AddBalance failed!

bvbfan: We should since if we not do it, order / offer and so on we have them in inconsistent state, they are not closed.

PR #408, opened 2021-05-20
Fix/prevent reverts

bvbfan:
/kind fix

Including merkle root of account system we don't need to revert chain to last checkpoint / anchor since we validate ongoing changes.


src/validation.cpp
```diff
@@ -3516,28 +3516,21 @@ bool CChainState::ActivateBestChainStep(CValidationState& state, const CChainPar
...
-                            // that block is invalid, let's keep waiting for the choosen one
-                            if (strReason.find("bad-cb-amount") != strReason.npos) {
-                                continue;
-                            }
-                        }
+                        // so we should not shoutdown entire network
                     }
-                    InvalidChainFound(vpindexToConnect.front());
-                    if (fCheckpointsEnabled && pindexConnect == pindexMostWork) {
+                    if (pindexConnect->nHeight < chainparams.GetConsensus().EunosHeight
```
monstrobishi, line 3532: Does this need to check height? I think we can still encounter problem before EunosHeight

monstrobishi (changes requested):
We should keep reorg behviour unless we include account system merkle root validation

Bushstar (approved):

monstrobishi (approved):

Issue #423, opened 2021-05-24
Setting of Eunos block and node release


uzyn:
#### What would you like to be added:

With the successful testing of Eunos testnet, let's wrap up for mainnet release.

Proposed Eunos mainnet block: 894,000. ETA Thu, June 3, 2021, 07:00 UTC

#### Why is this needed:

To pull off a mainnet upgrade.

PR #424, opened 2021-05-24
SPV add call, fix Windows connection issue and set Eunos height.

Bushstar:
* Adds spv_getpeers call to see connected peers on SPV.
* Set Windows to use non-blocking connections.
* Set Eunos height https://github.com/DeFiCh/ain/issues/423


uzyn (dismissed):
Thanks @Bushstar 
Do try to split PRs if possible for unrelated commits. Approving this one as we are about to lift off.

Bushstar:
@uzyn I would normally, put it in here to avoid waiting on two PRs to complete tests sequentially when we want to create a release somewhat ASAP.

monstrobishi (approved):

PR #428, opened 2021-05-25
defer chain lock to caller to allow atomic lock

### LINKED ISSUE #368: Ensure using deadlock avoidance mechanism on mutex acquiring

fuxingloh:
> Signed-off-by: Anthony Fieroni <bvbfan@abv.bg>
> Rebase for master

/kind fix

It ensures atomic lock on multiple mutexes

Fixes #368


bvbfan:
Close in favor of #371

PR #436, opened 2021-05-26
Fix to always return array for "amounts" in listaccounthistory

Mixa84:

/kind fix

Changed that all every result from listaccounthistory returns 'amounts' field in array format to unify it.
Adapted test accordingly to check the type for 'amounts' field.


uzyn (approved):
Thanks

ShengguangXiao (approved):

PR #438, opened 2021-05-27
Add new checkpoint to minimize undo data

bvbfan:
/kind chore

Checkpoint prevents undo data to grow too much


monstrobishi:
can we also add 875,000, save more space

ShengguangXiao (approved):

PR #448, opened 2021-05-28
Remove minter cache

bvbfan:
/kind fix

Cache can be used only over levedb records.


bvbfan:
Close in favour of #443

PR #462, opened 2021-05-31
remove db.Compact for performance testing

fuxingloh:

/kind fix

Fixes #


bvbfan:
Compact should stay, we can introduce a parameter dbs to be placed in memory.

PR #467, opened 2021-06-01
Faster calculation of anchoring teams

Bushstar:
Calculating anchoring teams takes about 2,000 mn every 120 blocks during sync, the update here reduces that time to about 130 ms every 120 blocks.

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -639,17 +639,31 @@ enum AnchorTeams {
 void CCustomCSView::CalcAnchoringTeams(const uint256 & stakeModifier, const CBlockIndex *pindexNew)
 {
     std::set<uint256> masternodeIDs;
-    int blockSample{7 * 2880}; // One week
-    const CBlockIndex* pindex = pindexNew;
+    const int blockSample = 7 * Params().GetConsensus().blocksPerDay(); // One week
 
     // Get active MNs from last week's worth of blocks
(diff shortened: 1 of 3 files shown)


src/masternodes/masternodes.cpp
```diff
@@ -639,17 +639,31 @@ enum AnchorTeams {
...
+            }
+
+            return true;
+        }, MNBlockTimeKey{});
+    } else {
+        const CBlockIndex* pindex = pindexNew;
+        for (int i{0}; pindex && i < blockSample; pindex = pindex->pprev, ++i) {
+            CKeyID minter;
+            if (pindex->GetBlockHeader().ExtractMinterKey(minter)) {
+                LOCK(cs_main);
```
bvbfan, line 662: Mutex out of the loop, it's slow to lock/unlock too often

src/masternodes/masternodes.cpp
```diff
@@ -639,17 +639,31 @@ enum AnchorTeams {
 void CCustomCSView::CalcAnchoringTeams(const uint256 & stakeModifier, const CBlockIndex *pindexNew)
 {
     std::set<uint256> masternodeIDs;
-    int blockSample{7 * 2880}; // One week
-    const CBlockIndex* pindex = pindexNew;
+    const int blockSample{7 * 2880}; // One week
```
bvbfan, line 642: We can use consensus blocksPerDay

bvbfan (approved):

bvbfan (approved):

PR #468, opened 2021-06-01
Set daily reward as daily and not per block

Bushstar:


src/validation.cpp
```diff
@@ -2616,7 +2616,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                     // Cast to avoid UniValue in GovVariable Export/Import
                     auto lpVar = dynamic_cast<LP_DAILY_DFI_REWARD*>(var.get());
                     if (lpVar && lpVar->dailyReward != subsidy) {
-                        lpVar->dailyReward = subsidy;
+                        lpVar->dailyReward = subsidy * ((60 * 60 * 24) / chainparams.GetConsensus().pos.nTargetSpacing);
```
bvbfan, line 2619: We have `BlocksPerDay` in Consensus.
https://github.com/DeFiCh/ain/blob/master/src/consensus/params.h#L131-L134

test/functional/feature_poolswap_mechanism.py
```diff
@@ -319,7 +319,7 @@ def run_test(self):
         assert(self.nodes[0].getblockcount() == 120) # eunos
 
         self.LP_DAILY_DFI_REWARD = self.nodes[0].getgov("LP_DAILY_DFI_REWARD")['LP_DAILY_DFI_REWARD']
-        assert_equal(self.LP_DAILY_DFI_REWARD, Decimal('103.08268000'))
+        assert_equal(self.LP_DAILY_DFI_REWARD, Decimal('14843.90592000')) # 144 blocks a day times 103.08268000
```
uzyn, line 322: Why is it 144? Functional test has a much longer block time?

Bushstar: Regtest is still on Bitcoin block times.

https://github.com/DeFiCh/ain/blob/118b8be697cc880de2a23b5a47b6031d3b49ef2b/src/chainparams.cpp#L697

src/validation.cpp
Bushstar: Updated to use BlocksPerDay.

uzyn (approved):

PR #473, opened 2021-06-02
Loan setcollateraltoken

### LINKED ISSUE #433: Loan: Initialization & setcollateraltoken
See https://github.com/DeFiCh/pinkpaper/pull/3

1. `setcollateraltoken TOKEN FACTOR PRICE_FEED_ID [ACTIVATE_AFTER_BLOCK]`
  - `TOKEN`: Token must not be the same decentralized token is issued by the same operator's loan program.
  - `FACTOR`: A number between `0` to `1`, inclusive. `1` being 100% collaterization factor.
  - `PRICE_FEED_ID`: Price feed from [oracle](../oracle) that is being used to price the token.
  - `ACTIVATE_AFTER_BLOCK` _(optional)_: If set, this will only be activated after the set block. The purpose is to allow good operators to provide sufficient warning to their users should certain collaterization factors need to be updated.
  - This very same transaction type is also used for updating and removing of collateral token, by setting the factor to `0`.

Mixa84:

/kind feature

Initial code for base of loan mechanism. It has one tx that sets collateral token for creating new tokens based on oracle price feeds.

Fixes #433

### DIFF
--- a/src/masternodes/loan.cpp
+++ b/src/masternodes/loan.cpp
@@ -1,0 +1,35 @@
+#include <masternodes/loan.h>
+#include <rpc/util.h> /// AmountFromValue
+#include <core_io.h> /// ValueFromAmount
+
+const unsigned char CLoanView::LoanSetCollateralTokenCreationTx           ::prefix = 0x10;
+const unsigned char CLoanView::LoanSetCollateralTokenKey                  ::prefix = 0x11;
+
+std::unique_ptr<CLoanView::CLoanSetCollateralTokenImpl> CLoanView::GetLoanSetCollateralToken(uint256 const & txid) const
+{
+    auto collToken = ReadBy<LoanSetCollateralTokenCreationTx,CLoanSetCollateralTokenImpl>(txid);
+    if (collToken)
+        return MakeUnique<CLoanSetCollateralTokenImpl>(*collToken);
+    return {};
+}
+
+Res CLoanView::LoanCreateSetCollateralToken(CLoanSetCollateralTokenImpl const & collToken)
+{
+    //this should not happen, but for sure
+    if (GetLoanSetCollateralToken(collToken.creationTx))
+        return Res::Err("setCollateralToken with creation tx %s already exists!", collToken.creationTx.GetHex());
+    if (collToken.factor > COIN)
+        return Res::Err("setCollateralToken factor must be lower or equal than 1!");
+    if (collToken.factor < 0)
+        return Res::Err("setCollateralToken factor must not be negative!");
+
+    WriteBy<LoanSetCollateralTokenCreationTx>(collToken.creationTx, collToken);
+    WriteBy<LoanSetCollateralTokenKey>(collToken.idToken, collToken.creationTx);
+
+    return Res::Ok();
+}
+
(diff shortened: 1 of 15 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -1616,6 +1629,35 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
         return !res ? res : mnview.ICXCloseMakeOfferTx(*offer, CICXMakeOffer::STATUS_CLOSED);
     }
 
+    Res operator()(const CLoanSetCollateralTokenMessage& obj) const {
+        auto res = CheckICXTx();
```
Bushstar, line 1633: It is possible that there is no change output, a dust amount of change gets added to the fee. Looking at CheckICXTx I think it in general should check that there are no more than 2 outputs and not 2 exactly, however we cannot change that behaviour without putting it behind a fork, post-fork it could then be simplified.

Mixa84: Yes, I mentioned earlier that we shouldn't check for exact number of vouts as it could be no change. But this can also be fixed (I used this approach before) with creating a change vout with 0 amount. This way we don't have to wait for new hard fork, and we also have this rigid check that someone cannot make a different tx than we expect...

Bushstar: Change output below the dust threshold will be a non-standard TX and should be rejected on mainnet.

https://github.com/DeFiCh/ain/blob/20bd608730116a5683fde678cb1a5380c05681c6/src/policy/policy.cpp#L126

src/masternodes/loan.h
```diff
@@ -0,0 +1,73 @@
+#ifndef DEFI_MASTERNODES_LOAN_H
+#define DEFI_MASTERNODES_LOAN_H
+
+#include <tuple>
```
Bushstar, line 4: I do not think that tuple, pubkey.h or script.h are used.

src/masternodes/loan.cpp
```diff
@@ -0,0 +1,46 @@
+#include <masternodes/loan.h>
+#include <rpc/util.h> /// AmountFromValue
+#include <core_io.h> /// ValueFromAmount
```
Bushstar, line 3: These do not seem to be used.

src/masternodes/loan.h
Mixa84: Copy/paste from ICX.

src/masternodes/mn_checks.cpp
Mixa84: Ok then. Yeah, in some previous implementation, my txes were special type so it wasn't considered as dust. We'll then check for 2 or less vouts on next fork.

src/masternodes/mn_checks.cpp
```diff
@@ -377,6 +386,10 @@ class CCustomMetadataParseVisitor : public boost::static_visitor<Res>
         auto res = isPostEunosFork();
         return !res ? res : serialize(obj);
     }
+    Res operator()(CLoanSetCollateralTokenMessage& obj) const {
+        auto res = isPostFortCanningFork();
+        return serialize(obj);
```
Bushstar, line 391: Will not fail under fork height, should be.

`return !res ? res : serialize(obj);`

Bushstar (approved):

PR #480, opened 2021-06-03
Include anchor tx in merkle root

bvbfan:
/kind fix

Fixes bad merkle root cause by not include anchor tx in merkle root calculation


bvbfan:
That's not correct.

PR #482, opened 2021-06-03
Hardfork to disable account merkle root check

bvbfan:
/kind fix

Disable account merkle root temporary we should investigate on anchor reward tx

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -320,7 +320,8 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
     int64_t nTime2 = GetTimeMicros();
 
     pblock->hashMerkleRoot = BlockMerkleRoot(*pblock);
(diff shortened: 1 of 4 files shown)


ShengguangXiao (dismissed):

ShengguangXiao (dismissed):

Bushstar (approved):

uzyn:
Block 895741 checkpoint  61bc1d73c720990dde43a3fec1f703a222ec5c265e6d491efd60eeec1bdb6dc3

Bushstar (approved):

src/miner.cpp
```diff
@@ -320,7 +320,8 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
     int64_t nTime2 = GetTimeMicros();
 
     pblock->hashMerkleRoot = BlockMerkleRoot(*pblock);
```
ShengguangXiao, line 322: Before 1.7.x don't have this line, should we keep it?

PR #483, opened 2021-06-03
Revert invalid chain

bvbfan:
/kind fix

Invalidate chain with merkle root mismatch


uzyn:
Tested and this invalidation works.

PR #487, opened 2021-06-04
Change staking to single threaded and precompute merkle root

monstrobishi:

Added comments for implementation of single threaded and precomputation of merkle root

/kind feature

Fixes #


src/miner.cpp
```diff
@@ -823,93 +826,118 @@ namespace pos {
...
-                LogPrintf("ThreadStaker (%s): waiting init...\n", operatorName);
+
+        int lastHeight = 0;
+        std::shared_ptr<CBlock> pblock;
+
+        for (auto it = args.begin(); it != args.end(); ) {
+            const auto& arg = *it;
+            const auto operatorName = arg.operatorID.GetHex();
+
+            pos::Staker staker;
```
ShengguangXiao, line 880: This definition can move outside of `for (auto it = args.begin(); it != args.end(); )`? Then can reuse the object.

src/miner.cpp
```diff
@@ -832,93 +838,100 @@ namespace pos {
...
+
+            if ((arg.nMaxTries != -1 && tried >= arg.nMaxTries)
+            || (arg.nMint != -1 && nMinted[arg.operatorID] >= arg.nMint)) {
+                it = args.erase(it);
+                continue;
+            }
 
-        nTried++;
-        if (trying() && notDone()) {
             std::this_thread::sleep_for(std::chrono::milliseconds(900));
```
monstrobishi, line 930: @bvbfan I think sleep here is sleeping between every MN, we do not want to sleep here every 900 ms! This needs to be moved out of MN loop

Bushstar (approved):

bvbfan (approved):

PR #498, opened 2021-06-07
Fix accounts performance issue

### LINKED ISSUE #470: Command listaccounthistory does not work like expected with the reserved word mine
#### What happened:

The node works fine for me except the command "listaccounthistory **mine**"
From the block ~726k no entries. (I also checked the address -> is_mine = true)

It looks like the reserved word mine is no longer working, because if i try it directly with an address where tokens of mine are located then it will work:

2021-06-02_11h50_36

If I use "all" then also from the block ~800k no entries.

I am now fully synchronized and with listaccounthistory mine used to get the last 100 history entries by default.

image

120434293-90f0e700-c37c-11eb-89e2-b98bf52276f3

I can reindex again and the same will happen. (tried 5 times with node 1.7.1-1.7.4)

#### What are your environment parameters:

Windows10, 11th Gen Intel(R) Core(TM) i9-11900F @ 2.50GHz   2.50 GHz, 32GB RAM

@bvbfan 
Here is my used defi conf:

rpcauth=xxx
rpcuser=xxx
rpcpassword=xxx
rpcconnect=127.0.0.1
testnet=0
regtest=0
datadir=path

[main]
wallet=wallet.dat
walletdir=path
rpcbind=127.0.0.1
rpcport=8555
checkpoints=0
spv=0
gen=0

[test]
rpcbind=127.0.0.1
rpcport=8555

bvbfan:
/kind fix

- It's address not only listaccounthistory but other commands that want to make filter by addresses owned by the wallet.

- Account history order is fixed in `listaccounthistory mine` as well.

- It prevents showing 0.00000000 commissions history

Fixes #470

### DIFF
--- a/src/bloom.cpp
+++ b/src/bloom.cpp
@@ -42,7 +42,7 @@ CBloomFilter::CBloomFilter(const unsigned int nElements, const double nFPRate, c
 inline unsigned int CBloomFilter::Hash(unsigned int nHashNum, const std::vector<unsigned char>& vDataToHash) const
 {
     // 0xFBA4C795 chosen as it guarantees a reasonable bit difference between nHashNum values.
-    return MurmurHash3(nHashNum * 0xFBA4C795 + nTweak, vDataToHash) % (vData.size() * 8);
+    return MurmurHash3(nHashNum * 0xFBA4C795 + nTweak, vDataToHash.data(), vDataToHash.size()) % (vData.size() * 8);
 }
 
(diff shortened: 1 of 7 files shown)


monstrobishi (approved):

salomonkarsten:
With wallet 2.5.0 no change.
Even after full sync with no snapshot. 
history is only displayed with an old address

PR #499, opened 2021-06-07
Limit max tip age to reasonable value

bvbfan:
/kind fix

Max tip age is used in IBD it allows miners to start their work too early which can result in high hash error.
It's a safe value since leaving IBD is happen one time per run we cannot end with stalling masternode.
Fee estimation tip age is in direct connection to max tip age and should be lowered as well.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -472,6 +472,7 @@ void SetupServerArgs()
     gArgs.AddArg("-dakotaheight", "Dakota fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
     gArgs.AddArg("-dakotacrescentheight", "DakotaCrescent fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
     gArgs.AddArg("-eunosheight", "Eunos fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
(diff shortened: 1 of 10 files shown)


src/validation.h
```diff
@@ -609,6 +601,14 @@ class CChainState {
...
     CChainState();
 
+    /**
+    * Whether this chainstate is undergoing initial block download.
+    *
+    * Mutable because we need to be able to mark IsInitialBlockDownload()
+    * const, which latches this for caching purposes.
+    */
+    mutable std::atomic<bool> m_cached_finished_ibd{false};
+
```
ShengguangXiao, line 611: Why don't keep the variable private and add a function to set it?

Bushstar: Probably sensible but this was done in a rush yesterday and this PR is to provide a temp solution to keep help stakers.

src/init.cpp
```diff
@@ -1203,6 +1203,7 @@ bool AppInitParameterInteraction()
         return InitError("unknown rpcserialversion requested.");
 
     nMaxTipAge = gArgs.GetArg("-maxtipage", DEFAULT_MAX_TIP_AGE);
+    nMaxTipAge = std::min(nMaxTipAge, DEFAULT_MAX_TIP_AGE);
```
ShengguangXiao, line 1206: Do we still limit the value to 10 minutes?

bvbfan: We can remove it.

Bushstar (approved):

PR #505, opened 2021-06-07
Solve pos puzzle before create and propagate a block

bvbfan:
/kind fix

It fixes `createmasternode` issue in account merkle root.

### DIFF
--- a/src/masternodes/rpc_masternodes.cpp
+++ b/src/masternodes/rpc_masternodes.cpp
@@ -39,7 +39,14 @@ UniValue mnToJSON(uint256 const & nodeId, CMasternode const& node, bool verbose,
 
         // Only get targetMultiplier for active masternodes
         if (node.IsActive()) {
-            obj.pushKV("targetMultiplier", pos::CalcCoinDayWeight(Params().GetConsensus(), node, GetTime(), ChainActive().Height()).getdouble());
+            auto stakerBlockTime = pcustomcsview->GetMasternodeLastBlockTime(node.operatorAuthAddress, ChainActive().Height());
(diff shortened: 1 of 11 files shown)


src/pos.cpp
```diff
@@ -136,7 +146,7 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
             // Special difficulty rule for testnet:
             // If the new block's timestamp is more than 2* 10 minutes
```
ShengguangXiao, line 147: Can also update the comments? Now should be 2 * 30 seconds.

src/masternodes/rpc_masternodes.cpp
```diff
@@ -39,7 +39,14 @@ UniValue mnToJSON(uint256 const & nodeId, CMasternode const& node, bool verbose,
 
         // Only get targetMultiplier for active masternodes
         if (node.IsActive()) {
-            obj.pushKV("targetMultiplier", pos::CalcCoinDayWeight(Params().GetConsensus(), node, GetTime(), ChainActive().Height()).getdouble());
+            auto stakerBlockTime = pcustomcsview->GetMasternodeLastBlockTime(node.operatorAuthAddress, node.creationHeight);
```
ShengguangXiao, line 42: The GetMasternodeLastBlockTime using createHeight should also change to the current block height?

bvbfan: Yes. That was the bug in #505 as well, thanks.

monstrobishi (dismissed):

Bushstar (approved):

PR #511, opened 2021-06-08
Keep mempool accounts view cache

bvbfan:
/kind feature

Mempool menages a view of account changes to its txs. It will prevent trigger of rewards recalculation on every insert.


src/txmempool.cpp
```diff
@@ -550,6 +550,19 @@ void CTxMemPool::removeConflicts(const CTransaction &tx)
     }
 }
 
+CTxMemPool::~CTxMemPool()
+{
+}
```
ShengguangXiao, line 555: Why need to add an empty destructor?

src/txmempool.cpp
```diff
@@ -957,9 +972,21 @@ size_t CTxMemPool::DynamicMemoryUsage() const {
...
     AssertLockHeld(cs);
     UpdateForRemoveFromMempool(stage, updateDescendants);
+    std::set<uint256> txids;
     for (txiter it : stage) {
+        txids.insert(it->GetTx().GetHash());
         removeUnchecked(it, reason);
     }
+    if (pcustomcsview && !txids.empty()) {
+        CCustomCSView view(accountsView());
+        accountsView().ForEachUndo([&](UndoKey const & key, CLazySerialize<CUndo>) {
```
ShengguangXiao, line 982: Why here don't just use `view.ForEachUndo`?

bvbfan: It could i like that when you iterate one view to not change it meanwhile, that's why another would help.

src/txmempool.cpp
bvbfan: `CCustomCSView` is forward declared in `txmempool.h` which is incomplete type, so if we did not provide destructor in the cpp we should `include` header with complete type. I prefer to not include but follow forward declaration.

Bushstar (approved):

PR #514, opened 2021-06-09
Increase time range for staking

Bushstar:
debug.log sample with extra logging to show when and what nLastCoinStakeSearchTime and lastBlockSeen get set to, staking range shows the range of times to try.

https://gist.github.com/Bushstar/6abd686cc71c318cfd711f331452993a


bvbfan (dismissed):

src/miner.cpp
```diff
@@ -829,15 +841,13 @@ namespace pos {
     }
 
     template <typename F>
-    bool Staker::withSearchInterval(F&& f) {
-        const int64_t nTime = GetAdjustedTime(); // TODO: SS GetAdjustedTime() + period minting block
+    void Staker::withSearchInterval(F&& f) {
+        // Mine up to max future minus 10 second buffer
+        nFutureTime = GetAdjustedTime() + (MAX_FUTURE_BLOCK_TIME_DAKOTACRESCENT - 10);
```
uzyn, line 846:
```suggestion
        nFutureTime = GetAdjustedTime() + (MAX_FUTURE_BLOCK_TIME_DAKOTACRESCENT - 5);
```

uzyn: Let's do 5 secs. I think it should be safe enough while not giving anyone any significant extra edge.

uzyn (approved):
Looks good.

src/miner.cpp
```diff
@@ -728,6 +732,14 @@ namespace pos {
 
         auto pblock = std::make_shared<CBlock>(pblocktemplate->block);
 
+        // Set search time if null or last block has changed
+        if (!nLastCoinStakeSearchTime || lastBlockSeen != tip->GetBlockHash()) {
+            // Expect this to be called sequentially but let's lock to be safe.
+            CLockFreeGuard lock(Staker::cs_staker);
```
ShengguangXiao, line 738: Now is single thread mining, do we still need lock?

Bushstar: No, but I'm paranoid that someone will make it "concurrent" again. I will remove.

src/miner.cpp
```diff
@@ -829,15 +842,13 @@ namespace pos {
     }
 
     template <typename F>
-    bool Staker::withSearchInterval(F&& f) {
-        const int64_t nTime = GetAdjustedTime(); // TODO: SS GetAdjustedTime() + period minting block
+    void Staker::withSearchInterval(F&& f) {
+        // Mine up to max future minus 10 second buffer
```
uzyn, line 846:
```suggestion
        // Mine up to max future minus 5 second buffer
```

uzyn (approved):
Comment fix

src/miner.cpp
```diff
@@ -728,6 +731,16 @@ namespace pos {
 
         auto pblock = std::make_shared<CBlock>(pblocktemplate->block);
 
+        // Set search time if null or last block has changed
+        if (!nLastCoinStakeSearchTime || lastBlockSeen != tip->GetBlockHash()) {
+            if (Params().NetworkIDString() == CBaseChainParams::REGTEST) {
+                nLastCoinStakeSearchTime = GetAdjustedTime();
+            } else {
+                nLastCoinStakeSearchTime = tip->GetMedianTimePast() + 1;
```
ShengguangXiao, line 739: Plus 1 is for what purpose?

Bushstar: Median time or less will fail with time-too-old, I'll add a comment.

monstrobishi (dismissed):

ShengguangXiao:
If the lastSearchTime < currentTime < futureTime, then the time range between lastSearchTime and currentTime is searched for two times, is this designated?

Bushstar:
@ShengguangXiao thanks for spotting this, will update. This works on subsequent loops but the first is a duplication of effort.

ShengguangXiao (approved):

ShengguangXiao (approved):

PR #517, opened 2021-06-10
Do not keep failed tx coins in miner cache

bvbfan:
/kind fix

Failed tx does not change accounts view but affect coins cache, it can result in coins override

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -485,7 +485,7 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
     std::set<uint256> checkedTX;
 
     // Copy of the view
-    CCoinsViewCache coins(&::ChainstateActive().CoinsTip());
+    CCoinsViewCache coinsView(&::ChainstateActive().CoinsTip());
 
     while (mi != mempool.mapTx.get<ancestor_score>().end() || !mapModifiedTx.empty())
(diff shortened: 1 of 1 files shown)


src/miner.cpp
```diff
@@ -618,6 +622,7 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
                 // Track checked TXs to avoid double applying
                 checkedTX.insert(tx.GetHash());
             }
+            coins.Flush();
```
ShengguangXiao, line 625: If breaked at line 619, then the `coins.Flush();` will not be called.

bvbfan: Yes, that's idea of the patch. Right now we keep in cache coin from failing tx.
```
2021-06-06T14:19:15Z addPackageTxs: Failed AddPoolLiquidity TX afbc9447762929e132632ac9a063fd4d0189f69092ea85a83bdced896272b8d9: AddPoolLiquidityTx: amount 0.00332584 is less than 0.00336510
2021-06-06T14:19:15Z 

************************
EXCEPTION: St11logic_error       
Adding new coin that replaces non-pruned entry       
```
You can see, tx is failing, but we keep coin in cache, then in next iteration we hit a coin override and throw.

Bushstar (approved):

PR #527, opened 2021-06-14
[Needs tests] Fix/limit blocktime range

bvbfan:
/kind fix

Limit the range of block time to -+ 60 sec to local node time


muirglacier:
Do not merge this pull request, it will break the entire blockchain.

In CheckContextualBlockHeader() you just check whether the blocktime is NOW()-60 or lower, and when it is then you reject the block. What if a node is syncing from scratch - i.e. synching old blocks to catch up to the current height? All those blocks would be falsely INVALID.

bvbfan:
Functional tests does not follow new block time range.

uzyn:
@hidiego Please update the tests.

uzyn:
Thanks @bvbfan Closing this for https://github.com/DeFiCh/ain/issues/575

PR #542, opened 2021-06-16
Keep minter pubkey in cache

bvbfan:
/kind fix

It will speed up sync since extracting minter key will be faster

### DIFF
--- a/src/primitives/block.h
+++ b/src/primitives/block.h
@@ -69,6 +69,7 @@ public:
         height = 0;
         mintedBlocks = 0;
         sig = {};
+        recoveredPubKey = CPubKey{};
     }
 
     bool IsNull() const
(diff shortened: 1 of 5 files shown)


Bushstar (approved):

src/primitives/block.h
```diff
@@ -84,9 +88,10 @@ class CBlockHeader
 
     bool ExtractMinterKey(CKeyID &key) const
     {
-        CPubKey recoveredPubKey{};
-        if (!recoveredPubKey.RecoverCompact(GetHashToSign(), sig)) {
-            return false;
+        if (!recoveredPubKey.IsValid()) {
+            if (!recoveredPubKey.RecoverCompact(GetHashToSign(), sig)) {
+                return false;
+            }
```
ShengguangXiao, line 94: This function may be not thread safe.

bvbfan: It does not need to be. It's used under `cs_main` lock.

Bushstar (approved):

Bushstar (approved):

PR #543, opened 2021-06-16
Do db compacting after undo pruning

bvbfan:
/kind fix

It calls db compacting only when huge amount of undo data is pruned


monstrobishi (approved):

fuxingloh:
I will test this build, thanks!

Bushstar (approved):

PR #556, opened 2021-06-21
On HTLC creation get last weeks filtered blocks

Bushstar:
When creation a HTLC address request from peers filtered blocks for the last week's worth of blocks on the Bitcoin network, this will then include any TXs that belong to our HTLC address if they occurred in the last week.

### DIFF
--- a/src/spv/bitcoin/BRPeerManager.cpp
+++ b/src/spv/bitcoin/BRPeerManager.cpp
@@ -592,6 +592,36 @@ static void _mempoolDone(void *info, int success)
     else peer_log(peer, "mempool request failed");
 }
 
+static void _rescanBlockchain(void *info, int success)
+{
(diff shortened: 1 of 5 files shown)


src/spv/bitcoin/BRPeerManager.cpp
```diff
@@ -592,6 +592,36 @@ static void _mempoolDone(void *info, int success)
...
+
+    free(info);
+
+    if (success) {
+        manager->lock.lock();
+
+        BRMerkleBlock *block = manager->lastBlock;
+
+        // Get block from one week ago
+        int32_t bitcoinWeek{144 * 7};
```
bvbfan, line 608: Use `constexpr` or it'll be VLA.

Bushstar: Thanks @bvbfan

bvbfan (approved):

PR #557, opened 2021-06-22
Apply token filter to rewards in history

bvbfan:
/kind fix

It filters rewards by token in account history


monstrobishi (changes requested):
Ensure no external services rely on existing behavior before merging

uzyn:
@bvbfan can you provide some writeups on what RPCs does this break and sample old and new outputs please?

bvbfan:
It does not break anything, it can return less results, it means when you set token filter BTC will not get rewads in DFI

Bushstar (approved):

PR #576, opened 2021-07-01
On-Chain-Governance

### LINKED ISSUE #344: On-Chain Governance

Refer to the Pinkpaper for implementation details: https://github.com/DeFiCh/pinkpaper/tree/main/governance

#### Why is this needed:

An on-chain governance system allows us to approve DFIPs (DeFiChain Improvement Proposals) in a decentralised manner.

bvbfan:
/kind feature

Fixes #344


Bushstar (approved):

src/masternodes/rpc_proposals.cpp
```diff
@@ -0,0 +1,622 @@
...
+               "\nReturns real time information about proposal state.\n",
+               {
+                        {"proposalId", RPCArg::Type::STR, RPCArg::Optional::NO, "The proposal id)"},
+               },
+               RPCResult{
+                       "{id:{...},...}     (obj) Json object with proposal vote information\n"
+               },
+               RPCExamples{
+                       HelpExampleCli("listvotes", "txid")
+                       + HelpExampleRpc("listvotes", "txid")
```
ShengguangXiao, line 443: Example should be `getproposal`

src/masternodes/rpc_proposals.cpp
```diff
@@ -0,0 +1,622 @@
...
+               },
+               RPCExamples{
+                       HelpExampleCli("createvoc", "'The voc title' '[{\"txid\":\"id\",\"vout\":0}]'")
+                       + HelpExampleRpc("createvoc", "'The voc title' '[{\"txid\":\"id\",\"vout\":0}]'")
+               },
+    }.Check(request);
+
+    if (pwallet->chain().isInitialBlockDownload()) {
+        throw JSONRPCError(RPC_CLIENT_IN_INITIAL_DOWNLOAD,
+                           "Cannot create a cfp while still in Initial Block Download");
```
Jouzo, line 198: should be "Cannot create a voc while still in Initial Block Download"

src/masternodes/mn_checks.cpp
```diff
@@ -1616,6 +1647,68 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        }
+        auto ownerDest = node->ownerType == 1 ? CTxDestination(PKHash(node->ownerAuthAddress))
+                                              : CTxDestination(WitnessV0KeyHash(node->ownerAuthAddress));
+        if (!HasAuth(GetScriptForDestination(ownerDest))) {
+            return Res::Err("tx must have at least one input from the owner");
+        }
+        if (!node->IsActive(height)) {
+            return Res::Err("masternode <%s> is not active", obj.masternodeId.GetHex());
+        }
+        if (node->mintedBlocks < 1) {
```
surangap, line 1698: From the pink paper
`The masternode has successfully minted at least 1 block during vote submission AND at vote finalizing block.`
not sure if we have the data to check the duration part of the above statement. pink paper should be updated??

Jouzo: From what I can see on `ain` side, the check is only this one. It is not following the pink paper implementation.
I will raise the issue on the governance PR.

surangap: hmm, looks like I read it wrong. there's an`at`. so it's two conditions at vote submission and at vote finalizing.

Jouzo: A check at vote finalizing block should be added then ?

bvbfan: It has https://github.com/DeFiCh/ain/pull/576/commits/d3ea5d201ad52d2cd2d7778d483a1696bea1fbe5#diff-97c3a52bc5fad452d82670a7fd291800bae20c7bc35bb82686c2c0a4ea7b5b98R2878
It gets all active nodes to finalizing height then count `yes` votes when masternode is part of active ones.

Jouzo: Ok sweet

Jouzo:
Can we merge `main` or `surangap/generatetoaddress` into this branch in order to be able to test listVotes on jellyfish side ?

src/masternodes/rpc_proposals.cpp
```diff
@@ -0,0 +1,622 @@
...
+    LockedCoinsScopedGuard lcGuard(pwallet); // no need here, but for symmetry
+
+    RPCTypeCheck(request.params, { UniValue::VSTR, UniValue::VSTR, UniValue::VSTR, UniValue::VARR }, true);
+
+    auto propId = ParseHashV(request.params[0].get_str(), "proposalId");
+    auto mnId = ParseHashV(request.params[1].get_str(), "masternodeId");
+    auto vote = CPropVoteType::VoteNeutral;
+
+    auto voteStr = request.params[2].get_str();
+    if (voteStr == "no") {
```
Jouzo, line 301: Can we check for uppercase "NO" instead to match CPropVoteToString return values ?
Right now, it takes 'no' as vote input and returns 'NO' in propVoteToJSON

bvbfan: We can do it case insensitive, all other rpcs gets lowercase as well as when you type commands you don't need to add upper case, mostly.

Jouzo: Ok I understand the logic. Would be nice to be case insensitive to have coherency between input and output

bvbfan:
Use #870 instead.

PR #599, opened 2021-07-07
ICX hardfork fixes

### LINKED ISSUE #555: ICX icx_submitexthtlc with error message
#### What happened:
```
./defi-cli icx_submitexthtlc '{"offerTx":"acc304ef3d494811df77c1587e07b5ee5a56df5bd0568c150fdcea69defec851","hash":"9a5a57656b47745fb6fabefc317be8d39efc553959d661cc6456d2db0336543a","amount":0.0001,"htlcScriptAddress":"3AqTafDs8VbAP3c9BoQivFbK6hbaGdbxdr","ownerPubkey":"02caa810fd93899abbf6b1691de94099301930ec4029c48101baaf67ee3b7e5e33","timeout":40}'
error code: -32600
error message:
Test ICXSubmitEXTHTLCTx execution failed:
negative amount: -0.00000021
```
It should success with transaction id

There are two places in the code may have problem.
https://github.com/DeFiCh/ain/blob/7a430185fd3029e9ee8974fde4e65c715fb9db04/src/masternodes/mn_checks.cpp#L1273
https://github.com/DeFiCh/ain/blob/7a430185fd3029e9ee8974fde4e65c715fb9db04/src/masternodes/mn_checks.cpp#L1396

### LINKED ISSUE #583: ICX not able to close offer for BTC/DFI order
#### What happened:
Trying to close offer on BTC/DFI order and returned:
`amount 0.01000000 is less than 10.00000000 (-32600)`

In initial implementation DFI funds were locked when having DFI offer. Lately it was changed but in closeoffer it wasn't removed to return funds from offer and the message is saying it is trying to sub the funds.
No security issues, offer will expire by itself. Only users won't be able to manually close it.

### LINKED ISSUE #595: ICX offer expiring showing in log it cannot sub from 0 balance
#### What happened:
`Can't subtract takerFee from offer txidAddr: amount 0.00000000 is less than 0.00036705`

These are shown in log when offer expires and trying to refund takerFee, but the atomic swap started at some point and burnt takerFee. It cannot refund something that doesn't exists, so there is no security issue.
Will be fixed in next hardfork release.

Mixa84:

/kind fix

PR contains fixes for the following issues:
- Not able to manually close offer on BTC/DFI order
- Trying to refund takerFee when offer expires if atomic swap already started and takerFee burnt. No more message in the log
- Always recalculating takerFee on HTLC create and leads to negative amount in refund which is not possible
- Not able to create tx if there is no change due to check for 2 vouts on every ICX tx
- Several small and rpc fixes
-

Fixes #555
Fixes #583 
Fixes #595 

#### Additional comments?:
Fixes will activate on EunosPaya height

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -1453,10 +1453,6 @@ public:
         if (!order)
             return Res::Err("order with creation tx %s does not exists!", offer->orderTx.GetHex());
 
-        auto exthtlc = mnview.HasICXSubmitEXTHTLCOpen(dfchtlc->offerTx);
-        if (!exthtlc)
-            return Res::Err("cannot claim, external htlc for this offer does not exists or expired!");
-
         // claim DFC HTLC to receiveAddress
         CalculateOwnerRewards(order->ownerAddress);
         CScript htlcTxidAddr(dfchtlc->creationTx.begin(), dfchtlc->creationTx.end());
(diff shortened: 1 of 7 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -1259,15 +1269,27 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                }
+            }
+            else
+            {
+                CAmount BTCAmount(static_cast<CAmount>((arith_uint256(submitdfchtlc.amount) * arith_uint256(order->orderPrice) / arith_uint256(COIN)).GetLow64()));
+                takerFee = CalculateTakerFee(BTCAmount);
+            }
 
             // refund the rest of locked takerFee if there is difference
             if (offer->takerFee - takerFee) {
```
ShengguangXiao, line 1295: If this value is negative then still have problem

Mixa84: It will go into else before EunosPaya, and after it will enter the if and calculate the takerFee as ratio of the current one so `offer->takerFee - takerFee` will never be negative.

surangap (dismissed):

surangap (approved):

PR #607, opened 2021-07-10
Timelock expiration

Bushstar:
If timelock has expired return 0 from GetTimelock. This approach will result in timelock expiration being calculated when checking proof-of-stake during block connection just once and only if the MN generating the block has a timelock value set. An alternative would be to check every single masternode on every connected block and zero any expired timelocks in the DB, this would also have to be reversed on reorg, this approach seems heavier.

The impact on stakers needs to be observed, will a single node running 50 MNs be able to stake all of them every second?

### DIFF
--- a/test/functional/feature_longterm_lockin.py
+++ b/test/functional/feature_longterm_lockin.py
@@ -105,13 +105,13 @@ class MasternodesTimelockTest (DefiTestFramework):
             self.nodes[0].resignmasternode(nodeid5)
         except JSONRPCException as e:
             errorString = e.error['message']
-        assert("Trying to resign masternode before expiration" in errorString)
+        assert("Trying to resign masternode before timelock expiration" in errorString)
 
         try:
             self.nodes[0].resignmasternode(nodeid10)
         except JSONRPCException as e:
             errorString = e.error['message']
-        assert("Trying to resign masternode before expiration" in errorString)
+        assert("Trying to resign masternode before timelock expiration" in errorString)
 
         # Time travel five years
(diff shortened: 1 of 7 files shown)


test/functional/feature_longterm_lockin.py
```diff
@@ -123,6 +123,9 @@ def run_test(self):
         result5 = self.nodes[0].getmasternode(nodeid5)
         assert_equal(result5[nodeid5]['state'], 'ENABLED')
 
+        # Timelock should no longer be present
+        assert_equal('timelock' not in result5[nodeid5], True)
+
         # Resign 5 year MN
```
uzyn, line 129: Can you add a test case that upon expiry of timelock, if the masternode does not resign, the multiplier does not accrue at faster speed. 

1. Get accrual speed before expiry. Ensure it's 1.5x or 2x.
2. Set mock time to past expiry.
3. Get accrual speed. It should still be 1.5x or 2x.
4. Publish a block from the masternode.
5. Get accrual speed. It should still now be 1x

uzyn (changes requested):
Can you add the test case please?

test/functional/feature_longterm_lockin.py
Bushstar: That test case is a problem in a Python test script, we cannot stake using the MNs we create in those scripts. We can see that the timelock is missing so that once the multiplier is wiped it should not accrue anymore, but starting a node using a masternode created in a test case is problematic.

hidiego (approved):

PR #618, opened 2021-07-19
Cap regtest blocktime to median time + 1 to avoid errors

Bushstar:
Regtest keeps the conventional backwards search logic for staking that can result in time-too-old errors, this logic is kept for compatibility with the testing framework, this commit prevents the minimum possible backwards time to be one more than median to avoid errors. Hopefully this will not interfere with tests.


prasannavl (dismissed):

prasannavl:
Thanks!

hidiego (approved):

bvbfan (approved):

PR #625, opened 2021-07-19
Fix to activate increase of offer exipry after EunoPaya

Mixa84:

/kind fix

Fixing offer expiry default period to be 20 blocks after EunosPaya height

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -1211,6 +1211,12 @@ public:
         if (!order)
             return Res::Err("order with creation tx " + makeoffer.orderTx.GetHex() + " does not exists!");
 
+        if (static_cast<int>(height) < consensus.EunosPayaHeight && makeoffer.expiry < CICXMakeOffer::DEFAULT_EXPIRY)
+            return Res::Err("offer expiry must be greater than %d!", CICXMakeOffer::DEFAULT_EXPIRY - 1);
+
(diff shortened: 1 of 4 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -1211,6 +1211,12 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
         if (!order)
             return Res::Err("order with creation tx " + makeoffer.orderTx.GetHex() + " does not exists!");
 
+        if (static_cast<int>(height) < consensus.EunosPayaHeight && makeoffer.expiry < CICXMakeOffer::DEFAULT_EXPIRY)
```
prasannavl, line 1214: Can we improve readability here a bit please, perhaps, by separating expiry out: 

```
        auto expiry = static_cast<int>(height) < consensus.EunosPayaHeight 
        ? CICXMakeOffer::DEFAULT_EXPIRY : 
        CICXMakeOffer::EUNOSPAYA_DEFAULT_EXPIRY;

        if (makeoffer.expiry < expiry) {
            return Res::Err("offer expiry must be greater than %d!", expiry - 1);
        }
```

Easier to reason with.

prasannavl (changes requested):

src/masternodes/mn_checks.cpp
Mixa84: Yeah, np. I did it like this somewhere else, but this was a quick fix so I rushed it little bit ;)

prasannavl: Haha, thank you! I think it's time we started slowly improve on the code base :) Once checked-in, it's very difficult to get any thing changed!

prasannavl (approved):

PR #628, opened 2021-07-20
Add support for subnode staking

Bushstar:
Note that after the EunosPaya hard fork the following entry in `getmininginfo`, `getmasternode` and `listmasternode`.

`"targetMultiplier": 1`

Will change to.

```
"targetMultipliers": [1, 1]
```

Each regular masternode will show two multiplier values, five year lock-ins will show three and ten year lock-ins will show four.


src/masternodes/rpc_masternodes.cpp
```diff
@@ -41,17 +41,17 @@ UniValue mnToJSON(uint256 const & nodeId, CMasternode const& node, bool verbose,
...
-            if (!stakerBlockTime) {
-                if (auto block = ::ChainActive()[Params().GetConsensus().DakotaCrescentHeight]) {
-                    stakerBlockTime = std::min(GetTime() - block->GetBlockTime(), Params().GetConsensus().pos.nStakeMaxAge);
+            // Get block times with next block as height
+            const auto subNodesBlockTime = pcustomcsview->GetBlockTimes(node.operatorAuthAddress, currentHeight + 1, node.creationHeight, timelock);
+
+            if (currentHeight >= Params().GetConsensus().EunosPayaHeight) {
+                const uint8_t loops = timelock == CMasternode::TENYEAR ? 4 : timelock == CMasternode::FIVEYEAR ? 3 : 2;
+                for (uint8_t i{0}; i < loops; ++i) {
+                    obj.pushKV(strprintf("multiplierSubnode%d", i), pos::CalcCoinDayWeight(Params().GetConsensus(), GetTime(), subNodesBlockTime[i]).getdouble());
```
uzyn, line 50: Can we make it as 2nd layer array? i.e. 

```js
multipliers: [12, 1, 42]
```

This is much better because you know exactly how many multipliers/subnodes you get, and a consumer can iterate through easily.

uzyn (changes requested):
Small comment on the RPC output.

src/masternodes/rpc_masternodes.cpp
Bushstar: That's is a much better idea.

src/pos_kernel.cpp
```diff
@@ -39,20 +37,38 @@ namespace pos {
...
         arith_uint256 targetProofOfStake;
         targetProofOfStake.SetCompact(nBits);
 
-        const auto hashProofOfStake = UintToArith256(CalcKernelHash(stakeModifier, creationHeight, coinstakeTime, masternodeID, params));
+        if (blockHeight >= static_cast<uint64_t>(params.EunosPayaHeight)) {
+            const uint8_t loops = timelock == CMasternode::TENYEAR ? 4 : timelock == CMasternode::FIVEYEAR ? 3 : 2;
+
+            // Check whether we meet hash for each subnode in turn
+            for (uint8_t i{0}; i < loops; ++i) {
+                const auto hashProofOfStake = UintToArith256(CalcKernelHashMulti(stakeModifier, creationHeight, coinstakeTime, masternodeID, loops));
```
ShengguangXiao, line 51: I think the parameter pass into CalcKernelHashMulti should be `i`, otherwise all subnode are mining the same hash.

Bushstar: Updated.

ShengguangXiao (dismissed):

PR #639, opened 2021-07-26
Fix delay for masternode activation/resign delay

### LINKED ISSUE #635: height in GetActivationDelay wrong
https://github.com/DeFiCh/ain/blob/1cdc7e05b8340fbd51ba76a1263a6e1fb194e6cd/src/masternodes/masternodes.cpp#L118-L124

Hi,

please check out the referenced code.
You are taking the ActivationDelay (and ResignDelay) from the current height, not from creationHeight. This can lead to problems when changing the delays in the future.

Mixa84:

/kind fix

When checking the state of MN, instead of getting the delay from params for current height it will use the height when MN was created/resigned. This way it will keep old delay and new delays set after some height will apply only on MNs create/resigned after that height.

Close #635

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -113,17 +113,20 @@ CMasternode::State CMasternode::GetState() const
 
 CMasternode::State CMasternode::GetState(int height) const
 {
+    int EunosPayaHeight = Params().GetConsensus().EunosPayaHeight;
+
     if (resignHeight == -1) { // enabled or pre-enabled
         // Special case for genesis block
-        if (creationHeight == 0 || (height < Params().GetConsensus().EunosPayaHeight && height >= creationHeight + GetMnActivationDelay(height))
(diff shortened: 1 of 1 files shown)


src/masternodes/masternodes.cpp
```diff
@@ -115,14 +115,16 @@ CMasternode::State CMasternode::GetState(int height) const
 {
     if (resignHeight == -1) { // enabled or pre-enabled
         // Special case for genesis block
-        if (creationHeight == 0 || height >= creationHeight + GetMnActivationDelay(height)) {
-            return State::ENABLED;
+        if (creationHeight == 0 || (height < Params().GetConsensus().EunosPayaHeight && height >= creationHeight + GetMnActivationDelay(height))
```
prasannavl, line 118: Readability: What do you think about lifting the `Params().GetConsensus().EunosPayaHeight` to a var on the top of the fn?

prasannavl (approved):

PR #645, opened 2021-07-29
Initalize spv connections under main lock

bvbfan:
/kind fix

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1781,6 +1781,9 @@ bool AppInitMain(InitInterfaces& interfaces)
             spv::pspv->Load();
         }
         panchors->Load();
+        if (spv::pspv) {
+            spv::pspv->Connect();
+        }
 
     } catch (const std::exception& e) {
         LogPrintf("%s\n", e.what());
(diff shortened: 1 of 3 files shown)


ShengguangXiao (dismissed):

src/spv/bitcoin/BRPeerManager.cpp
```diff
@@ -1431,27 +1447,30 @@ static void _peerRelayedBlock(void *info, BRMerkleBlock *block)
...
                 saveCount = (block->height % BLOCK_DIFFICULTY_INTERVAL) + BLOCK_DIFFICULTY_INTERVAL + 1;
                 _BRPeerManagerLoadMempools(manager);
             }
+            manager->lock.unlock();
         }
     }
 
     // Has last block changed?
-    if (!UInt256Eq(lastBlockHash, manager->lastBlock->blockHash))
+    if (!UInt256Eq(lastBlockHash, lastBlockHash))
```
Bushstar, line 1463: This will always be false. The test here is to see whether `manager->lastBlock` was updated in the preceding conditionals.

src/spv/bitcoin/BRPeerManager.cpp
```diff
@@ -1431,27 +1447,30 @@ static void _peerRelayedBlock(void *info, BRMerkleBlock *block)
...
     {
-        manager->blockNotify(manager->info, manager->lastBlock->blockHash);
+        manager->blockNotify(manager->info, lastBlockHash);
```
Bushstar, line 1465: Should be `manager->lastBlock->blockHash` if it differs from `lastBlockHash`.

src/spv/bitcoin/BRPeerManager.cpp
bvbfan: Will try another approach, this one is a bit aggressive.

Bushstar (approved):

PR #660, opened 2021-08-02
Add getvault + change id to Vault struct

hidiego:

/kind feature

Fixes CVault struct to include txId as id and be able to use in GetVault()

### DIFF
--- a/src/masternodes/loan.h
+++ b/src/masternodes/loan.h
@@ -122,7 +122,7 @@ using CVaultId = uint256;
 struct CVaultMessage {
     std::string ownerAddress;
     std::string schemeId;
-    bool isLiquidated{false};
+    bool isUnderLiquidation{false};
 
     ADD_SERIALIZE_METHODS;
 
@@ -143,7 +143,6 @@ struct CVault : public CVaultMessage
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action) {
         READWRITEAS(CVaultMessage, *this);
-        READWRITE(isLiquidated);
     }
 };
 class CLoanView : public virtual CStorageView {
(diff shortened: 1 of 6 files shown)


src/masternodes/loan.h
```diff
@@ -122,6 +122,7 @@ using CVaultId = uint256;
 struct CVaultMessage {
     std::string ownerAddress;
     std::string schemeId;
+    bool isLiquidated{false};
```
uzyn, line 125: `isLiquidated` is not a one-way switch. A vault can be under liquidation, and once successfully recovered its assets and get above minimum collateralization ratio, it can exit liquidation state. 

Probably `bool isUnderLiquidation` is a better name.

uzyn (changes requested):

src/masternodes/loan.h
```diff
@@ -122,7 +122,7 @@ using CVaultId = uint256;
 struct CVaultMessage {
     std::string ownerAddress;
     std::string schemeId;
-    bool isLiquidated{false};
+    bool isUnderLiquidation{false};
```
bvbfan, line 125: It should be serialized at line 134

src/masternodes/loan.h
```diff
@@ -135,7 +136,7 @@ struct CVaultMessage {
 
 struct CVault : public CVaultMessage
 {
-    bool isLiquidated{false};
+    CVaultId id{};
 
     ADD_SERIALIZE_METHODS;
```
bvbfan, line 142: It's a pointless struct now. Vault id should not be serialized nor part of value, it's the key otherwise it's waste of storage.

bvbfan (changes requested):

src/masternodes/loan.h
```diff
@@ -122,6 +122,7 @@ using CVaultId = uint256;
 struct CVaultMessage {
     std::string ownerAddress;
```
bvbfan, line 123: It should be `CScript` when it's shown should be used `ScriptToString`

bvbfan (changes requested):

bvbfan (approved):

Issue #666, opened 2021-08-11
Loan: liquidation trigger


uzyn:
Trigger vault liquidation if a vault falls below minimum collateralization ratio.

Covers:
- Liquidation trigger
- Automatically initiates auctions in $10k sizes.

PR #673, opened 2021-08-11
Collateralization ratio calculation

### LINKED ISSUE #621: Loan: Interest calculation
At the tick of every block, blockchain should keep track of the total amount of outstanding balance, with interest included.

Similar to DEX rewards, interest is calculated at every block. For performance reasons, it should be calculated lazily, i.e. at the tick of a block there is no need to calculate loan value for every single vaults.

Interest can come in 2 forms:
- Vault interest, based on loan scheme
- Loan token interest

See https://github.com/DeFiCh/pinkpaper/tree/main/loan#interest-rate

### Suggested data implementation

DB keeps track of
- Loan tokens taken out
- Interest balance
- Interest balance calculated at block
- Interest increment per block per loan tokens
- Collateralization ratio
- Liquidation block _(only need to trigger liquidation if block number is larger than this)_

Calculation needs to be triggered on the following events:

- Oracle price updates. This should happen every 5-15 minutes and not on every block. 
- Vault changes, e.g. more loan / more deposit.

bvbfan:
/kind feature

Fixes #621 #666

### DIFF
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -135,6 +135,11 @@ struct Params {
         static const uint32_t blocks = 60 * 60 * 24 / pos.nTargetSpacing;
         return blocks;
     }
+
+    uint32_t blocksCollateralizationRatioCalculation() const {
+        static const uint32_t blocks = 15 * 60 / pos.nTargetSpacing;
+        return blocks;
+    }
     /**
      * Minimum blocks including miner confirmation of the total of 2016 blocks in a retargeting period,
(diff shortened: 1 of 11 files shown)


hidiego (approved):
GetLoanSetLoanToken() GetLoanSetLoanTokenByID()  GetLoanSetCollateralToken() naming is very confusing could do with just GetLoanToken(), GetLoanTokenByID() and GetLoanCollateralToken()

Issue #656, opened 2021-08-02
Composite DEX swap, 1 transaction that swaps across multiple poolpairs.


uzyn:
User needs to provide the explicit path. Do not implement path finding algorithm as part of the blockchain consensus.

There should be RPC for it as well.

prasannavl:
Fixed by #676

PR #676, opened 2021-08-13
Feature composite swap

Bushstar:
This commit adds the ability to perform a single swap transaction that will swap coins between multiple pools involving as many as three pools total.

For example from LTC to DOGE with the following pool path.

LTC-DFI - DOGE-DFI

Or TSLA to LTC with the following pool path.

LTC-DFI - dUSD-DFI - TSLA-dUSD

Implements: https://github.com/DeFiCh/ain/issues/656

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -939,22 +939,7 @@ public:
             return Res::Err("tx must have at least one input from account owner");
         }
 
-        auto poolPair = mnview.GetPoolPair(obj.idTokenFrom, obj.idTokenTo);
-        if (!poolPair) {
-            return Res::Err("can't find the poolpair!");
-        }
-
-        CPoolPair& pp = poolPair->second;
-        return pp.Swap({obj.idTokenFrom, obj.amountFrom}, obj.maxPrice, [&] (const CTokenAmount &tokenAmount) {
-            auto res = mnview.SetPoolPair(poolPair->first, height, pp);
-            if (!res) {
-                return res;
-            }
-            CalculateOwnerRewards(obj.from);
-            CalculateOwnerRewards(obj.to);
-            res = mnview.SubBalance(obj.from, {obj.idTokenFrom, obj.amountFrom});
-            return !res ? res : mnview.AddBalance(obj.to, tokenAmount);
-        }, static_cast<int>(height));
+        return CPoolSwap(obj, height).ExecuteSwap(mnview, obj.poolIDs, this);
     }
 
     Res operator()(const CLiquidityMessage& obj) const {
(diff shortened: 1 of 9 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -2115,3 +2100,198 @@ bool IsMempooledCustomTxCreate(const CTxMemPool & pool, const uint256 & txid)
...
+            fromPoolsID.emplace(pool.idTokenA.v, id);
+        }
+
+        if (pool.idTokenA == obj.idTokenTo) {
+            toPoolsID.emplace(pool.idTokenB.v, id);
+        } else if (pool.idTokenB == obj.idTokenTo) {
+            toPoolsID.emplace(pool.idTokenA.v, id);
+        }
+        return true;
+        }, {0});
```
bvbfan, line 2121: Indentation.

src/masternodes/mn_checks.cpp
```diff
@@ -2115,3 +2100,198 @@ bool IsMempooledCustomTxCreate(const CTxMemPool & pool, const uint256 & txid)
...
+
+                // If a pool pairs matches from pair and to pair add it to the pool paths
+                if ((fromIt->first == pool.idTokenA.v && toIt->first == pool.idTokenB.v) ||
+                (fromIt->first == pool.idTokenB.v && toIt->first == pool.idTokenA.v)) {
+                    poolPaths.push_back({fromIt->second, id, toIt->second});
+                }
+            }
+        }
+        return true;
+        }, {0});
```
bvbfan, line 2170: Indentation.

src/masternodes/mn_checks.cpp
```diff
@@ -2115,3 +2100,198 @@ bool IsMempooledCustomTxCreate(const CTxMemPool & pool, const uint256 & txid)
...
+
+        // Test on copy of view
+        CCustomCSView dummy(view);
+
+        // Execute pool path
+        auto res = ExecuteSwap(dummy, path);
+
+        // Add error for RPC user feedback
+        if (!res) {
+            const auto token = pcustomcsview->GetToken(currentID);
```
bvbfan, line 2186: `dummy.GetToken` to be consistent.

src/masternodes/mn_checks.cpp
```diff
@@ -2115,3 +2100,198 @@ bool IsMempooledCustomTxCreate(const CTxMemPool & pool, const uint256 & txid)
...
+        // Record amount if more than previous or default value
+        if (res && result > bestPair.second) {
+            bestPair = {path, result};
+        }
+    }
+
+    return bestPair.first;
+}
+
+Res CPoolSwap::ExecuteSwap(CCustomCSView& view, std::vector<DCT_ID> poolIDs, const CCustomTxVisitor* txVisitor) {
```
bvbfan, line 2201: It doesn't need `txVisitor` just make a new cache over `view` and flush on success.
```
Res CPoolSwap::ExecuteSwap(CCustomCSView& mnview, const std::vector<DCT_ID>& poolIDs)
{
    CCustomCSView view(mnview);
    ...
    view.CalculateOwnerRewards(...);
    ...
    view.Flush(); // before successful return
}
```

bvbfan (changes requested):

src/masternodes/mn_checks.cpp
Bushstar: That's how CLion formats it. I'm oftening fighting with the IDE on things like this. Will update.

src/masternodes/mn_checks.cpp
```diff
@@ -2224,6 +2224,8 @@ Res CPoolSwap::ExecuteSwap(CCustomCSView& view, std::vector<DCT_ID> poolIDs) {
         poolPrice = obj.maxPrice;
     }
 
+    CCustomCSView intermediateView(view);
```
bvbfan, line 2227: Then use only that view in the loop otherwise SubBalance may fails due to distinct views. So just use intermediate view inside the loop

bvbfan (changes requested):

src/masternodes/mn_checks.cpp
Bushstar: Good point. Thanks for spotting that.

bvbfan (approved):

ShengguangXiao (approved):

PR #679, opened 2021-08-16
Add masternode `timelock` to `CCustomTxRpcVisitor`, now can be retrieved via `decodecustomtx` RPC.

surangap:

/kind fix

This will add masternode `timelock` to `CCustomTxRpcVisitor`. Then timelock data can be retrived with `decodecustomtx` or `getcustomtx` RPCs

Fixes #

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -166,6 +166,15 @@ std::string CMasternode::GetHumanReadableState(State state)
     }
 }
 
+std::string CMasternode::GetTimelockToString(TimeLock timelock)
+{
+  switch (timelock) {
+    case FIVEYEAR : return "FIVEYEARTIMELOCK";
+    case TENYEAR  : return "TENYEARTIMELOCK";
+    default       : return "NOTIMELOCK";
+  }
+}
+
 bool operator==(CMasternode const & a, CMasternode const & b)
 {
     return (a.mintedBlocks == b.mintedBlocks &&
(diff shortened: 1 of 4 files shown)


bvbfan (approved):

ShengguangXiao (dismissed):

Bushstar (approved):

prasannavl:
Thanks Suranga. 

@Bushstar - Are we cool with exposing the timelock in this format? MN views expose the timelocks as "<years> years" format. 

Pros: 
- Symmetry with RPC input. If we have to break the output RPC format, we likely would be breaking input RPC too. 

Cons:
- MN views use numerical format, which arguably might be more future proof.

surangap:
> Thanks Suranga.
> 
> @Bushstar - Are we cool with exposing the timelock in this format? MN views expose the timelocks as "<years> years" format.
> 
> Pros:
> 
> * Symmetry with RPC input. If we have to break the output RPC format, we likely would be breaking input RPC too.
> * Think all it's usage is just limited to the RPC areas.
> 
> Cons:
> 
> * MN views use numerical format, which arguably might be more future proof.
> * Anything else?
> 
> @surangap - any thoughts on using `NONE` instead of `NOTIMELOCK`? I do appreciate that you've picked a convention that goes with the existing timelock convention, but `NOTIMELOCK` also introduces a new keyword, I think -- do you think a generic NONE might work as well here?

I am fine with `NONE`. Also if you want to change it to the `<years> year` format, that would be okay too. I guess the timelock number is in weeks and we can work a simple logic to support all future values.

bvbfan (approved):

PR #682, opened 2021-08-16
Split loan / vault implementation

bvbfan:
/kind refactor


hidiego (approved):

src/masternodes/vault.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <flushablestorage.h>
+#include <masternodes/balances.h>
+#include <masternodes/res.h>
+#include <script/script.h>
+
+// use vault's creation tx for ID
+using CVaultId = uint256;
+
+struct CVaultMessage {
```
ShengguangXiao, line 15: Why this struct not name as `CCreateVaultMessage`?

hidiego: It is following convention in Loan but could be changed for better understanding of purpose

PR #686, opened 2021-08-18
Add spv_refundhtlcall

Bushstar:
Adds a call to refund all HTLCs in the wallet for which the correct private key is known and has enough confirmations.

Fixes a bug where only one HTLC output in a TX could be seen or spent.

### DIFF
--- a/src/spv/spv_rpc.cpp
+++ b/src/spv/spv_rpc.cpp
@@ -192,7 +192,7 @@ UniValue spv_createanchor(const JSONRPCRequest& request)
     result.pushKV("cost", cost);
     if (send) {
         result.pushKV("sendResult", sendResult);
-        result.pushKV("sendMessage", DecodeSendResult(sendResult));
+        result.pushKV("sendMessage", sendResult != 0 ? DecodeSendResult(sendResult) : "");
     }
 
     return result;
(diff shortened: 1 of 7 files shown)


src/spv/spv_wrapper.cpp
```diff
@@ -883,6 +885,11 @@ UniValue CSpvWrapper::GetHTLCReceived(const std::string& addr)
 
     auto htlcTransactions = BRListHTLCReceived(wallet, addressFilter);
 
+    // Sort by Bitcoin address
+    std::sort(htlcTransactions.begin(), htlcTransactions.end(), [](const std::pair<BRTransaction*, size_t>& lhs, const std::pair<BRTransaction*, size_t>& rhs){
+        return strcmp(lhs.first->outputs[lhs.second].address, rhs.first->outputs[rhs.second].address) < 0;
+    });
+
```
bvbfan, line 892: Why sorting? And why not in `BRListHTLCReceived` ?

Bushstar: Will still need to be sorted in BRListHTLCReceived in the same way. I was thinking of adding later an option for the user to specify sort order, like TXID, time, address and so on. I could have used all three in testing but address order sufficed.

bvbfan (approved):

src/spv/spv_rpc.cpp
```diff
@@ -1103,8 +1108,49 @@ UniValue spv_refundhtlc(const JSONRPCRequest& request)
...
+    result.pushKV("sendmessage", pair.second);
+    return result;
+}
+
+UniValue spv_refundhtlcall(const JSONRPCRequest& request)
+{
+    CWallet* const pwallet = GetWallet(request);
+
+    RPCHelpMan{"spv_refundhtlcall",
+               "\nGets all HTLC contracts stored in wallet and creates refunds transactions for a that have expired\n",
```
ShengguangXiao, line 1125: The help message reads not correct.

PR #688, opened 2021-08-19
Trigger auction, split auction into batches, initial auction bid data

### LINKED ISSUE #668: Loan: auction conclusion
Covers:
1. Auction conclusion settlement to all auctions winners.
2. Determines if a vault can exit liquidation state and re-enables it.

bvbfan:
/kind feature

Fixes #668 (data structures / auction trigger)

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2929,9 +2929,79 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                     for (const auto& col : collaterals.balances) {
                         cache.SubVaultCollateral(vaultId, {col.first, col.second});
                     }
+                    constexpr const uint64_t batchThreshold = 10000 * COIN; // 10k USD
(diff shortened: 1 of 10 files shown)


src/validation.cpp
```diff
@@ -2929,9 +2929,79 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                     for (const auto& col : collaterals.balances) {
                         cache.SubVaultCollateral(vaultId, {col.first, col.second});
                     }
+                    constexpr const uint64_t batchThreshold = 10000 * COIN; // 10k USD
```
ShengguangXiao, line 2932: With `constexpr` why still need `const`?

bvbfan: Actually it can be skipped, `constexpr` is evaluated at compile time, so when it's not `const` it can be changed compile time.

hidiego (approved):

hidiego (approved):

PR #700, opened 2021-08-30
Take loan feature

### LINKED ISSUE #671: Loan: take loan

Mixa84:

/kind feature

Initial development on `takeloan` function to "mint" tokens as a loan based on deposited collateral in vault.

Fixes #671

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -851,11 +851,14 @@ boost::optional<CCollateralLoans> CCustomCSView::CalculateCollateralizationRatio
     if (!vault || vault.val->isUnderLiquidation) {
         return {};
     }
+
     auto loanTokens = GetLoanTokens(vaultId);
     if (!loanTokens) {
         return {};
     }
+
     CCollateralLoans ret;
+
     for (const auto& loan : loanTokens->balances) {
         auto token = GetLoanSetLoanTokenByID(loan.first);
         assert(token);
@@ -867,6 +870,7 @@ boost::optional<CCollateralLoans> CCustomCSView::CalculateCollateralizationRatio
         auto value = loan.second + rate->interestToHeight + ((height - rate->height + 1) * rate->interestPerBlock);
         ret.loans.push_back({loan.first, MultiplyAmounts(price, value)});
     }
+
     for (const auto& col : collaterals.balances) {
         auto token = HasLoanSetCollateralToken({col.first, height});
         assert(token);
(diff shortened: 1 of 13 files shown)


Mixa84:
> So the difference is it may not contain loan tokens ? I don't see reason this function to exist, it's used in `getloan` if there is no loan tokens who cares for collaterals ?

The point was that `CalculateCollateralizationRatio` was not returning anything if there were no loans taken. I wanted to display total collateral value and total loan value. If we don't really need to see how much is in collateral if there is no loan then we can remove this function and I'll use `CalculateCollateralizationRatio`.

dcorral (approved):

src/masternodes/rpc_loan.cpp
```diff
@@ -970,6 +978,157 @@ UniValue getloanscheme(const JSONRPCRequest& request) {
...
+        auto collateral = pcustomcsview->GetCollateralAndLoanValue(vaultId, collaterals, height);
+
+        totalCollaterals += collateral->totalCollaterals();
+        totalLoans += collateral->totalLoans();
+
+        return true;
+    });
+
+    ret.pushKV("Collateral value (USD)",ValueFromAmount(totalCollaterals));
+    ret.pushKV("Loan value (USD)",ValueFromAmount(totalLoans));
```
ShengguangXiao, line 1126: Seems we don't use blanks in our keys.

Mixa84: Forgot, I will correct it.

bvbfan:
> > So the difference is it may not contain loan tokens ? I don't see reason this function to exist, it's used in `getloan` if there is no loan tokens who cares for collaterals ?
> 
> The point was that `CalculateCollateralizationRatio` was not returning anything if there were no loans taken. I wanted to display total collateral value and total loan value. If we don't really need to see how much is in collateral if there is no loan then we can remove this function and I'll use `CalculateCollateralizationRatio`.

Yep, there is nothing to calculate when there is no loan tokens, it should be the same for `getloan`, there is nothing to return when no loans.

src/masternodes/masternodes.cpp
```diff
@@ -847,6 +851,45 @@ boost::optional<CCollateralLoans> CCustomCSView::CalculateCollateralizationRatio
...
+            assert(rate && rate->height <= height);
+            auto oracle = GetOracleData(token->priceFeedTxid);
+            assert(oracle);
+            auto price = GetOraclePriceUSD(*oracle.val, token->symbol);
+            auto value = loan.second + rate->interestToHeight + ((height - rate->height + 1) * rate->interestPerBlock);
+            ret.loans.push_back({loan.first, MultiplyAmounts(price, value)});
+        }
+    }
+
     return ret;
```
bvbfan, line 893: Merge this function with main one.

src/masternodes/masternodes.h
```diff
@@ -287,6 +287,7 @@ struct CCollateralLoans {
     std::vector<CTokenAmount> loans; // in USD
 
     uint32_t ratio() const {
+        if (loans.empty()) return 0;
```
bvbfan, line 290: Return `std::numeric_limits<uint32_t>::max` because when only collaterals are in ratio is the higher one.

bvbfan (changes requested):

bvbfan:
Add `LoanTakeLoanCreationTx` and `LoanTakeLoanVaultKey` to `CheckPrefix` `CLoanView`

bvbfan (approved):

PR #707, opened 2021-09-02
Check subnode against key when getting its time

Bushstar:
After FortCanning when getting the last block time for each subnode we now make sure that the subnode number in the key matches the subnode we are iterating over.

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -400,8 +400,11 @@ std::vector<int64_t> CMasternodesView::GetSubNodesBlockTime(const CKeyID & minte
     for (uint8_t i{0}; i < SUBNODE_COUNT; ++i) {
         ForEachSubNode([&](const SubNodeBlockTimeKey &key, int64_t blockTime)
         {
-            if (key.masternodeID == nodeId)
-            {
+            if (height >= Params().GetConsensus().FortCanningHeight) {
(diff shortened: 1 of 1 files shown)


ShengguangXiao (approved):

src/masternodes/masternodes.cpp
```diff
@@ -400,8 +400,11 @@ std::vector<int64_t> CMasternodesView::GetSubNodesBlockTime(const CKeyID & minte
     for (uint8_t i{0}; i < SUBNODE_COUNT; ++i) {
         ForEachSubNode([&](const SubNodeBlockTimeKey &key, int64_t blockTime)
         {
-            if (key.masternodeID == nodeId)
-            {
+            if (height >= Params().GetConsensus().FortCanningHeight) {
+                if (key.masternodeID == nodeId && key.subnode == i) {
+                    times[i] = blockTime;
+                }
+            } else if (key.masternodeID == nodeId) {
```
bvbfan, line 407: I would prefer less code
```
auto subnode = height < Params().GetConsensus().FortCanningHeight || key.subnode == i;
if (key.masternodeID == nodeId && subnode)
```

prasannavl: I hear you, but in this particular case, I'm inclined to preserve the clarity of the current code, branching on FC - much lesser cognitive effort to rationalize.

PR #709, opened 2021-09-02
Drop usage of undos in rewards history

### LINKED ISSUE #699: Keep history of liquidity shares in history
Right now rewards prior last checkpoint are not shown or incorrect due to undo data pruning. It can be fixed by maintain full undo history by `-fcheckpoints=0` and reindex. By adding users shares in account history calculation of rewards can be done for unlimited time in the past.

bvbfan:
/kind feature

Fixes #699 

#### Additional comments?:

`-checkpoints` became obsolete and does nothing, it prints warning message for everyone explicit set it.
Rewards history restore balances and shares based on account history


Lysanthe:
I installed your client with that fix, and synchronized with reindex.

It now shows old rewards BUT also shows rewards where there shouldn't be one;)

E.G: dP8byuUexSdhaPMLsFa3vzfPjNRExKy8x1

The Rewards start with 2021-02-08. That seems fine so far and where not displayed with the official node software.

But after 2021-07-06 there all LP habe been sold, there are still rewards being returned until now.

bvbfan:
@Lysanthe  thanks for your test. You don't need to reindex, i agree with the results will try to find out a way to fix it. The problem is pre-Eunos remove liquidity could be problematic due to rewards was distributed per block (vs lazy now) so i will update the PR soon.

Lysanthe:
After further "research" it also seems that the the rewards returned before the last "RemovePoolLiquidity" is incorrect. 

Official client:
owner": "dP8byuUexSdhaPMLsFa3vzfPjNRExKy8x1",
      "blockHeight": 986983,
      "blockHash": "1254fa48f3e34448f9a59c78356f7cfbe1aaa5cadcb616958893927959c0e5b0",
      "blockTime": 1625585353,
      "type": "Rewards",
      "rewardType": "Coinbase",
      "poolID": "5",
      "amounts": [
        "0.00133145@DFI"
      ]
    },

New "feature":
 owner: "dP8byuUexSdhaPMLsFa3vzfPjNRExKy8x1",
        block_height: 986983,
        block_hash: "1254fa48f3e34448f9a59c78356f7cfbe1aaa5cadcb616958893927959c0e5b0",
        block_time: 2021-07-06T15:29:13,
        cat_name: "Rewards",
        cat_reward: Some(
            "Coinbase",
        ),
        txn: None,
        txid: None,
        pool_id: Some(
            "5",
        ),
        amounts: [
            Amount {
                token: "DFI",
                qty: 0.04384999,

bvbfan:
Official is incorrect here, to work it's needed reindex with `checkpooints=0` the new one works without.

Lysanthe:
> Official is incorrect here, to work it's needed reindex with `checkpooints=0` the new one works without.

Puhhhh. That's gonna be hard work for us all, to make everything correct.
So what does "checkpoints=0" in the current (official) implementation? My assumption was, that with checkpoints=0 alle rewards until the start of the liquidity mining (Nov/Dec 2020) are available (and correct). But that was obviously not true, since there's no history for current implementation.

I currently trying a different approach. I hacked the code and removed the part, where I think the pruning is done. That should increase the size, but all data should then be available, correct? (pleaassse let me be correct ;)

bvbfan:
> I currently trying a different approach. I hacked the code and removed the part, where I think the pruning is done. That should increase the size, but all data should then be available, correct? (pleaassse let me be correct ;)
Don't do it, if it was better we was doing by default.

Lysanthe:
Hi, 
I just  experimented with your latest commit (duno whether this mend  to be tested yet;))
The stopping of the rewards seems to work now. But they start to early.
Adr: dP8byuUexSdhaPMLsFa3vzfPjNRExKy8x1
Rewards are returned beginning from block 629983, when the first DFI where send to this address.  They shouldn't start before block 630346 with AddPoolLiquidity command

Lysanthe:
Created node from latest commits. Looking good so far now. Wasn't able analyse in every detail.
If anything comes up I'll let you know. 
Thanks so far. Hope this feature will make it in production soon

PR #711, opened 2021-09-03
Don't show untrusted txs in account history

bvbfan:
/kind fix

Fixes https://github.com/cakedefi/defichain-private/issues/679


ShengguangXiao (approved):

prasannavl (changes requested):
What do you think about about making `trusted` an optional parameter? And so probably good to have the default to false, so there are no incompatibilities.

bvbfan:
`listtransations` can be used to see all txs, here is better to not have untrusted

prasannavl:
I see your point. Let's merge it into 1.9.x / 2.0 branch instead since this is a breaking change.

bvbfan:
Sure it's best candidate to 2.0

PR #713, opened 2021-09-06
Fix without default loan schema node crash

ShengguangXiao:

/kind fix

1. Fix without default loan schema node crash.
2. Fix `takeloan` parameter `vaultId` in help message is incorrect.

Fixes #

### DIFF
--- a/src/masternodes/rpc_vault.cpp
+++ b/src/masternodes/rpc_vault.cpp
@@ -92,8 +92,6 @@ UniValue createvault(const JSONRPCRequest& request) {
         }
     }
     vault.ownerAddress = DecodeScript(ownerAddress);
-    const auto defaultScheme = pcustomcsview->GetDefaultLoanScheme();
-    vault.schemeId = defaultScheme ? defaultScheme.get() : "";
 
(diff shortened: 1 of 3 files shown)


src/masternodes/rpc_vault.cpp
```diff
@@ -92,10 +92,11 @@ UniValue createvault(const JSONRPCRequest& request) {
         }
     }
     vault.ownerAddress = DecodeScript(ownerAddress);
-    vault.schemeId = pcustomcsview->GetDefaultLoanScheme().get();
+    const auto defaultScheme = pcustomcsview->GetDefaultLoanScheme();
+    vault.schemeId = defaultScheme ? defaultScheme.get() : "";
```
bvbfan, line 96: Just remove it, `mn_check.cpp`

bvbfan (changes requested):

src/masternodes/rpc_vault.cpp
ShengguangXiao: Updated

bvbfan (approved):

PR #727, opened 2021-09-09
Add cargo build to makefile

Jouzo:

/kind feature

Adds cargo build to makefile.
By running `make ./src/defi-rust-cli`, one can now build a small cli example in Rust.
This is the first step to integrate Rust into `ain` codebase.


prasannavl:
Nice!

Couple of quick notes:

1. Let's just call it `cli` and move it into `src/rust` dir. No need to prefix it with `rust-`, and we can have all things rust in there, so it's easy to make a cargo workspace later.
2. As a next step, could you start off by actually calling the C api for one of these rpcs instead of using the btc client (we'll have to write the glue structures in rust).  
3. Let's use structopt for the cli.

hidiego (commented):
Looks great! :clap:

I would just suggest using the [Builder Pattern](https://github.com/clap-rs/clap#using-builder-pattern) or [Macros](https://github.com/clap-rs/clap#using-macros) to avoid extra dependencies. Also most probably it will be easier in the future for programmatically generating new rpc commands/templates.

Good start here truly!

prasannavl:
@hidiego - No need for either of those: structopt enums and structs - https://github.com/TeXitoi/structopt/blob/master/examples.

hidiego:
so no clap?

hidiego:
> @hidiego - No need for either of those: structopt enums and structs - https://github.com/TeXitoi/structopt/blob/master/examples. 

no clap then?

prasannavl:
structopt is on top of clap.rs. You'll almost never end up using it directly, unless needed. It's all handled by structopt.

Jouzo:
Closing for new PR limited to build process

PR #732, opened 2021-09-10
getvault show interest + #731 fix

### LINKED ISSUE #731: Createvault error message update.
https://github.com/DeFiCh/ain/blob/a0d49b69cf4907d035419a812f15493e1fa812f9/src/masternodes/mn_checks.cpp#L2063
currently, it returns the hex of the script.
Should either send the `address` of the owner which is more straightforward or should remove `obj.ownerAddress.GetHex()` part as other places which returns the same error.

hidiego:

/kind fix

Interest of loantoken is shown on getvault
Error message fixed #731 

Fixes #731

### DIFF
--- a/test/functional/feature_loan_vault.py
+++ b/test/functional/feature_loan_vault.py
@@ -5,11 +5,10 @@
 # file LICENSE or http://www.opensource.org/licenses/mit-license.php.
 """Test Loan Scheme."""
 
-from decimal import Decimal
 from test_framework.test_framework import DefiTestFramework
 
 from test_framework.authproxy import JSONRPCException
-from test_framework.util import assert_equal, assert_greater_than
+from test_framework.util import assert_equal
 import calendar
 import time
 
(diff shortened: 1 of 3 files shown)


test/functional/feature_loan_vault.py
```diff
@@ -348,10 +347,6 @@ def run_test(self):
 
         self.nodes[0].generate(1)
         self.sync_blocks()
-        vault1 = self.nodes[0].getvault(vaultId1)
-        assert_equal(vault1['loanAmount'], ['0.50000000@TSLA'])
-        assert_equal(vault1['collateralValue'], Decimal(2.00000000))
-        assert_greater_than(vault1['loanValue'],Decimal(0.5))
```
ShengguangXiao, line 354: Why remove but not update the value check?

hidiego: Sure! forgot to update! will do now

bvbfan (approved):

PR #742, opened 2021-09-14
Aggregate oracle price

Mixa84:

/kind fix

Changes priceFeedId filed in `setcollateraltoken` and `setloantoken` tx to be `token/currency` pair instead of txid of an oracle.

### DIFF
--- a/src/masternodes/rpc_loan.cpp
+++ b/src/masternodes/rpc_loan.cpp
@@ -1,5 +1,7 @@
 #include <masternodes/mn_rpc.h>
 
+#define MINIMUM_PRICEFEED_LENGTH 8
+
 extern UniValue tokenToJSON(DCT_ID const& id, CTokenImplementation const& token, bool verbose);
 
 UniValue setCollateralTokenToJSON(CLoanSetCollateralTokenImplementation const& collToken)
@@ -11,7 +13,7 @@ UniValue setCollateralTokenToJSON(CLoanSetCollateralTokenImplementation const& c
         return (UniValue::VNULL);
     collTokenObj.pushKV("token", token->CreateSymbolKey(collToken.idToken));
     collTokenObj.pushKV("factor", ValueFromAmount(collToken.factor));
-    collTokenObj.pushKV("priceFeedId", collToken.priceFeedTxid.GetHex());
+    collTokenObj.pushKV("priceFeedId", collToken.priceFeed.first + "/" + collToken.priceFeed.second);
     if (collToken.activateAfterBlock)
         collTokenObj.pushKV("activateAfterBlock", static_cast<int>(collToken.activateAfterBlock));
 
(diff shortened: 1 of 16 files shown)


src/masternodes/rpc_loan.cpp
```diff
@@ -446,7 +460,13 @@ UniValue updateloantoken(const JSONRPCRequest& request) {
...
         loanToken->name = trim_ws(metaObj["name"].getValStr());
     if (!metaObj["priceFeedId"].isNull())
-        loanToken->priceFeedTxid = uint256S(metaObj["priceFeedId"].getValStr());
+    {
+        std::string tmp = trim_ws(metaObj["priceFeedId"].getValStr());
+        if (tmp.length() < MINIMUM_PRICEFEED_LENGTH || tmp.find("/") == tmp.npos)
+            throw JSONRPCError(RPC_INVALID_PARAMETER,"price feed not in valid format - token/currency!");
+        loanToken->priceFeed.first = tmp.substr(0, tmp.find("/"));
+        loanToken->priceFeed.second = tmp.substr(tmp.find("/")+1, tmp.length()-1);
+    }
```
hidiego, line 469: This code is repeated several times, maybe is better to use a function.

src/masternodes/rpc_loan.cpp
```diff
@@ -1,5 +1,7 @@
 #include <masternodes/mn_rpc.h>
 
+#define MINIMUM_PRICEFEED_LENGTH 8
```
hidiego, line 3: DFI/USD for example would throw error, consider lowering this number.

dcorral (commented):
The price feeds are not checked against oracles anyhow so you can set loan/collateral tokens with unsupported pairs.

src/masternodes/rpc_loan.cpp
```diff
@@ -29,14 +29,34 @@ UniValue setLoanTokenToJSON(CLoanSetLoanTokenImplementation const& loanToken, DC
...
-    loanTokenObj.pushKV("priceFeedId", loanToken.priceFeedTxid.GetHex());
+    loanTokenObj.pushKV("priceFeedId", loanToken.priceFeed.first + "/" + loanToken.priceFeed.second);
     loanTokenObj.pushKV("interest", ValueFromAmount(loanToken.interest));
 
     UniValue ret(UniValue::VOBJ);
     ret.pushKV(loanToken.creationTx.GetHex(), loanTokenObj);
     return (ret);
 }
 
+PriceFeedPair DecodePriceFeed(const UniValue& value)
```
ShengguangXiao, line 40: Should add some unit test for this function

bvbfan: All the tests do it.

ShengguangXiao (approved):

PR #751, opened 2021-09-21
Split make.sh build, fix lints, cleanup

prasannavl:
/kind refactor

- Split make.sh build into deps, conf and make, so it's easier to split these for other tasks analyzers, lints, etc. 
- Simplify extra vars for make, conf and deps: 
  - `EXTRA_MAKE_ARGS => MAKE_ARGS`
  - `EXTRA_MAKE_CONF_ARGS => MAKE_CONF_ARGS`
  - `EXTRA_MAKE_DEPENDS_ARGS => MAKE_DEPS_ARGS`
- Introduce `MAKE_JOBS` var to target all parallel jobs at once.
- Introduce `MAKE_DEBUG` var as convenience to get debug builds (suffixes MAKE_CONF_ARGS with `--enable-debug`)
- Introduce `MAKE_COMPILER` that prepends the compiler flags to MAKE_CONF_ARGS, defaults to clang. 
- Cleanup redundant QT left over args
- Fix existing shell lints, primarily quotes related. 
- Mild refactor to pkg_install commands (that's only for convenience use - never used directly at the time of this), specifically to cleanup sudo, so it can be used in limited container env which disallow sudo)
- Add `HOST` to depends so target is carried forward to keep it platform agnostic (even though we don't really use this anywhere else)


ShengguangXiao:
So this change will use GCC as the default compiler again? Should we check the minimum GCC version?

bvbfan (changes requested):
GCC version should be 10.3 or Clang due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68070

prasannavl:
More PRs coming soon. Base updates that push the gcc up, and have better consistency to use the same compilers for CI tests and builds. 

This change doesn't affect the CI builds though. The docker images still use clang. It only affects local builds.

PR #752, opened 2021-09-22
Including tx fee into block rewards

ShengguangXiao:

/kind fix

Tx fees not included in the block rewards

Fixes # 

https://github.com/cakedefi/defichain-private/issues/681

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -239,7 +239,7 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
         coinbaseTx.vout.resize(2);
 
         // Explicitly set miner reward
-        coinbaseTx.vout[0].nValue = CalculateCoinbaseReward(blockReward, consensus.dist.masternode);
+        coinbaseTx.vout[0].nValue = nFees + CalculateCoinbaseReward(blockReward, consensus.dist.masternode);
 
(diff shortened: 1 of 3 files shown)


src/miner.cpp
```diff
@@ -239,7 +239,7 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
         coinbaseTx.vout.resize(2);
 
         // Explicitly set miner reward
-        coinbaseTx.vout[0].nValue = CalculateCoinbaseReward(blockReward, consensus.dist.masternode);
+        coinbaseTx.vout[0].nValue = nFees + CalculateCoinbaseReward(blockReward, consensus.dist.masternode);
```
bvbfan, line 242: It should be covered by hardfork hiehgt

bvbfan (changes requested):

src/miner.cpp
bvbfan: If we go without hardfork, newer nodes will generate bigger reward than older one, It's inconstancy.

ShengguangXiao: I will make the change in another PR based on 1.9.x

ShengguangXiao:
Close this in favor of https://github.com/DeFiCh/ain/pull/764

PR #762, opened 2021-09-28
Allow partial paybacks

### LINKED ISSUE #761: Update interest calculation to work better with partial payback

bvbfan:
/kind fix

Fixes #761

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -2304,6 +2304,9 @@ public:
         if (!res)
             return res;
 
+        if (!IsVaultPriceValid(mnview, obj.vaultId, height))
+            return Res::Err("Cannot update vault when any of the asset's price is invalid");
+
         if (auto collaterals = mnview.GetVaultCollaterals(obj.vaultId))
         {
             uint64_t totalDFI = 0, totalCollaterals = 0;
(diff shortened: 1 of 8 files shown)


test/functional/feature_loan_takeloan.py
```diff
@@ -336,7 +342,33 @@ def run_test(self):
         loans = self.nodes[0].getloaninfo()
 
         assert_equal(loans['collateralValueUSD'], Decimal('3000.00000000'))
-        assert_equal(loans['loanValueUSD'], Decimal('15.00017100'))
+        assert_equal(loans['loanValueUSD'], Decimal('15.00018040'))
```
Mixa84, line 345: Why have loan value changed with changing the way interest is stored? I see that number of blocks is the same up to this point, shouldn't the value be the same?
BTW, we haven't calced the amount it should show here but we just took the value test gave us. We should probably make a checkpoint for interest that we know how much it should be ....

bvbfan: The value should not be the same, partial payback not covers full amount, say `interestPerBlock` still count.

Mixa84: Ok, but we should still have some known value for interest (for example for loan on 10 TSLA and certain price in USD, after 10 or 100 blocks the interest should be this much) so we can detect in tests when calculation is not right.

PR #766, opened 2021-09-30
Poolswap height protection

bvbfan:
/kind feature


src/masternodes/rpc_customtx.cpp
```diff
@@ -185,6 +185,22 @@ class CCustomTxRpcVisitor : public boost::static_visitor<void>
         rpcInfo.pushKV("maxPrice", ValueFromAmount((obj.maxPrice.integer * COIN) + obj.maxPrice.fraction));
     }
 
+    void operator()(const CPoolSwapMessageV2& obj) const {
+        (*this)(obj.swapInfo);
```
Bushstar, line 189: Unused statement.

bvbfan: It used, it calls `void operator()(const CPoolSwapMessage& obj)`

Bushstar: Makes sense now, I was not seeing the second round brackets as a call operator!

Bushstar (approved):

PR #770, opened 2021-10-05
compile fix for current libboost

adrianschroeter:
/kind fix

fixes compile issues using current boost library (I haven't check which one exactly, but 1.77 broke and 1.66 keeps working)

### DIFF
--- a/src/fs.cpp
+++ b/src/fs.cpp
@@ -208,7 +208,7 @@ void ofstream::close()
 }
 #else // __GLIBCXX__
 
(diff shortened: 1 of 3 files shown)


adrianschroeter:
nah, it breaks older boost versions ... sorry, taking it back

adrianschroeter:
just for the record, it is fixed properly in bitcoin code in commit acb7aad27e . So you want to cherrypick this one most likely.

PR #771, opened 2021-10-05
Close vault

bvbfan:
/kind feature


src/masternodes/rpc_vault.cpp
```diff
@@ -167,7 +145,101 @@ UniValue createvault(const JSONRPCRequest& request) {
     CMutableTransaction rawTx(txVersion);
 
     CTransactionRef optAuthTx;
-    std::set<CScript> auths{vault.ownerAddress};
+    std::set<CScript> auths;
```
hidiego, line 148: Need provided ownerAddress auth

PR #772, opened 2021-10-06
Combine rewards, add new GovVars and height activated GovVars

Bushstar:


src/masternodes/govvariables/lp_loan_splits.h
```diff
@@ -0,0 +1,53 @@
...
+            splits.clear();
+            for (auto it = serliazedSplits.begin(); it != serliazedSplits.end(); /* advance */) {
+                splits.emplace(DCT_ID{it->first}, it->second);
+                serliazedSplits.erase(it++);
+            }
+        } else {
+            for (const auto& it : splits) {
+                serliazedSplits.emplace(it.first.v, it.second);
+            }
+            READWRITE(serliazedSplits);
```
bvbfan, line 46: It was for compatibility with old tx format, when we introduce new type, we can use simple `READWRITE`

Bushstar: Got it, moving from old `std::map<uint32_t, CAmount> serliazedSplits` to `std::map<DCT_ID, CAmount> splits`.

PR #826, opened 2021-10-25
Migrate masternode db records to V2

bvbfan:
/kind feature


bvbfan:
Overridden by #934 We will consider to close.

PR #866, opened 2021-11-03
Check against negative price

bvbfan:
/kind fix


bvbfan:
Defend by price check few lines above.

PR #869, opened 2021-11-03
change CreateCfp prefix to resolve conflict with CloseVault prefix

### LINKED ISSUE #816: CreatetCfp and closeVault should use different letters
#### What happened:

CreatetCfp and closeVault use the same letter => ‘e’. Hence, they have the same opcode

They should not share the same letter

Jouzo:

/kind chore

change CreateCfp prefix to resolve conflict with CloseVault prefix

#### what issue does it fix:

Fixes #816


Bushstar (approved):

bvbfan:
Use #870 instead.

PR #870, opened 2021-11-03
On-Chain-Governance

bvbfan:
/kind feature

Fixes #576 #869

### DIFF
--- a/src/masternodes/proposals.cpp
+++ b/src/masternodes/proposals.cpp
@@ -7,7 +7,7 @@
 std::string CPropTypeToString(const CPropType status)
 {
     switch(status) {
-    case CPropType::CommunityFundRequest:   return "CommunityFundRequest";
+    case CPropType::CommunityFundProposal:  return "CommunityFundProposal";
     case CPropType::BlockRewardReallocation:return "BlockRewardReallocation";
     case CPropType::VoteOfConfidence:       return "VoteOfConfidence";
     }
--- a/src/masternodes/proposals.cpp
+++ b/src/masternodes/proposals.cpp
@@ -2,6 +2,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file LICENSE or http://www.opensource.org/licenses/mit-license.php.
 
+#include <chainparams.h>
 #include <masternodes/proposals.h>
 
 std::string CPropTypeToString(const CPropType status)
@@ -34,7 +35,7 @@ std::string CPropVoteToString(const CPropVoteType vote)
     return "Unknown";
 }
 
-Res CPropsView::CreateProp(const CPropId& propId, uint32_t height, const CCreatePropMessage& msg, uint32_t votingPeriod)
+Res CPropsView::CreateProp(const CPropId& propId, uint32_t height, const CCreatePropMessage& msg)
 {
     CPropObject prop;
     static_cast<CCreatePropMessage&>(prop) = msg;
@@ -42,6 +43,7 @@ Res CPropsView::CreateProp(const CPropId& propId, uint32_t height, const CCreate
     auto key = std::make_pair(uint8_t(CPropStatusType::Voting), propId);
     WriteBy<ByStatus>(key, uint8_t(1));
 
+    auto votingPeriod = GetVotingPeriod();
     for (uint8_t i = 1; i <= prop.nCycles; ++i) {
         height += (height % votingPeriod) + votingPeriod;
         auto key = std::make_pair(height, propId);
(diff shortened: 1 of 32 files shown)


src/masternodes/proposals.cpp
```diff
@@ -0,0 +1,180 @@
...
+// Distributed under the MIT software license, see the accompanying
+// file LICENSE or http://www.opensource.org/licenses/mit-license.php.
+
+#include <masternodes/proposals.h>
+
+std::string CPropTypeToString(const CPropType status)
+{
+    switch(status) {
+    case CPropType::CommunityFundRequest:   return "CommunityFundRequest";
+    case CPropType::BlockRewardRellocation: return "BlockRewardRellocation";
```
Bushstar, line 11: Very minor issue, Rellocation is a typo, it should be Reallocation.

src/validation.cpp
```diff
@@ -2811,6 +2828,88 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
+                        case CPropType::BlockRewardRellocation:
+                            majorityThreshold = chainparams.GetConsensus().props.brp.majorityThreshold;
+                            break;
+                        case CPropType::VoteOfConfidence:
+                            majorityThreshold = chainparams.GetConsensus().props.voc.majorityThreshold;
+                            break;
+                        default:
+                            assert(false);
+                    }
+                    auto approved = lround(voteYes * 10000.f / voters) >= majorityThreshold;
```
Bushstar, line 2886: Pink paper says simple majority required for CFP, so yes vote should be more than 50%, not equal to it. 50% would be an equal split of yes and no and should not side with a yes vote under this circumstance. The pink paper says the same for VoC and block reallocation, that it should be more than 66.67% and not equal, though in this case more than or equal is less disagreeable.

src/masternodes/rpc_proposals.cpp
```diff
@@ -0,0 +1,601 @@
...
+
+/*
+ *  Issued by: any
+*/
+UniValue createcfp(const JSONRPCRequest& request)
+{
+    auto pwallet = GetWallet(request);
+
+    RPCHelpMan{"createcfp",
+               "\nCreates a Cummunity Fund Request" +
```
Bushstar, line 39: Typo, should be "Community". Also createcfp should be createcfr as it is a community fund request. Pinkpaper refers to the call as creategovcfr so uses cfr as well. I think we can take some liberty with the actual call name, but if we use cfp then the help should be renamed Community Fund Proposal.

bvbfan: Pinkpaper should be updated to show correct nick name `createcfp`, i will update the help, it should be `Community Fund Proposal`.

Bushstar (approved):

src/validation.cpp
```diff
@@ -3291,6 +3311,91 @@ void CChainState::ProcessGovEvents(const CBlockIndex* pindex, CCustomCSView& cac
...
+        return;
+    }
+
+    if (pindex->nHeight == chainparams.GetConsensus().GreatWorldHeight) {
+        auto balance = cache.GetBalance(chainparams.GetConsensus().foundationShareScript, DCT_ID{0});
+        if (balance.nValue > 0) {
+            cache.SubBalance(chainparams.GetConsensus().foundationShareScript, balance);
+            cache.AddCommunityBalance(CommunityAccountType::CommunityDevFunds, balance.nValue);
+        }
+    }
```
bvbfan, line 3325: I will gonna to remove that check, whether foundation receives tokens they will be transferred to `Community Dev Funds`

prasannavl: yup, sounds good.

src/validation.cpp
```diff
@@ -2585,8 +2602,8 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
         {
             if (tx.vout[j].scriptPubKey == Params().GetConsensus().burnAddress)
             {
-                pburnHistoryDB->WriteAccountHistory({Params().GetConsensus().burnAddress, static_cast<uint32_t>(pindex->nHeight), i},
-                                            {block.vtx[i]->GetHash(), static_cast<uint8_t>(CustomTxType::None), {{DCT_ID{0}, tx.vout[j].nValue}}});
+                pburnHistoryDB->WriteAccountHistory({tx.vout[j].scriptPubKey, static_cast<uint32_t>(pindex->nHeight), i},
+                                                    {block.vtx[i]->GetHash(), static_cast<uint8_t>(CustomTxType::None), {{DCT_ID{0}, tx.vout[j].nValue}}});
```
chanakasameera, line 2606: Will this cause the issue "`AccountHistory.blockHash` returned from `getaccounthistory` becomes undefined" ?
Related jellyfish test: https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-api-core/__tests__/category/account/getAccountHistory.test.ts#L46

chanakasameera (changes requested):

chanakasameera (changes requested):
Looks like `getmasternodeblocks` does not work as expected, it should return the number of blocks minted by a masternode ?, but now it returns a higher value.
Related jellyfish test: https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-api-core/__tests__/category/masternode/getMasternodeBlocks.test.ts#L178

src/validation.cpp
bvbfan: No, it shouldn't

bvbfan:
Unrelated to current PR

PR #873, opened 2021-11-05
getcustomtxcodes RPC

### LINKED ISSUE #806: Add RPC method to return available tx type indicator
We need all available tx types used for filter in `listaccounthistory`

#### Why is this needed:
It's needed to not look at source code to search them

Jouzo:

/kind feature

Adds a RPC method to list all available custom transaction types

Fixes #806

#### Additional comments?:

This RPC method is needed to use txtype filter in `listburnhistory`, `listaccounthistory` and `accounthistorycount`

### DIFF
--- a/src/masternodes/rpc_accounts.cpp
+++ b/src/masternodes/rpc_accounts.cpp
@@ -1018,6 +1018,9 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
             if (str.size() == 1) {
                 txType = CustomTxCodeToType(str[0]);
             }
+            if (txType == CustomTxType::None) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid tx type (" + str + ")");
+            }
         }
         if (!optionsObj["limit"].isNull()) {
             limit = (uint32_t) optionsObj["limit"].get_int64();
@@ -1267,6 +1270,8 @@ UniValue listburnhistory(const JSONRPCRequest& request) {
                 // Will search for type ::None if txtype not found.
                 txType = CustomTxCodeToType(str[0]);
                 txTypeSearch = true;
+            } else {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid tx type (" + str + ")");
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/masternodes/rpc_accounts.cpp
```diff
@@ -1802,6 +1802,30 @@ UniValue getburninfo(const JSONRPCRequest& request) {
...
+                       HelpExampleCli("listcustomtxtypes", "")
+                       + HelpExampleRpc("listcustomtxtypes", "")
+               },
+    }.Check(request);
+
+    UniValue typeObj(UniValue::VOBJ);
+    for (auto i = 0; i < std::numeric_limits<uint8_t>::max(); i++) {
+        auto type = CustomTxCodeToType(i);
+        if (type != CustomTxType::None && type != CustomTxType::Reject) {
+            typeObj.pushKV(ToString(type), std::string(1, i));
```
bvbfan, line 1831: Can we reverse because it will be pretty, `{one letter : description}`

bvbfan (changes requested):
Sanitize changes can be revert. If the requested tx type is none, we just not filter at all

Jouzo:
I added input sanitation as without it, you can pass any non valid value and get full result. Seems error prone. If one wants full result, he should call it without optional txtype parameters.

bvbfan (approved):

src/masternodes/rpc_accounts.cpp
```diff
@@ -1802,6 +1810,30 @@ UniValue getburninfo(const JSONRPCRequest& request) {
...
+               RPCResult{
+                       "{\"1\": \"ICXCreateOrder\", \"2\": \"ICXMakeOffer\", ...}     (object) List of custom transaction types { [single letter representation]: custom transaction type name}\n"
+               },
+               RPCExamples{
+                       HelpExampleCli("listcustomtxtypes", "")
+                       + HelpExampleRpc("listcustomtxtypes", "")
+               },
+    }.Check(request);
+
+    UniValue typeObj(UniValue::VOBJ);
```
prasannavl, line 1827: list* -- Are we not converging to use arrays for list in this RPC version?

prasannavl: But the output makes sense to be in an array. Wonder if we should just call it `getdefitxtypes`?

Jouzo: I wondered as well. It made sense to be called list and I initially did it with array but it's much more natural as an object. 
`getdefitxtypes` sounds good and is consistent with list* being only arrays

Jouzo: `getdftxtypes` ?

prasannavl: If I recall, the conclusion here was to use `getcustomtxcodes` ?

bvbfan (approved):

bvbfan (approved):

PR #884, opened 2021-11-08
Optimize COracle::SupportsPair search

prasannavl:
Improve search from O(n) to O(log n). `std::set` uses red black trees. Low hanging fruit on a mild hotspot function when we have many pairs. 

/kind fix


Bushstar (approved):

prasannavl:
`std::count` is already appears to be O(log N) from: https://en.cppreference.com/w/cpp/container/set/count

PR #893, opened 2021-11-10
Remove dead code

bvbfan:
/kind chore


Mixa84 (dismissed):

prasannavl:
I wonder if some of these `ToString` impls are something we should have been using for things like logging, etc better. Or do we already have sufficient logging around these messages, that this just redundant?

bvbfan:
In message data structures we use `ToString` nowhere, current ones just differs standard implementation.

prasannavl:
Waiting to merge other master PRs first. So, it's easier to rebase v3 onto updated master for easier conflicts resolution, if any.

PR #898, opened 2021-11-10
Add vault history DB and listvaulthistory RPC call

Bushstar:


src/masternodes/vaulthistory.h
```diff
@@ -55,14 +55,62 @@ struct VaultHistoryValue {
     }
 };
 
+struct VaultStateKey {
+    uint256 vaultID;
+    uint32_t blockHeight;
```
bvbfan, line 60: Could you compare to `AuctionHistoryKey` where height is first? It's faster to navigate through records when height is specified. I plan to rework account history to match auction one.

src/masternodes/vaulthistory.h
```diff
@@ -55,14 +55,62 @@ struct VaultHistoryValue {
...
+        }
+    }
+};
+
+struct VaultStateValue {
+    bool isUnderLiquidation;
+    TAmounts collaterals;
+    CCollateralLoans collateralsValues;
+    std::vector<CAuctionBatch> auctionBatches;
+    uint32_t ratio;
```
bvbfan, line 84: We can remove `ratio`, `collateralsValues` provides it by calling its method.

src/masternodes/vaulthistory.h
```diff
@@ -55,14 +55,62 @@ struct VaultHistoryValue {
...
+        }
+        else {
+            uint32_t blockHeight_ = ~blockHeight;
+            READWRITE(WrapBigEndian(blockHeight_));
+        }
+    }
+};
+
+struct VaultStateValue {
+    bool isUnderLiquidation;
```
bvbfan, line 80: `auctionBatches` not empty means in liquidation.

src/masternodes/vaulthistory.h
```diff
@@ -55,14 +55,62 @@ struct VaultHistoryValue {
...
+    TAmounts collaterals;
+    CCollateralLoans collateralsValues;
+    std::vector<CAuctionBatch> auctionBatches;
+    uint32_t ratio;
+    std::string schemeID;
```
bvbfan, line 85: Probably it's better to keep entire scheme, the name could be changed or scheme could be deleted. Then the name will not provide any significant info.

bvbfan (changes requested):

src/masternodes/vaulthistory.h
Bushstar: We want the ratio and collateralsValues stored from that point in time.

src/masternodes/vaulthistory.h
Bushstar: I did mention this, my original plan was to store the ratio of the scheme at the time but it was decided that the scheme name would suffice.

src/masternodes/vaulthistory.h
Bushstar: I do not think that height will often be specified when accessing vault records and with height first won't we have to return `true` when `isMatchVault` is `false` in `ForEachVaultHistory`, so we will end up looking through all records instead of just the ones relating to the vault specific in `getvaulthistory`?

https://github.com/DeFiCh/ain/blob/2dc3439a3248549762e192be8ff4ad8e0ab04728/src/masternodes/rpc_vault.cpp#L1322

src/masternodes/vaulthistory.h
Bushstar: Updated.

src/masternodes/vaulthistory.h
bvbfan: Correct, `collateralsValues.ratio()` is the same as `ratio` no?

src/masternodes/vaulthistory.h
bvbfan: I don't think there a `promise` that name will be sufficient, `CLoanScheme` look perfect fit here.

Bushstar: I agree, CLoanScheme would be better here. Will get back on this one.

src/masternodes/vaulthistory.h
Bushstar: I understand now, I misunderstood. This does not work, the state displayed by getvaulthistory is not really the actual state, it is a composite of state from the start and end of ProcessLoanEvents. The ratio shown is from before liquidation, where CCollateralLoans is stored after liquidation where the collateral and loans have been moved out of CCollateralLoans and into auction batches.

bvbfan (approved):

PR #907, opened 2021-11-16
Prevent loanpayback amount which sets interestPerBlock to zero

Mixa84:

/kind fix

When loan is payed back, first the whole interest is payed and then what is left is subtracted from base. If most of the loan is payed and small amount is left, the interest per block will be smaller than 1 satoshi which will set interestPerBlock to 0. This PR will prevent paying back the amount which leads to 0 interest. So either whole or less amount could be payed back.


src/masternodes/mn_checks.cpp
```diff
@@ -2663,6 +2663,15 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
             if (!res)
                 return res;
 
+            if (static_cast<int>(height) >= consensus.GreatWorldHeight)
```
bvbfan, line 2666: should it be 
`if (static_cast<int>(height) >= consensus.GreatWorldHeight && subLoan < it->second)`
it means owner can repay entire loan on interestPerBlock == 0?

PR #929, opened 2021-11-24
Migrate account history db records

### LINKED ISSUE #812: Rebuild history db index
History db index can be rebuild to <height, owner, txn>

#### Why is this needed:
* faster lookup
* no full reindex
* one time upgrade
* clean and reliable code

bvbfan:
/kind feature

Fixes #812 

It improves `listaccounthistory mine` performance, as well as history reverting maintainability.


Bushstar (approved):

Bushstar (approved):

src/masternodes/rpc_accounts.cpp
```diff
@@ -1074,108 +1074,82 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
             }
             --count;
         }
 
-        if (!noRewards && count && lastHeight > workingHeight) {
-            onPoolRewards(view, key.owner, workingHeight, lastHeight,
+        return count != 0;
+    };
+
+    paccountHistoryDB->ForEachAccountHistory(shouldContinueToNextAccountHistory, {~0u, {}, ~0u});
```
surangap, line 1125: This is equivalent to a full scan as I can see. Do not make use of the index properly. earlier it was like
```
AccountHistoryKey startKey{account, maxBlockHeight, std::numeric_limits<uint32_t>::max()};
paccountHistoryDB->ForEachAccountHistory(shouldContinueToNextAccountHistory, startKey);
```
so the iteration starts from a given `startKey`. now it starts from one end, and iterates, filters within the lambda.
I think we could pass the starting key?🤔

Also, this is important for `Whale` because it depends on the old index `<owner, height,  txn>` to get max performance by passing the `owner`. Now the index has been changed to `<height, owner, txn>` it has to be taken into account @fuxingloh @canonbrother @ivan-zynesis

fuxingloh: Hmmm, we plan to expose this endpoint for `Whale` for Light Wallet and Scan users.

However, if this greatly benefits the Full Node users, I can scrap that endpoint for now. 🤔

prasannavl:
Moving to 3.0 since these indexes are utilised highly by the SDKs so might need larger window at the least.

PR #937, opened 2021-11-26
Adds possibility to use "*" to auto-select address in placeauctionbid and paybackloan

### LINKED ISSUE #827: Add possibility to auto-select account from wallet in auctionbid and loanpayback

Currently, an address as to be passed in to bid on an auction or payback a loan.
This would improve the consistency between different RPCs.
Other RPC calls, such as `addpoolliquidity` and `sendtokenstoaddress` give the possibility to pass "*" or "{}" to auto-select an account from wallet.

It would also improve the UX. 
Right now, a user has to calculate the right amount of token and send it to an address in order to be able to bid on an auction or payback a loan.

#### Why is this needed:

Consistency between RPCs
Better UX

Jouzo:

/kind feature

Adds possibility to use "*" to auto-select address in placeauctionbid and paybackloan.
Will select the address with the highest amount of tokens.

Fixes #827


src/masternodes/rpc_loan.cpp
```diff
@@ -1169,16 +1169,41 @@ UniValue paybackloan(const JSONRPCRequest& request) {
...
-        loanPayback.from = DecodeScript(metaObj["from"].getValStr());
-    else
-        throw JSONRPCError(RPC_INVALID_PARAMETER,"Invalid parameters, argument \"from\" must not be null");
-
     if (!metaObj["amounts"].isNull())
         loanPayback.amounts = DecodeAmounts(pwallet->chain(), metaObj["amounts"], "");
     else
         throw JSONRPCError(RPC_INVALID_PARAMETER,"Invalid parameters, argument \"amounts\" must not be null");
 
+    if (!metaObj["from"].isNull()) {
```
bvbfan, line 1177: Can you check for null first
```
if (metaObj["from"].isNull()) {
    throw JSONRPCError(RPC_INVALID_PARAMETER,"Invalid parameters, argument \"from\" must not be null");
}
```
It looks complex now.

bvbfan (approved):

bvbfan (dismissed):

bvbfan (approved):

PR #897, opened 2021-11-10
Fix list rpc consistency

hidiego:
#### What kind of PR is this?:

/kind fix

#### What this PR does / why we need it:
Add additional parameter `jsonformat ("list" | "object")` to list* RPCs:
- `jsonformat = "list" -> [ {obj}, {obj} ]` 
- `jsonformat = "object" -> { key: {obj}, key: {obj} }`

PR #940, opened 2021-12-01
Fix list* RPC consistency with defid flag

Jouzo:

/kind feature

- Adds a flag to defid to format list RPC as object or array.
- Keeps backward compatibility by omitting flag
- lighter than #897

- Example output :
  
```
cat defi.conf
rpcjsonformat=array

./src/defi-cli -testnet listmasternodes '{"limit":2}'
[
  {
    "ownerAuthAddress": "7MpAUiHtgwFm3pDAfVK7a64vsrz9yFsHjK",
    "operatorAuthAddress": "7MP5V6tAxrL466a4wqL5xbUE3DQnJrvpwf",
    "rewardAddress": "",
    "creationHeight": 435638,
    "resignHeight": 519711,
    "resignTx": "8e8b0bd7dc6270b638dcce2a4bd1904d711d23a749431f494aa77b6f93706bfd",
    "banTx": "0000000000000000000000000000000000000000000000000000000000000000",
    "state": "RESIGNED",
    "mintedBlocks": 46,
    "ownerIsMine": false,
    "operatorIsMine": false,
    "localMasternode": false,
    "id": "41a26cfef6d73d26933e304be45104c3dc85462a396ae2986ff1b1569f462d00"
  },
  {
    "ownerAuthAddress": "7KaNzvfYxLpSiYQwFw9BYCaVoxypWZNkWW",
    "operatorAuthAddress": "79oQZ8xxcyaqsJ6xov5vhG2eVxhPiiynm2",
    "rewardAddress": "",
    "creationHeight": 248858,
    "resignHeight": 267237,
    "resignTx": "3bc251e6ba6c7991799d678707c1e68c33b2b4b8605bce68911dbd2d5b69f1b8",
    "banTx": "0000000000000000000000000000000000000000000000000000000000000000",
    "state": "RESIGNED",
    "mintedBlocks": 0,
    "ownerIsMine": false,
    "operatorIsMine": false,
    "localMasternode": false,
    "id": "b998589d8b3333ca6c65c410a298d7a5751e547cdfdcdae20e6808459ff7ac00"
  }
]
```

```
cat defi.conf
rpcjsonformat=object

{
  "0": {
    "token": "DFI",
    "tokenId": "897dcc7567cd883d9e77e06ee786856db3708c27b75da936e55072fc5300ba89",
    "factor": 1.00000000,
    "fixedIntervalPriceId": "DFI/USD",
    "activateAfterBlock": 686335
  },
  "1": {
    "token": "BTC",
    "tokenId": "7705ba258dc7c2a7fb8a8e55d65726d83fbeadcabd9b857f29527dbe4da46f99",
    "factor": 1.00000000,
    "fixedIntervalPriceId": "BTC/USD",
    "activateAfterBlock": 686308
  },
  "2": {
    "token": "ETH",
    "tokenId": "356c9fce2d11579eb4f724fc410090a8fb740dea1bb7f87a01c46c2f466ec0e7",
    "factor": 0.70000000,
    "fixedIntervalPriceId": "ETH/USD",
    "activateAfterBlock": 686308
  }
}
```

#### Discussion

- Can add a flag or header to override defid flag if needed.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -585,6 +585,7 @@ void SetupServerArgs()
     gArgs.AddArg("-rpcworkqueue=<n>", strprintf("Set the depth of the work queue to service RPC calls (default: %d)", DEFAULT_HTTP_WORKQUEUE), ArgsManager::ALLOW_ANY | ArgsManager::DEBUG_ONLY, OptionsCategory::RPC);
     gArgs.AddArg("-server", "Accept command line and JSON-RPC commands", ArgsManager::ALLOW_ANY, OptionsCategory::RPC);
     gArgs.AddArg("-rpcallowcors=<host>", "Allow CORS requests from the given host origin. Include scheme and port (eg: -rpcallowcors=http://127.0.0.1:5000)", ArgsManager::ALLOW_ANY, OptionsCategory::RPC);
+    gArgs.AddArg("-rpcjsonformat=<format>", "Format list as specificed json format. (Options are 'object' and 'array')", ArgsManager::ALLOW_ANY, OptionsCategory::RPC);
 
 #if HAVE_DECL_DAEMON
     gArgs.AddArg("-daemon", "Run in the background as a daemon and accept commands", ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
--- a/src/rpc/server.cpp
+++ b/src/rpc/server.cpp
@@ -437,7 +437,7 @@ UniValue CRPCTable::execute(const JSONRPCRequest &request) const
         UniValue result;
         for (const auto& command : it->second) {
             if (ExecuteCommand(*command, request, result, &command == &it->second.back())) {
-                return result;
+                return FormatResult(request, result);
             }
         }
     }
--- a/src/rpc/util.cpp
+++ b/src/rpc/util.cpp
@@ -734,3 +734,32 @@ std::vector<CScript> EvalDescriptorStringOrObject(const UniValue& scanobject, Fl
     }
     return ret;
 }
+
+UniValue FormatListResult(const UniValue& result) {
+    if (result.isObject() && gArgs.GetArg("-rpcjsonformat", "") == "array") {
+        UniValue ret(UniValue::VARR);
+        for (const auto& key : result.getKeys()) {
+            auto elem = result[key];
+            elem.pushKV("id", key);
+            ret.push_back(elem);
+        }
+        return ret;
+    } else if (result.isArray() && gArgs.GetArg("-rpcjsonformat", "") == "object") {
+        UniValue object{UniValue::VOBJ};
+        std::vector<UniValue> values{result.getValues()};
+
+        int listSize = result.size();
+        for (int i = 0; i < listSize; i++) {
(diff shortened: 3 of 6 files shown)


Jouzo:
## Notes regarding list* RPC consistency :

Summary of conversation on PR https://github.com/DeFiCh/ain/pull/897:
- Only four list RPC are sending their result as object : `listmasternodes`, `listpoolpairs`, `listpoolshares` and `listtokens`.
The consistency issue arise while we were developing FC list* RPC but all of these were brought to par with the rest and correctly send their result as array.
- These four RPCs are scheduled to be rewritten in RPC v2 to add filtering and pagination.
- No known use cases nor demand from users for the possibilty to have `jsonformat=object`.

In that case, I don't see the value in either https://github.com/DeFiCh/ain/pull/940 nor https://github.com/DeFiCh/ain/pull/897.

If we still want to address it, we can go with either a global flag or a local flag passed done the RPC pipeline:

- Global flag as in PR https://github.com/DeFiCh/ain/pull/940 is absolute. Either all list as array or all list as object. 
 Pro: Set once and forget about it.
 Cons: Cannot request format per RPC (Is that something we should allow ? I don't see a use case for this)

- Local flag added to each command : 
 Pro: more granularity
 Cons: Complexify the pipeline


If we wait for API V2:
Regarding the versioning of the API, should we use a dedicated endpoint to dispatch to new RPCs as it is already done for `-rpcwallet` flag. 

https://github.com/DeFiCh/ain/blob/dc769f32b7bb23dd6f305bcb9d918b98597b7ce5/src/defi-cli.cpp#L352

```
// check if we should use a special wallet endpoint
std::string endpoint = "/";
if (!gArgs.GetArgs("-rpcwallet").empty()) {
    std::string walletName = gArgs.GetArg("-rpcwallet", "");
    char *encodedURI = evhttp_uriencode(walletName.c_str(), walletName.size(), false);
    if (encodedURI) {
        endpoint = "/wallet/"+ std::string(encodedURI);
        free(encodedURI);
    }
    else {
        throw CConnectionFailed("uri-encode failed");
    }
}
```

https://github.com/DeFiCh/ain/blob/dc769f32b7bb23dd6f305bcb9d918b98597b7ce5/src/httprpc.cpp#L275
```
if (g_wallet_init_interface.HasWalletSupport()) {
    RegisterHTTPHandler("/wallet/", false, HTTPReq_JSONRPC);
}
```

For the versioning could be :

```
// check if we should use a special wallet endpoint
std::string endpoint = "/";
if (!gArgs.GetArgs("-v2").empty()) {
        endpoint = "/v2/";
    } 
}
```

This would then be called either :

`defi-cli -v2 listtokens '{"limit":1}'`
`curl --user username --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "listtokens", "params": [{"limit":1}] }' -H 'content-type: text/plain;' http://127.0.0.1:18554/v2`

This seems clearer than using a dedicated header for the JSONRPC such as `X-defichain-v2` and a `-v2` flag on the CLI.

Jouzo:
Closing this as it will be addressed in RPC v2

PR #956, opened 2021-12-07
Better define composite swap behaviour

Bushstar:


bvbfan (dismissed):

src/masternodes/mn_checks.cpp
```diff
@@ -1175,6 +1175,12 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
             return Res::Err("tx must have at least one input from account owner");
         }
 
+        if (height >= static_cast<uint32_t>(Params().GetConsensus().FortCanningMuseumHeight)) {
+            if (obj.poolIDs.size() == 1 || obj.poolIDs.size() > 3) {
+                return Res::Err(strprintf("Unexpected number of pool IDs, only 2 or 3 allowed, %d provided", obj.poolIDs.size()));
+            }
```
fuxingloh, line 1181: What do you think about enforcing singular `PoolID` (making it length of 1 or 2 or 3 allowed) for those without hops? This makes it more deterministic/explicit, we don't need to do any lookup by Symbol to find the pair. 

Also, if we have multiple Pairs of the same symbol pair, we can enforce which pair to take. Although that's not the case now.

fuxingloh (approved):

surangap (approved):

PR #970, opened 2021-12-09
Introduce new CRPCStats class to save key RPC statistics

Jouzo:

/kind feature

Save RPC latency and payload size on each call.

Introduces 2 RPCs:

- getrpcstats
- listrpcsstats

`getrpcstats` :

```
defi-cli -regtest getrpcstats getblockcount
{
  "name": "getblockcount",
  "lastUsedTime": 1639041504,
  "latency": {
    "max": 1,
    "min": 0,
    "avg": 0
  },
  "payload": {
    "max": 37,
    "min": 37,
    "avg": 37
  },
  "count": 3,
  "history": [
    {
      "timestamp": 1639041311,
      "latency": 0,
      "payload": 37
    },
    {
      "timestamp": 1639041348,
      "latency": 1,
      "payload": 37
    },
    {
      "timestamp": 1639041504,
      "latency": 1,
      "payload": 37
    }
  ]
}
```

```
defi-cli -regtest listrpcstats
[
  {
    "name": "getblockcount",
    "lastUsedTime": 1639041504,
    "latency": {
      "max": 1,
      "min": 0,
      "avg": 0
    },
    "payload": {
      "max": 37,
      "min": 37,
      "avg": 37
    },
    "count": 3,
    "history": [
      {
        "timestamp": 1639041311,
        "latency": 0,
        "payload": 37
      },
      {
        "timestamp": 1639041348,
        "latency": 1,
        "payload": 37
      },
      {
        "timestamp": 1639041504,
        "latency": 1,
        "payload": 37
      }
    ]
  },
  {
    "name": "getrpcstats",
    "lastUsedTime": 1639041506,
    "latency": {
      "max": 0,
      "min": 0,
      "avg": 0
    },
    "payload": {
      "max": 329,
      "min": 167,
      "avg": 248
    },
    "count": 2,
    "history": [
      {
        "timestamp": 1639041487,
        "latency": 0,
        "payload": 167
      },
      {
        "timestamp": 1639041506,
        "latency": 0,
        "payload": 329
      }
    ]
  },
  {
    "name": "listmasternodes",
    "lastUsedTime": 1639041356,
    "latency": {
      "max": 1,
      "min": 1,
      "avg": 1
    },
    "payload": {
      "max": 4255,
      "min": 4255,
      "avg": 4255
    },
    "count": 1,
    "history": [
      {
        "timestamp": 1639041356,
        "latency": 1,
        "payload": 4255
      }
    ]
  },
  {
    "name": "listrpcstats",
    "lastUsedTime": 1639041536,
    "latency": {
      "max": 0,
      "min": 0,
      "avg": 0
    },
    "payload": {
      "max": 608,
      "min": 608,
      "avg": 608
    },
    "count": 1,
    "history": [
      {
        "timestamp": 1639041536,
        "latency": 0,
        "payload": 608
      }
    ]
  },
  {
    "name": "listunspent",
    "lastUsedTime": 1639041363,
    "latency": {
      "max": 3400,
      "min": 3400,
      "avg": 3400
    },
    "payload": {
      "max": 5313266,
      "min": 5313266,
      "avg": 5313266
    },
    "count": 1,
    "history": [
      {
        "timestamp": 1639041363,
        "latency": 3400,
        "payload": 5313266
      }
    ]
  }
]
```

### DIFF
--- a/src/rpc/stats.cpp
+++ b/src/rpc/stats.cpp
@@ -1,0 +1,203 @@
+#include <rpc/stats.h>
+
+#include <rpc/server.h>
+#include <rpc/util.h>
+
+UniValue RPCStats::toJSON() {
+    UniValue stats(UniValue::VOBJ),
+             latencyObj(UniValue::VOBJ),
+             payloadObj(UniValue::VOBJ),
+             historyArr(UniValue::VARR);
+
+    latencyObj.pushKV("min", latency.min);
+    latencyObj.pushKV("avg", latency.avg);
+    latencyObj.pushKV("max", latency.max);
+
+    payloadObj.pushKV("min", payload.min);
+    payloadObj.pushKV("avg", payload.avg);
+    payloadObj.pushKV("max", payload.max);
+
+    for (auto const &entry : history) {
+        UniValue historyObj(UniValue::VOBJ);
+        historyObj.pushKV("timestamp", entry.timestamp);
+        historyObj.pushKV("latency", entry.latency);
+        historyObj.pushKV("payload", entry.payload);
+        historyArr.push_back(historyObj);
+    }
+
+    stats.pushKV("name", name);
+    stats.pushKV("count", count);
+    stats.pushKV("lastUsedTime", lastUsedTime);
+    stats.pushKV("latency", latencyObj);
+    stats.pushKV("payload", payloadObj);
+    stats.pushKV("history", historyArr);
+    return stats;
+}
+
+RPCStats RPCStats::fromJSON(UniValue json) {
+    RPCStats stats;
+
+    stats.name = json["name"].get_str();
+    stats.lastUsedTime = json["lastUsedTime"].get_int64();
+    stats.count = json["count"].get_int64();
+    if (!json["latency"].isNull()) {
+        auto latencyObj  = json["latency"].get_obj();
+        stats.latency = {
+            latencyObj["min"].get_int64(),
+            latencyObj["avg"].get_int64(),
+            latencyObj["max"].get_int64()
+        };
+    }
+
+    if (!json["payload"].isNull()) {
+        auto payloadObj  = json["payload"].get_obj();
+        stats.payload = {
+            payloadObj["min"].get_int64(),
+            payloadObj["avg"].get_int64(),
+            payloadObj["max"].get_int64()
+        };
+    }
+
+    if (!json["history"].isNull()) {
+        auto historyArr = json["history"].get_array();
+        for (const auto &entry : historyArr.getValues()) {
+            auto historyObj = entry.get_obj();
+            StatHistoryEntry historyEntry {
+                historyObj["timestamp"].get_int64(),
+                historyObj["latency"].get_int64(),
+                historyObj["payload"].get_int64()
+            };
+            stats.history.push_back(historyEntry);
+        }
+    }
+    return stats;
+}
+
+void CRPCStats::add(const std::string& name, const int64_t latency, const int64_t payload)
+{
+    auto stats = CRPCStats::get(name);
+    if (stats) {
+        stats->count++;
+        stats->lastUsedTime = GetSystemTimeInSeconds();
+        stats->latency = {
+            std::min(latency, stats->latency.min),
+            stats->latency.avg + (latency - stats->latency.avg) / stats->count,
+            std::max(latency, stats->latency.max)
+        };
+        stats->payload = {
+            std::min(payload, stats->payload.min),
+            stats->payload.avg + (payload - stats->payload.avg) / stats->count,
+            std::max(payload, stats->payload.max)
+        };
+    } else {
+        stats = { name, latency, payload };
+    }
+    stats->history.push_back({ stats->lastUsedTime, latency, payload });
+
+    CLockFreeGuard lock(lock_stats);
+    map[name] = *stats;
+}
+
+UniValue CRPCStats::toJSON() {
+    CLockFreeGuard lock(lock_stats);
+
+    UniValue ret(UniValue::VARR);
+    for (auto &[_, stats] : map) {
+        ret.push_back(stats.toJSON());
+    }
+    return ret;
+}
+
+static UniValue getrpcstats(const JSONRPCRequest& request)
+{
+    RPCHelpMan{"getrpcstats",
+        "\nGet RPC stats for selected command.\n",
+        {
+            {"command", RPCArg::Type::STR, RPCArg::Optional::NO, "The command to get stats for."}
+        },
+        RPCResult{
+            " {\n"
+            "  \"name\":               (string) The RPC command name.\n"
+            "  \"latency\":            (json object) Min, max and average latency.\n"
+            "  \"payload\":            (json object) Min, max and average payload size in bytes.\n"
+            "  \"count\":              (numeric) The number of times this command as been used.\n"
+            "  \"lastUsedTime\":       (numeric) Last used time as timestamp.\n"
+            "  \"history\":            (json array) History of last 5 RPC calls.\n"
+            "  [\n"
+            "       {\n"
+            "           \"timestamp\": (numeric)\n"
+            "           \"latency\":   (numeric)\n"
+            "           \"payload\":   (numeric)\n"
+            "       }\n"
+            "  ]\n"
+            "}"
+        },
+        RPCExamples{
+            HelpExampleCli("getrpcstats", "getblockcount") +
+            HelpExampleRpc("getrpcstats", "\"getblockcount\"")
+        },
+    }.Check(request);
+
+    if (!statsRPC.isActive()) {
+        throw JSONRPCError(RPC_INVALID_REQUEST, "Rpcstats is desactivated.");
+    }
+
+    auto command = request.params[0].get_str();
+    auto stats = statsRPC.get(command);
+    if (!stats) {
+        throw JSONRPCError(RPC_INVALID_PARAMS, "No stats for this command.");
+    }
+    return stats->toJSON();
+}
+
+static UniValue listrpcstats(const JSONRPCRequest& request)
+{
+    RPCHelpMan{"listrpcstats",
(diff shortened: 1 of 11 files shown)


src/rpc/stats.cpp
```diff
@@ -0,0 +1,124 @@
...
+            avg_payload = payload,
+            max_payload = payload,
+            count       = 1;
+
+    auto stats = statsRPC.get(name);
+    if (!stats.empty()) {
+        count = stats["count"].get_int() + 1;
+
+        latencyObj = stats["latency"].get_obj();
+        if (latency < latencyObj["min"].get_int()) min_latency = latency;
```
bvbfan, line 26: Consider use `std::min`
```
min_latency = std::min(min_latency, latencyObj["min"].get_int64());
```

src/rpc/stats.cpp
```diff
@@ -0,0 +1,124 @@
...
+        if (latency > latencyObj["max"].get_int()) max_latency = latency;;
```
bvbfan, line 27: Use `std::max`
```
max_latency = std::min(max_latency, latencyObj["max"].get_int64());
```

src/rpc/stats.cpp
```diff
@@ -0,0 +1,124 @@
...
+        for (; i < historyArr.size(); i++) {
+            history.push_back(historyArr[i]);
+        }
+    }
+
+    data.pushKV("name", name);
+
+    latencyObj.pushKV("min", latency);
+    latencyObj.pushKV("avg", latency);
+    latencyObj.pushKV("max", latency);
```
bvbfan, line 49: use `min/max/avg_latency`

src/rpc/stats.cpp
```diff
@@ -0,0 +1,124 @@
...
+    data.pushKV("latency", latencyObj);
+
+    payloadObj.pushKV("min", payload);
+    payloadObj.pushKV("avg", payload);
+    payloadObj.pushKV("max", payload);
```
bvbfan, line 54: Use `min/max/avg_payload`

src/rpc/stats.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        latencyObj = stats["latency"].get_obj();
+        if (latency < latencyObj["min"].get_int()) min_latency = latency;
+        if (latency > latencyObj["max"].get_int()) max_latency = latency;;
+        avg_latency = latencyObj["avg"].get_int() + (latency - latencyObj["avg"].get_int()) / count;
+
+        payloadObj = stats["payload"].get_obj();
+        if (payload < payloadObj["min"].get_int()) min_payload = payload;;
+        if (payload > payloadObj["max"].get_int()) max_payload = payload;;
+        avg_payload = payloadObj["avg"].get_int() + (payload - payloadObj["avg"].get_int()) / count;
```
bvbfan, line 33: Same as latency calculation

src/rpc/stats.cpp
```diff
@@ -0,0 +1,124 @@
...
+            count       = 1;
+
+    auto stats = statsRPC.get(name);
+    if (!stats.empty()) {
+        count = stats["count"].get_int() + 1;
+
+        latencyObj = stats["latency"].get_obj();
+        if (latency < latencyObj["min"].get_int()) min_latency = latency;
+        if (latency > latencyObj["max"].get_int()) max_latency = latency;;
+        avg_latency = latencyObj["avg"].get_int() + (latency - latencyObj["avg"].get_int()) / count;
```
bvbfan, line 28: Why not `avg_latency = (latencyObj["avg"].get_int64() + latency) / count` ?

bvbfan (changes requested):

src/rpc/stats.cpp
Jouzo: Would not work.
Let's take latency values 2 3 10:
first avg: 2 / 1 == 2
second: (2 + 3) / 2 == 2.5 
third: (2.5 + 10) / 3 == 4.16~ (should be 5)

See https://math.stackexchange.com/questions/22348/how-to-add-and-subtract-values-from-an-average for formula explanation

bvbfan: Alright then can we do just, (old value + new value) / 2, i still think that formula is bit complicated than it should. If it still needed i'm fine.

Jouzo: would not work either
taking third from above :
(2.5 + 10) / 2 == 7.5

What could work is 
(old value * (count - 1) + new value) / count 
(2.5 * 2 + 10) / 3 == 5

bvbfan: Ok

bvbfan (approved):

src/rpc/stats.cpp
prasannavl: Running avg. looks good. Don't store the cache as JSON (Univalue), however. Just use raw objects. It's pretty pointless to do this per call. Just serialize the object on output.

prasannavl:
Use something on the lines of 

```
  RPCStats {
    name: string,
    lastUsedTime: int64,
    latency: MinMaxStatEntry { min, max, avg }, 
    payload: MinMaxStatEntry,
    count: uint64,
    history: std::vector<StatHistoryEntry>
  }
  
  std::vector<RPCStats>
  ```
  
  for the internal memory repr and serialize / deserliaze in one-go. 
  
  Currently, each UniValue is:  
  
  ```
      UniValue::VType typ;
    std::string val;                       // numbers are stored as C++ strings
    std::vector<std::string> keys;
    std::vector<UniValue> values;
  ```

File: src/rpc/stats.h [hunk 7]
```diff
@@ -54,30 +54,36 @@ struct RPCStats {
...
 {
 private:
+    std::atomic_bool lock_stats{false};
     std::map<std::string, RPCStats> map;
-public:
     bool active{DEFAULT_RPC_STATS};
 
+public:
+    bool isActive() {
+        return active;
```
prasannavl, line 63: This needs to be guarded too. But can be granular. Use another atomic bool, and return. Same way for set.

src/rpc/stats.h
```diff
@@ -54,30 +54,36 @@ struct RPCStats {
...
+    }
+    void setActive(bool isActive) {
+        active = isActive;
```
prasannavl, line 66: Read above.

return active.load(), if active is atomic bool.

src/rpc/stats.h
prasannavl: `return active.load();`

bvbfan (approved):

bvbfan (approved):

src/rpc/stats.cpp
```diff
@@ -0,0 +1,203 @@
...
+            stats.history.push_back(historyEntry);
+        }
+    }
+    return stats;
+}
+
+void CRPCStats::add(const std::string& name, const int64_t latency, const int64_t payload)
+{
+    auto stats = CRPCStats::get(name);
+    if (stats) {
```
prasannavl, line 79: No locks held during this point. It's mutated freely. And if a reference of the same stats is held from another method (see later), this can end up seeing partial writes resulting in races.

src/rpc/stats.cpp
```diff
@@ -0,0 +1,203 @@
...
+            HelpExampleRpc("getrpcstats", "\"getblockcount\"")
+        },
+    }.Check(request);
+
+    if (!statsRPC.isActive()) {
+        throw JSONRPCError(RPC_INVALID_REQUEST, "Rpcstats is desactivated.");
+    }
+
+    auto command = request.params[0].get_str();
+    auto stats = statsRPC.get(command);
```
prasannavl, line 146: Example of another method that keeps the reference. (see above).

src/rpc/stats.cpp
bvbfan: Nope, there is no shared resource here. `CRPCStats::get` returns a copy in the stack which is not shared between threads.

src/rpc/stats.cpp
bvbfan: Nope, same as above.

bvbfan:
@Jouzo  rebase to master, revert your last changes, they are incorrect.

src/rpc/stats.cpp
prasannavl: Reviewing again, the above statement is absolutely correct. They are all simple value types. Resolving this.

prasannavl: Reviewing again, the above statement is absolutely correct. They are all simple value types. Resolving this.

src/rpc/stats.cpp
prasannavl: Agreed. Same as above.

prasannavl: Agreed. Same as above.

prasannavl:
Based on https://github.com/DeFiCh/ain/pull/970#discussion_r876680857, the shared resource usage is well covered as it is right now. 

All the values are used are simple value types that are copied over on the stack as stated in the above. Let's make sure there are no usage patterns that cause dead lock and we can proceed to merge.

prasannavl:
Based on https://github.com/DeFiCh/ain/pull/970#discussion_r876680857, the shared resource usage is well covered as it is right now. 

All the values are used are simple value types that are copied over on the stack as stated in the above. Let's make sure there are no usage patterns that cause dead lock and we can proceed to merge.

src/rpc/stats.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    CLockFreeGuard lock(lock_stats);
+    map[name] = *stats;
+}
+
+UniValue CRPCStats::toJSON() {
+    CLockFreeGuard lock(lock_stats);
+
+    UniValue ret(UniValue::VARR);
+    for (auto &[_, stats] : map) {
```
prasannavl, line 105: Might be better to just copy the map, and do the toJSON conversion outside of the lock.

prasannavl:
Build failure was due to v3 branch. Should be good now.

prasannavl (approved):

PR #974, opened 2021-12-10
Reuse interest per block functionality

bvbfan:
/kind chore


Mixa84:
I like it as a template.

Lets just try for EraseInterest as CAmount{1} to set a minimum interest of 1 sat and see if it works in all cases.

bvbfan:
That's not a hardfork, probably superset by #991

bvbfan:
Fixed in #991

PR #976, opened 2021-12-10
Calculate rewards according to balance change

bvbfan:
/kind refactor

### DIFF
--- a/src/masternodes/masternodes.cpp
+++ b/src/masternodes/masternodes.cpp
@@ -826,11 +826,49 @@ bool CCustomCSView::CanSpend(const uint256 & txId, int height) const
     return !pair || pair->second.destructionTx != uint256{} || pair->second.IsPoolShare();
 }
 
-bool CCustomCSView::CalculateOwnerRewards(CScript const & owner, uint32_t targetHeight)
+static void UpdateRewardsTransparent(CCustomCSView & view, CScript const & owner, uint32_t height)
+{
+    CCustomCSView mnview(view);
+    mnview.CalculateOwnerRewards(owner, height);
+    mnview.Flush();
+}
+
+CTokenAmount CCustomCSView::GetBalancePlusRewards(CScript const & owner, DCT_ID tokenID, uint32_t height)
+{
+    CCustomCSView view(*this);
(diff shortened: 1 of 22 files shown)


Bushstar (approved):

src/masternodes/masternodes.cpp
```diff
@@ -871,11 +871,42 @@ bool CCustomCSView::CanSpend(const uint256 & txId, int height) const
...
+{
+    UpdateRewardsTransparent(*this, owner, height);
+    return AddBalances(owner, balances);
+}
+
+Res CCustomCSView::SubBalances(CScript const & owner, CBalances const & balances, uint32_t height)
+{
+    UpdateRewardsTransparent(*this, owner, height);
+    return SubBalances(owner, balances);
+}
```
prasannavl, line 903: Same for all these - add `AndFlush` - so it's clear for the callers that this will involve IO?

bvbfan: There is no need of `AndFlush` since flush is on child cache, if it's not done changes are not going to be part of parent view.

prasannavl: Changes aren't. But it still call the DB and makes the IO call.

prasannavl: I think child cache still ends up in the DB to flush be it a part of parent view or not, and trigger IO ops due to the Flush.

I don't see any implementations of `::Flush` that do not hit the DB?

prasannavl: https://github.com/defich/ain/blob/pr/976/src/dbwrapper.cpp/#L190-L191 - `WriteBatch` always hits level db `Write`. And leveldb write ends up locking the writer regardless of if there's changes are not. 

The only consolation is it may not have `fSync` - so it may not be as terrible unless other parallel IO is queued up, but regardless, we can easily just avoid the whole path.

prasannavl: Ok, see what you mean. https://github.com/defich/ain/blob/master/src/coins.cpp#L147 - Ok, look good!

bvbfan:
Blocked by #1095

PR #988, opened 2021-12-14
Collateral increases in value and should be able to be withdrawn

Jouzo:

/kind fix

When a collateral increases a value, it should be allowed to be withdrawn even if ratio with DFI ratio is still below 50%.


test/functional/feature_loan_vault.py
```diff
@@ -455,5 +451,42 @@ def run_test(self):
...
+        # BTC doubles in price
+        oracle1_prices = [{"currency": "USD", "tokenAmount": "1@DFI"}, {"currency": "USD", "tokenAmount": "1@TSLA"}, {"currency": "USD", "tokenAmount": "2@BTC"}]
+        timestamp = calendar.timegm(time.gmtime())
+        self.nodes[0].setoracledata(oracle_id1, timestamp, oracle1_prices)
+        self.nodes[0].generate(11)
+
+        # Should be able to withdraw part of BTC
+        self.nodes[0].withdrawfromvault(vaultId4, address, "0.1@BTC")
+        self.nodes[0].generate(1)
+        vault = self.nodes[0].getvault(vaultId4)
```
Jouzo, line 489:
```suggestion
```

src/masternodes/mn_checks.cpp
```diff
@@ -2505,7 +2505,7 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
                         if (col.nTokenId == DCT_ID{0})
                             totalDFI += col.nValue;
 
-                    if (totalDFI < collateralsLoans.val->totalCollaterals / 2)
+                    if (totalDFI < collateralsLoans.val->totalCollaterals / 2 && obj.amount.nTokenId == DCT_ID{0})
```
bvbfan, line 2508: It's a hardfork, guard against GW at line 2503
```
if (obj.amount.nTokenId == DCT_ID{0} || static_cast<int>(height) < consensus.GreatWorldHeight)
```

bvbfan (approved):

bvbfan (dismissed):

prasannavl (approved):

src/masternodes/mn_checks.cpp
```diff
@@ -2533,8 +2533,13 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
                             totalDFI += col.nValue;
 
-                    if (totalDFI < collateralsLoans.val->totalCollaterals / 2)
-                        return Res::Err("At least 50%% of the vault must be in DFI");
+                    if (static_cast<int>(height) < consensus.FortCanningHillHeight) {
+                        if (totalDFI < collateralsLoans.val->totalCollaterals / 2)
+                            return Res::Err("At least 50%% of the vault must be in DFI");
+                    } else {
+                        if (MultiplyAmounts(totalDFI, 100) < MultiplyAmounts(collateralsLoans.val->totalLoans, scheme->ratio) / 2)
+                            return Res::Err("At least 50%% of the vault must be in DFI");
```
uzyn, line 2541:
```suggestion
                            return Res::Err("At least 50% of the collateral must be in DFI");
```

src/masternodes/mn_checks.cpp
```diff
@@ -2639,8 +2644,13 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
                 if (col.nTokenId == DCT_ID{0})
                     totalDFI += col.nValue;
 
-            if (totalDFI < collateralsLoans.val->totalCollaterals / 2)
-                return Res::Err("At least 50%% of the vault must be in DFI when taking a loan");
+            if (static_cast<int>(height) < consensus.FortCanningHillHeight) {
+                if (totalDFI < collateralsLoans.val->totalCollaterals / 2)
+                    return Res::Err("At least 50%% of the vault must be in DFI when taking a loan.");
```
uzyn, line 2649:
```suggestion
                    return Res::Err("At least 50% of the collateral must be in DFI when taking a loan.");
```

src/masternodes/mn_checks.cpp
```diff
@@ -2639,8 +2644,13 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            } else {
+                if (MultiplyAmounts(totalDFI, 100) < MultiplyAmounts(collateralsLoans.val->totalLoans, scheme->ratio) / 2)
+                    return Res::Err("At least 50%% of the vault must be in DFI when taking a loan.");
```
uzyn, line 2652:
```suggestion
                    return Res::Err("At least 50% of the collateral must be in DFI when taking a loan.");
```

uzyn (changes requested):
Minor typo. 

For future: Probably good for us to have a centralized error management instead of repeating these.

src/masternodes/mn_checks.cpp
Jouzo: Thanks, I will fix it. we just have to keep the double % otherwise it works as a conversion specifier

bvbfan (dismissed):

PR #991, opened 2021-12-17
High precision interest calculation

### LINKED ISSUE #984: Higher than COIN precision for interest rates
Related: https://github.com/DeFiCh/ain/issues/983

### Summary

Instead of just using the low 64 bits. If we can either store the whole of 256 bits, or use 128 bit name space, we can hit multiple birds with one stone, also making fractional loans fair.

### Considerations

- Using the whole 256 bits should be relatively straight-forward. Shouldn't require much additional testing, just that it wastes a lot of space. 
- Using 128 bits can be more optimal as it can address our needs well while also being a better balance, but this means, we'll have to track the 2 units of 64 bit uint, and carefully translate them back and forth to `CAmounts`.

Mixa84:

/kind feature

Improves the precision for calculating interest. It will make interest be precise on smaller loan amounts so that it doesn't get rounded to 1 sat and be unfair for small amounts of loans.

Fixes #984


src/arith_uint256.cpp
```diff
@@ -148,13 +149,18 @@ double base_uint<BITS>::getdouble() const
 template <unsigned int BITS>
 std::string base_uint<BITS>::GetHex() const
 {
-    return ArithToUint256(*this).GetHex();
+    uint256 b;
+    for(int x=0; x<WIDTH; ++x)
+        WriteLE32(b.begin() + x*4, pn[x]);
+    return b.GetHex();
```
bvbfan, line 155: That's not look correct, It should have similar code to `uint256::GetHex` here

src/arith_uint256.cpp
```diff
@@ -148,13 +149,18 @@ double base_uint<BITS>::getdouble() const
...
 }
 
 template <unsigned int BITS>
 void base_uint<BITS>::SetHex(const char* psz)
 {
-    *this = UintToArith256(uint256S(psz));
+    uint256 a(uint256S(psz));
+    for(int x=0; x<WIDTH; ++x)
+        pn[x] = ReadLE32(a.begin() + x*4);
```
bvbfan, line 163: Here it's wrong, it should copy the lowest not he highest bits, i would prefer similar to `uint256::SetHex` here. Probably it should get bytes count.

src/arith_uint256.cpp
```diff
@@ -224,6 +230,16 @@ template void base_uint<256>::SetHex(const std::string&);
...
 
+template base_uint<128>& base_uint<128>::operator<<=(unsigned int);
+template base_uint<128>& base_uint<128>::operator>>=(unsigned int);
+template base_uint<128>& base_uint<128>::operator*=(uint32_t b32);
+template base_uint<128>& base_uint<128>::operator*=(const base_uint<128>& b);
+template base_uint<128>& base_uint<128>::operator/=(const base_uint<128>& b);
+template int base_uint<128>::CompareTo(const base_uint<128>&) const;
+template bool base_uint<128>::EqualTo(uint64_t) const;
+template std::string base_uint<128>::GetHex() const;
+template std::string base_uint<128>::ToString() const;
```
bvbfan, line 241: It's redundant, the above as well. Explicit template specialization can be based on entire class not on specific function.
`template<> class base_uint<128>;`

src/arith_uint256.cpp
```diff
@@ -224,6 +230,16 @@ template void base_uint<256>::SetHex(const std::string&);
...
+
```
bvbfan, line 242: Also add `arith_uint128` as type and use it elsewhere.

src/masternodes/loan.cpp
```diff
@@ -212,14 +231,40 @@ inline float InterestPerBlockFloat(CAmount amount, CAmount tokenInterest, CAmoun
     return MultiplyAmounts(netInterest, amount) / (365.f * Params().GetConsensus().blocksPerDay());
 }
 
-CAmount TotalInterest(const CInterestRate& rate, uint32_t height)
+base_uint<128> TotalInterest(const CInterestRateV2& rate, uint32_t height)
```
bvbfan, line 234: Probably `TotalInterest` can do the ceil

bvbfan (changes requested):

src/chainparams.cpp
```diff
@@ -917,6 +921,7 @@ void CRegTestParams::UpdateActivationParametersFromArgs(const ArgsManager& args)
     UpdateHeightValidation("Eunos Paya", "-eunospayaheight", consensus.EunosPayaHeight);
     UpdateHeightValidation("Fork canning", "-fortcanningheight", consensus.FortCanningHeight);
     UpdateHeightValidation("Fork canning museum", "-fortcanningmuseumheight", consensus.FortCanningMuseumHeight);
+    UpdateHeightValidation("Fork canning new", "-FortCanningHillheight", consensus.FortCanningHillHeight);
```
Bushstar, line 924: lowercase for -fortcanninghillheight

src/init.cpp
```diff
@@ -473,6 +473,7 @@ void SetupServerArgs()
     gArgs.AddArg("-eunospayaheight", "EunosPaya fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
     gArgs.AddArg("-fortcanningheight", "Fort Canning fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
     gArgs.AddArg("-fortcanningmuseumheight", "Fort Canning Museum fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
+    gArgs.AddArg("-FortCanningHillheight", "Fort Canning New fork activation height (regtest only)", ArgsManager::ALLOW_ANY, OptionsCategory::CHAINPARAMS);
```
Bushstar, line 476: lowercase for -fortcanninghillheight here as well.

bvbfan (dismissed):

bvbfan (dismissed):

bvbfan (dismissed):

bvbfan (dismissed):

bvbfan (dismissed):

bvbfan (dismissed):

prasannavl (approved):

PR #994, opened 2021-12-20
Custom transaction version and expiration

Bushstar:


bvbfan (approved):

src/masternodes/mn_rpc.cpp
```diff
@@ -252,6 +252,14 @@ std::string ScriptToString(CScript const& script) {
...
 }
 
+int chainHeight(interfaces::Chain::Lock& locked_chain)
+{
+    LOCK(locked_chain.mutex());
+    if (auto height = locked_chain.getHeight())
+        return *height;
+    return 0;
+}
+
```
bvbfan, line 262: That's old-fashioned, let's remove it, just use `view.GetLastHeight` instead

bvbfan (changes requested):

src/masternodes/mn_rpc.cpp
Bushstar: Agreed. There's a lot of entries and on Saturday I was out of time so I restored this function. Will remove.

PR #998, opened 2021-12-23
Add ATTRIBUTES Gov var

Bushstar:


bvbfan (dismissed):

src/masternodes/mn_checks.cpp
```diff
@@ -1352,26 +1352,54 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                // Make sure ORACLE_BLOCK_INTERVAL only updates at end of interval
                 const auto diff = height % mnview.GetIntervalBlock();
                 if (diff != 0) {
                     // Store as pending change
                     storeGovVars({var, height + mnview.GetIntervalBlock() - diff});
                     continue;
                 }
+            } else if (var->GetName() == "ATTRIBUTES") {
+                // Add to existing ATTRIBUTES instead of overwriting.
+                auto govVar = mnview.GetVariable(var->GetName());
```
bvbfan, line 1370: Check for nullptr, just preventing unexpected error.
It can happen if authorized person do a mistake, it should not shutdown entire network.

src/validation.cpp
```diff
@@ -2808,8 +2808,13 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
             auto storedGovVars = cache.GetStoredVariables(static_cast<uint32_t>(pindex->nHeight));
             for (const auto& var : storedGovVars) {
                 CCustomCSView govCache(cache);
-                // Ignore any Gov variables that fail to validate, apply or be set.
-                if (var->Validate(govCache) && var->Apply(govCache, pindex->nHeight) && govCache.SetVariable(*var)) {
+                // Add to existing ATTRIBUTES instead of overwriting.
+                if (var->GetName() == "ATTRIBUTES") {
+                    auto govVar = mnview.GetVariable(var->GetName());
```
bvbfan, line 2814: Same as above, check for nullptr

bvbfan (changes requested):

src/masternodes/mn_checks.cpp
Bushstar: nullptr checked for here.

https://github.com/DeFiCh/ain/blob/0b63d4f4bd5099ef6b24689debf2fd6568d75862/src/masternodes/mn_checks.cpp#L380

src/validation.cpp
Bushstar: Will do. I do not think that should happen but let's add check to make sure.

src/masternodes/mn_checks.cpp
bvbfan: Actually it's fine. I was a bit for `mnview.GetVariable(var->GetName())` to return `nullptr`, but it's actually checked and there is "ATTRIBUTES", in some weird situation if variable somehow does not exists it could be a problem. But probably it should not happen at all.

bvbfan (approved):

prasannavl:
Just leaving this in the PR for future:

- We can likely switch the gov attributes to a trie based structure to keep it a flat, yet efficient to parse through.
- Better to add a string limit of say 256 and expand as needed to keep input bounded, to nodes from being spammed with invalid txs of large sizes.

PR #1006, opened 2021-12-28
log: Adjust masternode staking output to fix issue #648

### LINKED ISSUE #648: Change output of debug.log on Masternodeoperators
to discuss:

_Part 1:_
debug=staking:
- to much output for a longer operating in production with debugmode staking, but without, we have too less infos for a minted block and cant monitor chainsplits or rollbacks.
Proposal: Add following lines to normal Logmode without debug=staking
**2021-05-05T17:19:45Z MakeStake: kernel found
2021-05-05T17:19:45Z new proof-of-stake block found hash: 067cf6f79e103ffa735bd110847ee37ea7beec5e935c6ac7b780e6288361ca9c**
2021-05-05T17:19:45Z UpdateTip: new best=067cf6f79e103ffa735bd110847ee37ea7beec5e935c6ac7b780e6288361ca9c height=82789......
2021-05-05T17:19:45Z ThreadStaker (f2bb742370d0045359bdc3c371b382fbaa42dfce): minted a block!

atm only the last two lines are in debug.log without debug=staking

_Part 2:_
reduce Loglevel when debug=staking is disabled:

Without debug=staking, the debug.log is flooded with

2021-07-30T12:20:17Z ThreadStaker: (833......db786ce) waiting init...
2021-07-30T12:20:17Z ThreadStaker: (811......5239e28) waiting init...
2021-07-30T12:20:17Z ThreadStaker: (9bf......a7cd4bc) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (dc9......c1ff8f5) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (538......89e6a5c) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (bd8......972b8f7) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (41d......aa3a40f) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (051......934da91) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (591......76c8594) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (16b......c63ff78) waiting init...
2021-07-30T12:20:18Z ThreadStaker: (968......2dfe46e) waiting init...

In my oppinion Operators with waiting init must only visible with debug=staking or we need another switch to turn off this info in debug.log.

What do you think, maybe you have special reasons for this behavior?

chrizog:

/kind fix

If the option debug="staking" is used, there is too much output in a longer operation of a masternode. The output is flooded with "ThreadStaker: (833......db786ce) waiting init...".
However, if no debug="staking" is used, there is almost no information in the output about newly minted blocks.
The log levels are adjusted, so that "waiting init" is only visible if debug=staking is activated. Without debug=staking it is not shown, but still some important information about newly minted blocks is logged.

Fixes #648


prasannavl:
Nice! Thank you @chrizog for your contribution. Will review it shortly for minor tweaks, if any and should be good to go after.

src/miner.cpp
```diff
@@ -917,22 +917,22 @@ void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams
             }
             static std::atomic<uint64_t> time{0};
             if (GetSystemTimeInSeconds() - time > 120) {
-                LogPrintf("ThreadStaker: unlock wallet to start minting...\n");
+                LogPrint(BCLog::STAKING, "ThreadStaker: unlock wallet to start minting...\n");
```
prasannavl, line 920: I don't think we want to do this, as this is a user actionable item. If this log is being displayed, it's to prompt to user to either unlock, or remove gen=1 if the intent is not to stake.

src/miner.cpp
```diff
@@ -917,22 +917,22 @@ void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams
...
                 time = GetSystemTimeInSeconds();
             }
             std::this_thread::sleep_for(std::chrono::milliseconds(100));
         }
     }
 
-    LogPrintf("ThreadStaker: started.\n");
+    LogPrint(BCLog::STAKING, "ThreadStaker: started.\n");
```
prasannavl, line 927: This is a single log message per staker. Better to leave an auditable log point.

src/miner.cpp
```diff
@@ -793,7 +793,7 @@ namespace pos {
                     {
                         if (shouldIgnoreMint(ctxState.subNode, blockHeight, creationHeight, subNodesBlockTime, chainparams)) 
                             break;
-                        LogPrint(BCLog::STAKING, "MakeStake: kernel found\n");
+                        LogPrintf("MakeStake: kernel found\n");
```
prasannavl, line 796: How about we leave this to staking log, but instead tweak the other (`LogPrint(BCLog::STAKING, "ThreadStaker: (%s) minted a block!\n", operatorName);`) one to always print since that's more informative?

src/miner.cpp
```diff
@@ -818,7 +818,7 @@ namespace pos {
...
```
prasannavl, line 821: Same as above.

src/miner.cpp
```diff
@@ -951,23 +951,23 @@ void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams
...
                     continue;
                 }
                 else if (status == Staker::Status::minted) {
-                    LogPrintf("ThreadStaker: (%s) minted a block!\n", operatorName);
+                    LogPrint(BCLog::STAKING, "ThreadStaker: (%s) minted a block!\n", operatorName);
                     nMinted[arg.operatorID]++;
                 }
                 else if (status == Staker::Status::initWaiting) {
-                    LogPrintf("ThreadStaker: (%s) waiting init...\n", operatorName);
+                    LogPrint(BCLog::STAKING, "ThreadStaker: (%s) waiting init...\n", operatorName);
```
prasannavl, line 963: How about instead, we keep the first 2 unchanged. However, for the `waiting init` throttle the log message so there's only one every few mins for `waiting init` if staking log isn't enabled.

src/miner.cpp
```diff
@@ -917,22 +917,22 @@ void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams
...
+    LogPrint(BCLog::STAKING, "ThreadStaker: started.\n");
 
     while (!args.empty()) {
         boost::this_thread::interruption_point();
 
         while (fImporting || fReindex) {
             boost::this_thread::interruption_point();
 
-            LogPrintf("ThreadStaker: waiting reindex...\n");
+            LogPrint(BCLog::STAKING, "ThreadStaker: waiting reindex...\n");
```
prasannavl, line 935: A better approach here would be use a throttled logger method for staking log, that throttles psuedo actionable info, like `waiting reindex`, `waiting init`.

src/miner.cpp
```diff
@@ -951,23 +951,23 @@ void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams
...
-                    LogPrintf("ThreadStaker: (%s) waiting init...\n", operatorName);
+                    LogPrint(BCLog::STAKING, "ThreadStaker: (%s) waiting init...\n", operatorName);
                 }
                 else if (status == Staker::Status::stakeWaiting) {
                     LogPrint(BCLog::STAKING, "ThreadStaker: (%s) Staked, but no kernel found yet.\n", operatorName);
                 }
             }
             catch (const std::runtime_error &e) {
-                LogPrintf("ThreadStaker: (%s) runtime error: %s\n", e.what(), operatorName);
+                LogPrint(BCLog::STAKING, "ThreadStaker: (%s) runtime error: %s\n", e.what(), operatorName);
```
prasannavl, line 970: Again, I believe both of these should be clearly auditable points to verify what happened, either if orphaned txs are used somewhere prematurely or if there was an error.

prasannavl (changes requested):

prasannavl:
Thanks for this. I think a better approach here would be the following:

- Use a time throttled logger method for staking log, that throttles psuedo actionable info, like waiting reindex, waiting init and 
unlock wallet.
- Pass all those logs to the time throttled log, and for the time throttled log behaviour, if staking log is enabled, always log. And if it's not, throttle these. 

An alternative that can be quick fix in the meantime: 

- Switch the non informative ones:  

>  LogPrint(BCLog::STAKING, "MakeStake: kernel found\n");

to staking only.

-  If the above is done, we can enable this instead of the above: 

> LogPrintf("New proof-of-stake block found hash: %s\n", hashBlock.GetHex());

chrizog:
> Thanks for this. I think a better approach here would be the following:
> 
> * Use a time throttled logger method for staking log, that throttles psuedo actionable info, like waiting reindex, waiting init and
>   unlock wallet.
> * Pass all those logs to the time throttled log, and for the time throttled log behaviour, if staking log is enabled, always log. And if it's not, throttle these.

I totally agree. I would suggest to add a method "LogPrintFiltered(LogFlags, LogFilter, Args..)" to the Logger class. The LogFilter can be a time throttle filter. It would keep the possibility for adding different log filters in the future. I will push an example here, it can be adjusted then.

I will put the PR to WIP and come back then.

src/miner.cpp
chrizog: Will change them back to LogPrintf as before 👍

src/miner.cpp
chrizog: Will change it back

Stonygan:
> Thanks for this. I think a better approach here would be the following:
> 
> * Use a time throttled logger method for staking log, that throttles psuedo actionable info, like waiting reindex, waiting init and
>   unlock wallet.
> * Pass all those logs to the time throttled log, and for the time throttled log behaviour, if staking log is enabled, always log. And if it's not, throttle these.
> 
> An alternative that can be quick fix in the meantime:
> 
> * Switch the non informative ones:
> 
> > LogPrint(BCLog::STAKING, "MakeStake: kernel found\n");
> 
> to staking only.
> 
> * If the above is done, we can enable this instead of the above:
> 
> > LogPrintf("New proof-of-stake block found hash: %s\n", hashBlock.GetHex());
> 
> Let me know what you think.

I think the same, all error messages like Wallet unlock or Threadstaker terminated must visible everytime in the Log, not only if a explicit parameter is given.
Threadstaker started is one time message, thats ok when its displayed after the restart of defid.
For me the best way is with debug=staking to write to log:
ThreadStaker: (833......db786ce) waiting init..
ThreadStaker: (833......db786ce) Staked, but no kernel found yet.
and without debug=staking log all whats needed to operate and monitor a masternode:
MakeStake: kernel found
new proof-of-stake block found hash:
ThreadStaker: (833......db786ce) minted a block!
ThreadStaker: waiting reindex...
ThreadStaker: unlock wallet to start minting...
ThreadStaker: started.
ThreadStaker: (%s) terminated due to a staking error

That are all messages which not flood the Log and important to monitor the Node.


The ability to see in the log without "debug=staking" which threadstaker minted the block and which hash it minted is immensely important for us for monitoring, especially when more than one staker is running on a server or a staker is running on more than one server.

chrizog:
As proposed the logs that are not disturbing and only one-time logs are activated again (to Logprintf).

The logs for "waiting init" and "staked but no kernel found yet" are now time throttled and logged at most every 10 minutes for each staker (could be adjusted to less or more minutes).

If debug=staking, then everything is logged immediately without time throttling to have the full information to debug the node.

src/logging.h
```diff
@@ -167,4 +168,50 @@ static inline void LogPrint(const BCLog::LogFlags& category, const Args&... args
...
+    uint64_t millis_{0};
+    uint64_t last_time_millis_{0};
+
+public:
+    TimeThrottledFilter() = delete;
+    explicit TimeThrottledFilter(uint64_t millis) : millis_(millis)
+    {
+    }
+
+    virtual bool filter() override
```
bvbfan, line 191: You can remove `virtual` `override` is the correct function identifier.

src/logging.h
```diff
@@ -167,4 +168,50 @@ static inline void LogPrint(const BCLog::LogFlags& category, const Args&... args
...
+    {
+        int64_t current_time = GetTimeMillis();
+        // First call or at least millis_ ms to last call
+        if ((last_time_millis_ == 0) || ((current_time - last_time_millis_) > millis_)) {
+            last_time_millis_ = current_time;
+            return true;
+        }
+        return false;
```
bvbfan, line 199: Reverse logic here. When i call function called filter is expect to understand is something to be filtered or not. So `true` means should be filtered

src/logging.h
```diff
@@ -167,4 +168,50 @@ static inline void LogPrint(const BCLog::LogFlags& category, const Args&... args
...
+    virtual ~ILogFilter(){};
+    virtual bool filter() = 0;
+};
+
+// Implementation that logs at most every millis_ milliseconds
+class TimeThrottledFilter : public ILogFilter
+{
+private:
+    uint64_t millis_{0};
+    uint64_t last_time_millis_{0};
```
bvbfan, line 183: Can we have those 2 variable `atomic_uint64_t` Currently we don't use threads but may change so atomic is more legitimated.

bvbfan (changes requested):
I like the changes overall

src/logging.h
chrizog: Done

src/logging.h
chrizog: Done

src/logging.h
chrizog: Adjusted 👍

src/logging.h
```diff
@@ -167,4 +168,50 @@ static inline void LogPrint(const BCLog::LogFlags& category, const Args&... args
...
+{
+    // Log everything directly if category is enabled..
+    if (LogAcceptCategory((category))) {
+        LogPrintf(args...);
+    } else { // .. and otherwise time throttle
+        if (!time_throttled_filter.filter()) {
+            LogPrintf(args...);
+        }
+    }
+}
```
bvbfan, line 214: I forgot to mention here, Category should be enabled and not filtered to be shown, the problem is throttle will override disabled category so it should be
```
if (LogAcceptCategory(category) && !time_throttled_filter.filter()) {
    LogPrintf(args...);
}
```

chrizog: I think this is intended and it's not overriding. The idea above proposed was that if the category is enabled it should log everything (no throttle) for detailed monitoring or debugging.
If the category is not enabled, you get the log throttled (e.g. every 10 minutes), so the debug.log is not flooded but still you get some information sometimes (e.g. the "waiting init" in the staker).

bvbfan: But if you don't enable the category it should not throttle at all, so that's why we need category enabled + not filtered.

chrizog: The implementation is intended like that. If the category is enabled you want the full output (= no throttling). If the category is not enabled, it should log throttled (e.g. only every 10 minutes), to still have some information, but don't get a flooded debug.log.

This is what @prasannavl mentioned above:
> * Use a time throttled logger method for staking log, that throttles psuedo actionable info, like waiting reindex, waiting init and
>   unlock wallet.
> * Pass all those logs to the time throttled log, and for the time throttled log behaviour, if staking log is enabled, always log. And if it's not, throttle these.

chrizog: Hi @bvbfan and @prasannavl , do you have any updates on this improvement?
If I rebase to master, can we merge it? Or do you have more comments on it?
Thanks!

prasannavl:
Thanks so much @chrizog - this is really good work :) - Sorry for the delay - we had to triage things a bit due to the 2.6.x.

Couple of final thoughts however:

1. I don't think the first condition is needed. I think we can just remove it and simplified.

```
if ((last_time_millis_ == 0) || ((current_time - last_time_millis_) > millis_)) {
```

2. If we indeed go down this path, `LogPrintThrottled` -> `LogPrintCategoryOrThrottled` may be better so it's intent is clearer. 

3. One concern I have about the current approach is that, two different messages are throttled under the same filter. What this means is that one of these messages could be swallowed due to the other.

4. While I really like your abstracted approach here with `ILogFilter`, I also observed the current way it's done is https://github.com/DeFiCh/ain/pull/1006/files#diff-6bf9d2e37d503f2f3381cd3a41c26e0b9670f26c2cd4571e98d2af9e3767c1a1R926-R930. 

So, if we introduce a filter, the filter needs to be able to take this pattern over. 

Thinking through the above, I would like you to consider if it's better to simplify it even further than prematurely engineering an abstraction that might never end up getting utilised elsewhere :) 

I think using the existing pattern in 4, we can just go ahead merge it right away. While it's admittedly not as "nice" as what you've written, I don't realistically see this being used in many other places at the moment, and if we do, we likely need to give more thought to what the global abstraction has to look like (does the filter need to take the args, does these timed objected need to be held on when the category is on, etc).

prasannavl:
Please disregard the previous note. Thinking further, what I wrote above cannot be applied here. Let me come back on what better options we have here.

prasannavl:
Ok, I do think your approach is the a really good way to assimilate with the current code cleanly. However, for the larger abstraction, I think we have options at this point:

1. Use a keyed approach. This is going to make the `LogPrintCategoryOrThrottled`,  (actually a better name now would be `LogPrintFiltered`, since it's abstracted) take another additional key. 
2. Cache the last message inside it, and also let the abstraction pass on `args` into `filter` method. The latter part is likely something that's beneficial to do regardless for the abstraction to be flexible.

prasannavl:
Fastest option however, is to just use https://github.com/DeFiCh/ain/pull/1006/files#diff-6bf9d2e37d503f2f3381cd3a41c26e0b9670f26c2cd4571e98d2af9e3767c1a1R926-R930 with a `thread_local` variable instead. 

Since each staker runs on a different thread, it's already staker scoped.

chrizog:
Hi, I am not sure, if the link from your last comment points to the place you intended. It just points to my files changed in this PR, so I am not sure what you propose as the "fastest option pattern" in your last comment.

chrizog:
The interface was probably engineering too much abstraction. I simplified it a lot now and used the keyed approach with a thread_local map.

prasannavl:
Hi @chrizog - this is fantastic - thank you so much! Sorry for taking so long to come back to this.  

Just made some minor tweaks to merging: 

- Renaming to `LogPrintCategoryOrThreadThrottled`, so intent is clearer. 
- Revert message case, to prevent breakage of tool dependencies, if any - even though your casing was the more sensible one. Will add the casing fix to v3 instead :)

prasannavl:
Thank you @chrizog for helping iterate on this. Much appreciated. It's not as elaborate as it was with filter interfaces as you initially conceived - but it's simpler and well contained. When use-cases expand, that would be a nice time to take the approach you have planted the seed for and generalize it. :)

PR #1007, opened 2021-12-29
Calculate collateral ratio when there is actual price update

bvbfan:
/kind fix


prasannavl:
Remapping to 3.0 since this is non-essential for 2.6.0.

PR #1014, opened 2022-01-03
Bump protocol version

Bushstar:


prasannavl:
Bump proto version only

PR #1018, opened 2022-01-03
Reject block that contains custom tx produced by incorrect masternode

bvbfan:
/kind fix

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -3938,7 +3938,10 @@ bool CChainState::ActivateBestChainStep(CValidationState& state, const CChainPar
             if (!ConnectTip(state, chainparams, pindexConnect, pindexConnect == pindexMostWork ? pblock : std::shared_ptr<const CBlock>(), connectTrace, disconnectpool)) {
                 if (state.IsInvalid()) {
                     fContinue = false;
-                    if (state.GetRejectReason() == "high-hash") {
+                    if (state.GetRejectReason() == "high-hash"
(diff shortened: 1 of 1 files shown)


prasannavl (approved):

dell1503:
I dont understand this
Commit.
So there is after the last update a block produced by a faulty Masternode?
But this means we have an issue with the MN as well?
Thanks for explanation.

prasannavl:
> So there is after the last update a block produced by a faulty Masternode?

No. 

> But this means we have an issue with the MN as well?

Disabled or unsupported TXs (say, ICX TX submitted from a 2.3.x node to the 2.4.x MN which considers it invalid due to being disabled) causes the MN to rollback excessively, which stalls the blockchain when all miners end up in many rollbacks. This commits provides better behaviour for TXs that's considered invalid by the current protocol (say, 2.4.x/2.5.x).

dell1503:
> > So there is after the last update a block produced by a faulty Masternode?
> 
> 
> 
> No. 
> 
> 
> 
> > But this means we have an issue with the MN as well?
> 
> 
> 
> Disabled or unsupported TXs (say, ICX TX submitted from a 2.3.x node to the 2.4.x MN which considers it invalid due to being disabled) causes the MN to rollback excessively, which stalls the blockchain when all miners end up in many rollbacks. This commit provides better behaviour for TXs that's considered invalid by the current protocol (say, 2.4.x/2.5.x).   

Than you for explanation!

PR #1023, opened 2022-01-05
Floor token swap result

bvbfan:
/kind fix


uzyn (approved):

src/masternodes/poolpairs.cpp
```diff
@@ -429,6 +429,11 @@ Res CPoolPair::Swap(CTokenAmount in, PoolPrice const & maxPrice, std::function<R
     }
 
     CAmount result = slopeSwap(in.nValue, reserveF, reserveT, height >= Params().GetConsensus().BayfrontGardensHeight);
+    if (height >= Params().GetConsensus().FortCanningHillHeight && result > 0) {
```
uzyn, line 432: This should be done at `slopeSwap()` and not outside of it - which then creates inconsistencies on the outcome.

uzyn (changes requested):

prasannavl (approved):

PR #1024, opened 2022-01-06
Payback of DUSD loan with DFI

### LINKED ISSUE #999: Tracking Issue for DFI loan payback
Proposal: https://github.com/DeFiCh/dfips/issues/99

Mixa84:

/kind feature

It allows to pay back the loan of currently only DUSD in DFI directly (without swapping on DEX to DUSD) on oracle price of DFI/USD.

Fixes #999


src/masternodes/mn_checks.cpp
```diff
@@ -2692,21 +2700,40 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                    res = mnview.SubBalance(obj.from, CTokenAmount{tokenId, subLoan});
+                    if (!res)
+                        return res;
+
+                    LogPrint(BCLog::LOAN, "CLoanTakeLoanMessage(): Swapping %s interest to DFI - %lld, height - %d\n", loanToken->symbol, subInterest, height);
+                    res = SwapToDFIOverUSD(mnview, tokenId, subInterest, obj.from, consensus.burnAddress, height);
+                }
+                else
+                {
+                    CAmount subLoanInDFI = mnview.GetAmountInCurrency(subLoan, {"DFI","USD"}, false, true, true);
```
bvbfan, line 2727: Declare 3 variables
```
bool useNextPrice = false. requireLivePrice = true, reverseDirection = true;
````

File: src/masternodes/mn_checks.cpp [hunk 2]
```diff
@@ -2692,21 +2700,40 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                    res = TransferTokenBalance(DCT_ID{0}, subLoanInDFI, obj.from, consensus.burnAddress);
```
bvbfan, line 2728: Just use `subBalance` defi amount, loaned tokens aren't burned.

bvbfan (changes requested):

src/masternodes/mn_checks.cpp
Mixa84: But DFI should be like burned. And we should introduce the tracking of it and that is still not done.
```The primary objective during burn is to keep a way to track just these burns in a dedicated manner. One option is to use a dedicated address to burn, so we have separate accountability, to allow for this economics be tweaked in the future. Any other options should also be acceptable if it helps the implementation easier.```
I think this is also meant for DFI token, but maybe I misunderstood it.

File: src/masternodes/mn_checks.cpp (hunk 1 above)
Comment (Mixa84) [reply src/masternodes/mn_checks.cpp:2727]: Here, reverse direction is not the way, so it is false. And all params are default, but maybe it is more readable to put them there.

**Review by uzyn — CHANGES_REQUESTED** [review]
Please include unit tests.

**Review by kuegi — CHANGES_REQUESTED** [review]

File: src/masternodes/masternodes.cpp [hunk 3]
```diff
@@ -920,13 +920,25 @@ CAmount CCollateralLoans::precisionRatio() const
...
 
         auto price = priceResult.val.get();
+
+        if (reverseDirection)
+        {
+            if (price == 0)
+                return Res::Err("Price is zero (%s - %s/%s)", GetDecimaleString(price), priceFeedId.first, priceFeedId.second);
+            auto amountInCurrency = DivideAmounts(amount, price);
+            if (MultiplyAmounts(amountInCurrency, price) != amount)
+                amountInCurrency += 1;
```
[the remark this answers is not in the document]
solros: The reverse direction is only needed when overpaying.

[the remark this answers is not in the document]
solros: Both `MultiplyAmounts` and `DivideAmounts` round down. So I'd say adding this 1 here is okay.

src/masternodes/mn_checks.cpp
solros: > Here, reverse direction is not the way, so it is false. And all params are default, but maybe it is more readable to put them there.

No, reverse direction is correct since we want to get back the amount in DFI from DUSD.

src/masternodes/mn_checks.cpp
```diff
@@ -2692,21 +2700,40 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                    if (!res)
+                        return res;
+
+                    LogPrint(BCLog::LOAN, "CLoanTakeLoanMessage(): Burning interest in DFI directly - %lld, height - %d\n", subInterest, height);
+                    CAmount subInterestInDFI = mnview.GetAmountInCurrency(subInterest, {"DFI","USD"}, false, true, true);
```
solros, line 2733: This (in combination with the +1 in the reverse direction for `GetAmountInCurrency`) can lead to rounding errors where the available DFI balance is less than the amount that should be burned.

One possible fix would be defining `subInterestInDFI` as the minimum of `kv.second - subLoanInDFI` and the expression you have. I.e.,

    CAmount subInterestInDFI = min(kv.second - subLoanInDFI, mnview.GetAmountInCurrency(...) )

However, since when repaying with DFI, both the interest and the loan amount are burned. So this could be made simpler (and less prone to rounding issues) by combining both burns into one transaction? E.g.:

    CAmount subDebtInDFI = min(kv.second, mnview.GetAmountInCurrency(subInterest + subLoan, {"DFI","USD"}, false, true, true);
    res = TransferTokenBalance(DCT_ID{0}, subDebtInDFI, obj.from, consensus.burnAddress);

src/masternodes/mn_checks.cpp
bvbfan: `subBalance` is well enough here.

src/masternodes/mn_checks.cpp
bvbfan: I like the solution

src/masternodes/masternodes.cpp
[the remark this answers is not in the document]
kuegi: AFAIU DivideAmounts doesn't round up.

sample case:
price= 1000
amount = 1 fi
amountInCurrency = 0 -> increased to 1
so we return an amountInCurrency which is 1000x in value of the given amount. This magnitude of "rounding Error" shouldn't be done "silently" within this function IMHO. (on the long run this would lead to a similar error like the 1 fi swap-bug on the DEX)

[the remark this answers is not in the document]
kuegi: > The reverse direction is needed when overpaying.

good point. in that case I would suggest to only use it in case of overpaying, and not everytime then.
AFAIU your proposed changes in mn_checks already does that 👍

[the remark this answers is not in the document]
Mixa84: Just to break the problem little bit and explain some aspects:

- reverse direction is necessary as we don't have oracle feeds on price of USD/DFI. So when user inputs how much DFI he wants to pay we need to convert it to DUSD, then sub the interest and see how much is left for subtracting the base of loan, and in the end those amounts mut be converted again to DFI to sub the balance and here we need reverse direction. What could be done to make this easier is suggested in below comment and I'm thinking of it now is that we could burn the complete DFI in one go and just go one direction to see how much of that is in DUSD and sub the interest and loan part.

- The part with celing 1 satoshi on DFI side is not a bad thing, it basically insures that in the price conversion you don't get to cover some satoshis less on the loan because of the conversions in price, but in the end if the covering of loan costs 1 (or few) more satoshi on the DFI side, that is just the price of it for using DFI to cover the DUSD loan. But 1 satoshi is better than dex price ;)

- I will try to make this simpler, and I ilike the solution for just to sub the amount of DFI inputed and then calc what is that in DUSD and sub interest and what is left on the base.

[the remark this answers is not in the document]
Mixa84: The only problem I see with easier solution is if you overpay. Then we need to convert back from DUSD to DFI to see what is the exact amount of DFI we need to burn.

src/masternodes/mn_checks.cpp
```diff
@@ -2645,6 +2645,14 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
         for (const auto& kv : obj.amounts.balances)
         {
             DCT_ID tokenId = kv.first;
+            auto paybackAmount = kv.second;
+
+            if (height >= Params().GetConsensus().FortCanningHillHeight && kv.first == DCT_ID{0})
+            {
+                tokenId = mnview.GetToken("DUSD")->first;
+                paybackAmount = mnview.GetAmountInCurrency(paybackAmount, {"DFI","USD"});
```
solros, line 2653: The DFIP contained a 1% penalty fee for paying back a DUSD loan in DFI. I cannot see this in the code. Shouldn't this be accounted for here?

src/masternodes/mn_checks.cpp
Mixa84: Yes you are correct, I thought he was commenting on another line.

src/masternodes/mn_checks.cpp
Mixa84: Thanks, the penalty fee will be added from gov variable. It is still a WIP.

src/masternodes/masternodes.cpp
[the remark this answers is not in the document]
kuegi: > * The part with celing 1 satoshi on DFI side is not a bad thing, it basically insures that in the price conversion you don't get to cover some satoshis less on the loan because of the conversions in price, but in the end if the covering of loan costs 1 (or few) more satoshi on the DFI side, that is just the price of it for using DFI to cover the DUSD loan. But 1 satoshi is better than dex price ;)

agreed for the usecase of paying back a loan with DFI. But noone guarantees that this method will only ever be used for that. Having this behaviour in here "implicitly", might lead to problems down the road when someone uses it without realizing this implications. thats my only concern.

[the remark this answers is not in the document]
Mixa84: Currently the feature is hardcoded to payback the DUSD loan when you supply DFI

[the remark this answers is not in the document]
kuegi: > Currently the feature is hardcoded to payback the DUSD loan when you supply DFI

exactly. "currently". But this method is only part of the change, not part of the feature. it might and will definitly be used in other contexts further down the road. Thats why its not called "getDUSDPerDFIPrice" right?
please always consider methods as standalone contract, not only in the way they are used right now.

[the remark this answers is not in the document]
Mixa84: For v3 it will probably be refactored to be able to payback any dToken in loan. But right now, for the next hardfork it was decided to imolement this way and limit for DUSD only.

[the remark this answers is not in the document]
kuegi: ok. just wanted to raise awareness.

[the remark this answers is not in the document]
Mixa84: I have refactored the logic of payback with DFI to be little bit simpler.
- Now reverse direction will be done only in case of overpay, because it needs to be calculated how much of DFI should be really burnt and the rest not touched.
NOTE: overpay will be common way to payback because you are not sure in which block your tx is going to end up so if it doesn't get in next block but after the next then interest will be higher, so overpay assures you have cleared whole loan if that is the intention but only necessary amount of DFI will be used.
- All other loan paybacks that are not overpay with DFI will only be calced in direction of DFI -> DUSD and then loan and interest will be subtracted from loan account and DFI will be burnt in whole amount (not interest and loan separate), so there is no need for reverse direction.

Still waiting for #998 to add DFI payback fee.

src/masternodes/mn_checks.cpp
Mixa84: As now loan and interest are burned combined in one action, I will leave it to transfer balance to burn address if you agree?

src/masternodes/masternodes.cpp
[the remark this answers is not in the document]
prasannavl: Thanks @kuegi for the reviews and the observations. Just sharing some more context around DUSD hard coding: Given the short time for the hard fork - we didn't want to change the protocol message structure to allow all tokens than just DFI for payback - since that requires changes to all downstream APIs and SDKs. So we decided to do this in 2 phases:

Phase 1: Just do the minimal change set for DUSD.
Phase 2: For v3, this will go through another refactor making it generic along with the changes to the payback structure.

bvbfan (dismissed):

bvbfan (dismissed):

src/masternodes/mn_checks.cpp
```diff
@@ -2674,9 +2675,71 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            const auto pAttributes = mnview.GetAttributes();
+            if (pAttributes) {
+                attrs = pAttributes->attributes;
+                CDataStructureV0 activeKey{AttributeTypes::Token, tokenDUSD->first.v, TokenKeys::PaybackDFI};
+                try {
+                    const auto& value = attrs.at(activeKey);
+                    auto valueV0 = boost::get<const CValueV0>(&value);
+                    if (valueV0) {
+                        const auto active = boost::get<const bool>(valueV0);
+                        if (active || *active) {
```
kuegi, line 2691: I assume that this has some logic why it is written that way. but its super hard to read and looks fishy. 🤔

Bushstar: Good spot, || is a bug. Updated.

bvbfan (dismissed):

Issue #521, opened 2021-06-11
Pool Slippage Protection is taking spot price instead of actual price


muirglacier:
Just wanted to address an issue,
the slippage protection as defined here
https://github.com/DeFiCh/ain/blob/master/src/masternodes/poolpairs.cpp#L389-L393
is not working correctly in all cases.

Right now, all it does is take the spot price and compare it to the maximum price the trader specified in his transaction.
There are some cases and scenarios, where this can go completely wrong.

Imagine you are a malicious masternode, and there is a low liquidity pool pair.
You could add heaps of liquidity, and wait for an unsuspected trader to place his order.
The app will show him some potential trade result accounting for the slippage with the current liquidity in the pool, and he will most likely go ahead with the trade.

Now, a malicious masternode could wait for a PoolSwap transaction, and "sandwitch" it with a "Remove Liquidity" -> Victims Pool Swap -> Attackers Arbitrage -> "Add Liquidity" Transaction.

The reason this works is because the spot price stays the same on "remove liquidity", but slippage gets a LOT higher the less liquidity is in the pool. However, slippage protection is not looking at the liquidity, just at the spot price.

Not critical as there is enough liquidity in the "important pools", hence I didnt submit it as a bug bounty.

My suggestion is: Instead of specifying MAX PRICE, let the user specify MIN TOKENS RETURNED.

(This is consensus critical, maybe keep it low priority for one of the future hard forks)

kuegi:
it seems like the additional check in fort canning fixed this already:
https://github.com/DeFiCh/ain/blob/06f2ed8edaec7ea0741ec66e1210b1e8549dec23/src/masternodes/mn_checks.cpp#L3642-L3648

PR #1039, opened 2022-01-16
[poolpairs] improv price-calc for maxPrice check

### LINKED ISSUE #1038: maxPrice for swap check doesn't include slipage
In https://github.com/DeFiCh/ain/blob/06f2ed8edaec7ea0741ec66e1210b1e8549dec23/src/masternodes/poolpairs.cpp#L410-L413 the check for maxPrice only checks against the current price of the pool and doesn't even account for commissions. But the real executed price will be higher, specially for big swap amounts. 

The outcome of the swap is (simplified from calc in https://github.com/DeFiCh/ain/blob/06f2ed8edaec7ea0741ec66e1210b1e8549dec23/src/masternodes/poolpairs.cpp#L459-L467 ) :
`swappedAmount = unswapped * (reserveT / (reserveF + unswapped))`
where unswapped is the inValue, already reduced by the fee.
So the full calculation could be written as:
`swappedAmount = inValue * (inValueReducedByFee/inValue)*(reserveT / (reserveF + inValueReducedByFee))`

Since the calc is `swapped = inValue / price` the estimated price should be : 
```
inValueReducedByFee= inValue - (inValue*commission);
price= (inValue/inValueReducedByFee)*(reserveF+inValueReducedByFee)/(reserveT)
```

Note: in composite swaps, its not as relevant, as the estimated outcome of a swap in the path is correct so the next maxPrice-check in the path will account for it, but specially for direct swaps its a big issue. 
easiest seen by executing 
`testpoolswap {"from":"address1","tokenFrom":"USDT","amountFrom":1000000,"to":"address2","tokenTo":"DFI","maxPrice":3.2}`
which results in 
278 930 (but 1000000 / 3.2 = 312 500 )

kuegi:
probably need a bit rework / cleanup in the actual calculation.
need unittests

/kind fix

improve the calculation of the estimated price for the check against maxPrice in a swap. 
Current calculation doesn't account for slipage due to the actual swap. see referenced issue for more details.

Fixes #1038

### DIFF
--- a/src/masternodes/poolpairs.cpp
+++ b/src/masternodes/poolpairs.cpp
@@ -407,9 +407,16 @@ Res CPoolPair::Swap(CTokenAmount in, PoolPrice const & maxPrice, std::function<R
 
     auto const maxPrice256 = arith_uint256(maxPrice.integer) * PRECISION + maxPrice.fraction;
     // NOTE it has a bug prior Dakota hardfork
-    auto const price = height < Params().GetConsensus().DakotaHeight
+    auto price = height < Params().GetConsensus().DakotaHeight
                               ? arith_uint256(reserveT) * PRECISION / reserveF
                               : arith_uint256(reserveF) * PRECISION / reserveT;
+    if (height >= Params().GetConsensus().FortCanningHillHeight) {
+        auto reducedInValue= in.nValue
+        if (commission) {
+            reducedInValue -= MultiplyAmounts(in.nValue, commission);
+        }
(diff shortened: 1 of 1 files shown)


muirglacier:
Should fix https://github.com/DeFiCh/ain/issues/521

bvbfan:
We do it post swap now, it's more accurate
https://github.com/DeFiCh/ain/blob/master/src/masternodes/mn_checks.cpp#L3905-L3912

PR #1045, opened 2022-01-19
Smart contract exchange of BTC for DFI

Bushstar:


src/chainparams.cpp
```diff
@@ -243,6 +243,9 @@ class CMainParams : public CChainParams {
         consensus.accountDestruction.insert(GetScriptForDestination(DecodeDestination("dJEbxbfufyPF14SC93yxiquECEfq4YSd9L", *this)));
         consensus.accountDestruction.insert(GetScriptForDestination(DecodeDestination("8UAhRuUFCyFUHEPD7qvtj8Zy2HxF5HH5nb", *this)));
 
+        consensus.smartContracts.clear();
+        consensus.smartContracts["DFIP2201"] = GetScriptForDestination(CTxDestination(WitnessV0KeyHash(std::vector<unsigned char>{0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0})));
```
bvbfan, line 247: Can we use `std::vector<unsigned char>(20, 0)` `{}` - it's `initializer_list`

src/masternodes/govvariables/attributes.cpp
```diff
@@ -39,15 +39,23 @@ static ResVal<int32_t> VerifyInt32(const std::string& str) {
     return {int32, Res::Ok()};
 }
 
-static ResVal<CAmount> VerifyPct(const std::string& str) {
+static ResVal<CAmount> VerifyFloat(const std::string& str) {
     CAmount amount = 0;
     if (!ParseFixedPoint(str, 8, &amount) || amount < 0) {
-        return Res::Err("Percentage must be a positive integer or float");
+        return Res::Err("Amount must be a float");
```
bvbfan, line 45: It still could be an integer.

bvbfan (approved):

src/chainparams.cpp
Bushstar: For this leave it in full, next smart contract address will be copy and paste of this with 1 set at the end.

src/masternodes/govvariables/attributes.cpp
Bushstar: Update to "Amount must be positive value"

src/masternodes/mn_checks.cpp
```diff
@@ -1327,6 +1343,131 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            if (minSwap && amount < minSwap) {
+                return Res::Err("Below minimum swapable amount, must be at least " + ValueFromAmount(minSwap).getValStr() + " BTC");
+            }
+
+            const auto token = mnview.GetToken(id);
+            if (!token) {
+                return Res::Err("Specified token not found");
+            }
+
+            if (token->symbol != "BTC" || token->name != "Bitcoin" || !token->IsDAT()) {
```
kuegi, line 1416: why not check directly for the token Id of dBTC which is 2 ? Do we want other DAT token with the name "Bitcoin" and the symbol "BTC" to be swapable too?

Bushstar: Symbol on DAT must be unique, Bitcoin name lookup is redundant here.

src/chainparams.cpp
prasannavl: May be we use something like a boost preprocessor macro to update in compile time and let that increment later. For now copy paste looks to do.

PR #1050, opened 2022-01-24
Rename vars. Update test to avoid errors.

Bushstar:


src/masternodes/mn_checks.cpp
```diff
@@ -1431,19 +1431,19 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
                 return std::move(resVal);
             }
 
-            CDataStructureV0 premiumKey{AttributeTypes::Param, ParamIDs::DFIP2201, DFIP2201Keys::Premium};
-            CAmount premium{2500000};
+            CDataStructureV0 feePctKey{AttributeTypes::Param, ParamIDs::DFIP2201, DFIP2201Keys::Premium};
```
kuegi, line 1434: why keep it "DFIP2201Keys::Premium" when "premium" is misleading enough that you change the variable?

prasannavl: Nice catch. This is just a tiny (and contained) left over of a refactor. The initial rationale was to use a more generic `fee` term here. However, this is a direct implementation of the DFIP, and that's what is the expectation of these supplementary smart contracts mechanism here is - so we decided to stay true to the DFIP and retain the premium wording.

prasannavl: Going ahead and merging, since it's well contained and don't want to keep the PR held on this note.

PR #1052, opened 2022-01-24
Allow set of arguments at running tests

bvbfan:
/kind feature

It allows to start `defid` via Valgrind of GDB, via `export DEFID="/usr/bin/gdb -ex run --args ~/ain/src/defid"`


prasannavl:
Paths with spaces will end up getting split incorrectly. 

Eg: DEFID="/home/some\ test\ folder/defid" could result in the binary path being `/home/some\` which will fail. 

Meanwhile, setting the DEFID="..../myscript.sh" that execs defid, appropriately wrapped could remain a reasonable way to achieve this.

prasannavl:
Closing this for now, as it still ends up in caveats. Current is one step more, but can use external steps for now.

PR #1086, opened 2022-01-29
Add safety asserts on invalid values, full spectrum tests

prasannavl:
- Adding more tests, just for better coverage.


src/masternodes/loan.cpp
```diff
@@ -510,8 +510,13 @@ std::string GetInterestPerBlockHighPrecisionString(base_uint<128> value) {
...
         std::string GetInterestPerBlockString() {
             std::ostringstream result;
-            result << GetInterestPerBlockMagnitude() << ".";
-            result << std::setw(24) << std::setfill('0') << GetInterestPerBlockDecimal();
+            auto mag = GetInterestPerBlockMagnitude();
+            auto dec = GetInterestPerBlockDecimal();
+            // While these can happen theoretically, they should be out of range of
+            // operating interest. If this happens, something else went wrong.
+            assert(mag >= 0);
+            assert(dec >= 0);
```
kuegi, line 518: I would be super careful with asserts in production code. Just reminding that the last time the chain stood still was due to an assert in productioncode. 
could we instead just check the condition, log an error and return an empty string (or meaningfull "invalidNumber")?

bvbfan: @kuegi  no need to worry about, assert a part of rpc layer, say network isn't affect, just a single node. In some situations asserts are pretty needed to find out serious bugs which could affect network in bad aspects.

prasannavl: Thanks for the thoughtful notes @kuegi. As @bvbfan said, there's no risk of network getting affected in this place, as it's the RPC. However, your thoughts did lead me to refactor this a bit, so it's even safer and this doesn't get used unintentionally elsewhere in a place where this can happen. 

That said, would like to echo @bvbfan on the above. Crash is a feature, not a bug, when there are serious contract violations :) - while it's always a challenging question philosophically on if network halting should happen - the current ICX dBTC scenario is a good example to ponder over on if there was an assert in place crashing the network swiftly, fix it promptly, rather than the swaps that happened :)

kuegi: Thanks for the reply @prasannavl . I completly understand what you mean and agree in a way. 
For me, asserts are the absolute last resort and should only be used in cases where nothing but a crash is an option. But I also think that this last resort should nearly never be used or made sure that it really never happens. It's like burning down the house instead of keeping it clean and the walls strong. Sometimes this is necessary, but in 99% of the cases it should be prefered to not let it come to that.
For me, asserts should be used with the meaning "We made sure that this will never happen, but if it does, we need to crash to prevent further problems"  and not with the "This shouldn't happen and it probably never will, lets just put an assert in there".

For me, asserts are a good step in the development-cycle of a feature. If the developer themself thinks that this should (and would) never happen, it makes sense to put an assert into the first version of the code to make this clear. From that the next step should be the question "can this really ever happen? if it might happen: how can we prevent it earlier? and is a crash really the best option here?"

regarding your note of the ICX: yes, would have been great if there was an assert. Like "assert(numberOfincomingBTC == numberOfcreatedBTC)" and this would be one of the cases where I say an assert makes sense. After making sure in the code that this should really never happen and having the assert just in for the unlikely case that everyone missed something. Exactly for the reason you mentioned: "better crash than let a wrong state get to the network". 
But again: such an assert would have let to a thorough investigation in the review, if this assert will really never happen.

I think we are on the same page here 👍

bvbfan (approved):

PR #1101, opened 2022-02-07
Filter out misbehaving node

bvbfan:
/kind fix


Bushstar (approved):

src/net_processing.cpp
```diff
@@ -1070,6 +1070,12 @@ static bool MaybePunishNode(NodeId nodeid, const CValidationState& state, bool v
     case ValidationInvalidReason::TX_MEMPOOL_POLICY:
         break;
     }
+    if (state.GetRejectCode() == REJECT_CUSTOMTX
```
surangap, line 1073: test comment

PR #1102, opened 2022-02-08
Implement `getaccounthistory` RPC

chanakasameera:

/kind feature

Currently, we have listaccounthistory on DeFiCh/ain. However, there isn't a way to filter to get one AccountHistory and we need an RPC to get single account history to resolve balance transfer.

https://github.com/DeFiCh/jellyfish/issues/1025

Fixes #

### DIFF
--- a/src/masternodes/rpc_accounts.cpp
+++ b/src/masternodes/rpc_accounts.cpp
@@ -1211,7 +1211,7 @@ UniValue getaccounthistory(const JSONRPCRequest& request) {
                     {"owner", RPCArg::Type::STR, RPCArg::Optional::NO,
                         "Single account ID (CScript or address)."},
                     {"blockHeight", RPCArg::Type::NUM, RPCArg::Optional::NO,
-                        "Height to iterate from (downto genesis block)."},
+                        "Block Height to search in."},
                     {"txn", RPCArg::Type::NUM, RPCArg::Optional::NO,
                         "for order in block."},
                },
--- a/src/masternodes/rpc_accounts.cpp
+++ b/src/masternodes/rpc_accounts.cpp
@@ -1204,7 +1204,6 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
 }
 
 UniValue getaccounthistory(const JSONRPCRequest& request) {
-    auto pwallet = GetWallet(request);
 
     RPCHelpMan{"getaccounthistory",
                "\nReturns information about account history.\n",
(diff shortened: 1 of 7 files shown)


src/masternodes/rpc_accounts.cpp
```diff
@@ -1203,6 +1203,91 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
+            [&](COutputEntry const & entry, CBlockIndex const * index, CWalletTx const * pwtx) {
+                if (accountId == EncodeDestination(entry.destination) && blockHeight == index->nHeight && txn == entry.vout) {
+                    ret = outputEntryToJSON(entry, index, pwtx);
+                    return true;
+                }
+                else {
+                    return --count != 0;
+                }
+            }
+        );
```
fuxingloh, line 1285: Do we need this? 

My assumption is since we are searching for one history, we don't need to support searching in your own wallet, as you need to explicitly provide all "owner, block height, txn" for this RPC method.

chanakasameera: If we remove this block, getaccounthistory works only for results[0] (please refer test/functional/rpc_getaccounthistory.py:62), therefore decided to add this.

src/masternodes/rpc_accounts.cpp
```diff
@@ -1203,6 +1203,91 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
+               RPCExamples{
+                       HelpExampleCli("getaccounthistory", "cscript 160 10")
+                       + HelpExampleCli("getaccounthistory", "cscript, 160, 10")
+               },
+    }.Check(request);
+
+    std::string accountId = request.params[0].getValStr();
+    CScript owner = DecodeScript(accountId);
+    uint32_t blockHeight = std::stoul(request.params[1].getValStr());
+    uint32_t txn = std::stoul(request.params[2].getValStr());
```
surangap, line 1231: use `.get_int()` since it has runtime errors defined.

chanakasameera: What if we change input param [1] and [2] type to RPCArg::Type::NUM, then it would be more efficient, right ?

src/masternodes/rpc_accounts.cpp
```diff
@@ -1203,6 +1203,72 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
+    paccountHistoryDB->ForEachAccountHistory([&](AccountHistoryKey const & key, AccountHistoryValue const & value) {
+        if (startKey.blockHeight > key.blockHeight) {
+            return false;
+        }
+        if (owner != key.owner) {
+            return false;
+        }
+        CScopeAccountReverter(view, key.owner, value.diff);
+        return true;
+    }, {owner, std::numeric_limits<uint32_t>::max(), std::numeric_limits<uint32_t>::max()});
```
surangap, line 1265: looking at the `listaccounthistory` implementation, this is only one piece of code for adding pool rewards to the output. by default `listaccounthistory` adds rewards to the output. do we need rewards history also via `getaccounthistory` @fuxingloh ? if yes, we might have to add other piece of the code also, and if not, we can remove this I think.

fuxingloh: We don't need pool reward, since you can't get pool reward by specifying the keys.

bvbfan (dismissed):

surangap (dismissed):

surangap:
Please include this in an upcoming release.

src/masternodes/rpc_accounts.cpp
```diff
@@ -1203,6 +1203,46 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
...
 
+UniValue getaccounthistory(const JSONRPCRequest& request) {
+
+    RPCHelpMan{"getaccounthistory",
+               "\nReturns information about account history.\n",
+               {
+                    {"owner", RPCArg::Type::STR, RPCArg::Optional::NO,
+                        "Single account ID (CScript or address)."},
+                    {"blockHeight", RPCArg::Type::NUM, RPCArg::Optional::NO,
+                        "Height to iterate from (downto genesis block)."},
```
surangap, line 1214:
```suggestion
                        "Block Height to search in."},
```

surangap (approved):

bvbfan (approved):

fuxingloh (approved):

PR #1108, opened 2022-02-11
Move loan and collateral token to Gov var

Bushstar:


src/masternodes/govvariables/attributes.cpp
```diff
@@ -319,21 +390,90 @@ Res ATTRIBUTES::Apply(CCustomCSView & mnview, const uint32_t height)
...
         auto attrV0 = std::get_if<CDataStructureV0>(&attribute.first);
-        if (attrV0 && attrV0->type == AttributeTypes::Poolpairs) {
-            uint32_t poolId = attrV0->typeId;
-            auto pool = mnview.GetPoolPair(DCT_ID{poolId});
-            if (!pool) {
-                return Res::Err("No such pool (%d)", poolId);
-            }
-            auto tokenId = attrV0->key == PoolKeys::TokenAFeePCT ?
-                                        pool->idTokenA : pool->idTokenB;
+        if (attrV0) {
```
bvbfan, line 393: Make it
```
if (!attrV0) {
    continue;
}
```

src/masternodes/govvariables/attributes.cpp
```diff
@@ -274,6 +327,24 @@ Res ATTRIBUTES::Validate(const CCustomCSView & view) const
...
+                           attrV0->key == TokenKeys::LoanMintingInterest) {
+                    if (view.GetLastHeight() < Params().GetConsensus().GreatWorldHeight) {
+                        return Res::Err("Cannot be set before GreatWorld");
+                    }
+                } else if (attrV0->key == TokenKeys::FixedIntervalPriceId) {
+                    if (view.GetLastHeight() < Params().GetConsensus().GreatWorldHeight) {
+                        return Res::Err("Cannot be set before GreatWorld");
+                    }
+                    if (const auto currencyPair = std::get_if<CTokenCurrencyPair>(&attribute.second)) {
+                        if (!OraclePriceFeed(const_cast<CCustomCSView&>(view), *currencyPair)) {
```
bvbfan, line 342: Could be moved to apply ?

src/masternodes/loan.cpp
```diff
@@ -49,27 +63,35 @@ void CLoanView::ForEachLoanCollateralToken(std::function<bool (CollateralTokenKe
...
-std::unique_ptr<CLoanView::CLoanSetCollateralTokenImpl> CLoanView::HasLoanCollateralToken(CollateralTokenKey const & key)
+std::optional<CLoanView::CLoanSetCollateralTokenImpl> CLoanView::HasLoanCollateralToken(CollateralTokenKey const & key)
 {
     auto it = LowerBound<LoanSetCollateralTokenKey>(key);
-    if (it.Valid() && it.Key().id == key.id)
+    if (it.Valid() && it.Key().id == key.id) {
         return GetLoanCollateralToken(it.Value());
+    } else if (const auto view = dynamic_cast<const CCustomCSView*>(this)) {
+        return GetCollateralTokenFromAttributes(*view, key.id);
+    }
```
bvbfan, line 73: Make a pure virtual function in `CLoanView::GetCollateralTokenFromAttributes` then implement it in `CCustomCSView`.

src/masternodes/loan.cpp
```diff
@@ -49,27 +63,35 @@ void CLoanView::ForEachLoanCollateralToken(std::function<bool (CollateralTokenKe
...
 {
-    auto loanToken = ReadBy<LoanSetLoanTokenKey,CLoanSetLoanTokenImpl>(id);
-    if (loanToken)
-        return std::make_unique<CLoanSetLoanTokenImpl>(*loanToken);
+    auto loanToken = ReadBy<LoanSetLoanTokenKey, CLoanSetLoanTokenImpl>(id);
+    if (loanToken) {
+        return loanToken;
+    } else if (const auto view = dynamic_cast<const CCustomCSView*>(this)) {
+        return GetLoanTokenFromAttributes(*view, id);
+    }
```
bvbfan, line 93: Same as above.

bvbfan (changes requested):

src/masternodes/govvariables/attributes.cpp
Bushstar: Does not apply anything here, only validates.

bvbfan: Yup, just only to skip `const_cast`

bvbfan (approved):

src/masternodes/govvariables/attributes.cpp
```diff
@@ -60,6 +59,30 @@ static ResVal<CAmount> VerifyPct(const std::string& str) {
...
+    }
+    auto token = trim_all_ws(value[0]).substr(0, CToken::MAX_TOKEN_SYMBOL_LENGTH);
+    auto currency = trim_all_ws(value[1]).substr(0, CToken::MAX_TOKEN_SYMBOL_LENGTH);
+    if (token.empty() || currency.empty()) {
+        return Res::Err("Empty token / currency");
+    }
+    return {{token, currency}, Res::Ok()};
+}
+
+static ResVal<std::string> VerifyBool(const std::string& str) {
```
kuegi, line 75: just a note, but: "VerifyCurrencyPair" makes a currencyPair Result from the string. Why is VerifyBool returning the string in a result and not returning a bool-result?

src/validation.cpp
```diff
@@ -2843,6 +2875,55 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
+            // Apply fixed_interval_price_id first
+            std::map<std::string, std::string> attrsFirst;
+            std::map<std::string, std::string> attrsSecond;
+
+            int loanCount = 0, collateralCount = 0;
+            for (const auto& [id, token] : loanTokens) {
+                std::string prefix{"v0/token/" + std::to_string(id.v) + '/'};
+                attrsFirst[prefix + "fixed_interval_price_id"] = token.fixedIntervalPriceId.first + '/' + token.fixedIntervalPriceId.second;
+                attrsSecond[prefix + "loan_minting_enabled"] = token.mintable ? "true" : "false";
+                attrsSecond[prefix + "loan_minting_interest"] = KeyBuilder(ValueFromAmount(token.interest).get_real());
```
kuegi, line 2907: I know this is just a one-time action, but can't we use the values from the map here with the enum as key? hardcoded string keys like that are so errorprone. :/

src/masternodes/govvariables/attributes.cpp
Bushstar: Got it, good thinking.

src/masternodes/govvariables/attributes.cpp
Bushstar: It cannot be bool as ResVal already overloads bool, conflicts in ResVal are.

```
operator T() const
operator bool() const
```

It is simple enough here to returns the value and compare to "true" when applying to `attribValue`.

bvbfan: Right. It can return `int` then just check `== 1`. We could avoid strings compare.

src/validation.cpp
bvbfan: We could. 
```
auto attr = cache.GetAttributes();
attr->attributes[key] = value;
...
cache.SetVariable(*attr);
```

Bushstar: Updated to use the map from ATTRIBUTES.

src/masternodes/govvariables/attributes.cpp
Bushstar: Okey dokey. Let's do that then.

Bushstar: We always have to compare string as the value is a string, we can do it in VerifyBool, safer than multiple locations.

```
static ResVal<int> VerifyBool(const std::string& str) {
    if (str != "true" && str != "false") {
        return Res::Err(R"(Value must be either "true" or "false")");
    }
    return {str == "true", Res::Ok()};
}
```

Then we set it with...

`attribValue = bool(*res.val);`

Bushstar: https://github.com/DeFiCh/ain/pull/1108/commits/a56455f582933bc42797e14c2ea1bb81d7d98266

bvbfan (approved):

bvbfan (approved):

prasannavl (approved):

PR #1110, opened 2022-02-15
Add txn-no as a pagination token into listaccounthistory rpc

chanakasameera:

/kind feature

Currently listaccounthistory on DeFiCh/ain. There isn't a way to effectively pagination as you can't provide txn-no as a pagination token. To support slice pagination; we need to be able to provide to all params within the AccountHistoryKey. Address & BlockHeight is currently supported but, txn-no isn't.
https://github.com/DeFiCh/jellyfish/issues/1036

Fixes #


src/masternodes/rpc_accounts.cpp
```diff
@@ -1089,6 +1097,10 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
             return false;
         }
 
+        if(txn != std::numeric_limits<uint32_t>::max() && txn != key.txn) {
+            return true;
+        }
+
```
bvbfan, line 1103: Move it below category check, at line 1125

bvbfan (changes requested):

bvbfan (approved):

src/masternodes/rpc_accounts.cpp
```diff
@@ -1117,6 +1125,10 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
             return true;
         }
 
+        if(txn != std::numeric_limits<uint32_t>::max() && txn != key.txn) {
+            return true;
+        }
+
```
surangap, line 1131: hmm. @chanakasameera two things `filtration` and `pagination` within the same RPC. need to be careful. 
https://github.com/DeFiCh/jellyfish/issues/1036 requirement is to address slice pagination.
This way it's acting like a filter that we don't want and hardly someone would use.

bvbfan (dismissed):

src/masternodes/rpc_accounts.cpp
```diff
@@ -1184,6 +1192,9 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
                 return txs.count(pwtx->GetHash()) || startBlock > index->nHeight || index->nHeight > maxBlockHeight;
             },
             [&](COutputEntry const & entry, CBlockIndex const * index, CWalletTx const * pwtx) {
+                if(txn != std::numeric_limits<uint32_t>::max() && txn != entry.vout) {
+                    return true;
+                }
```
surangap, line 1197: do we need this? still feels like a `filteration`

bvbfan: Good point, It should be better to be
```
if (txn != std::numeric_limits<uint32_t>::max() && index->nHeight == maxBlockHeight && txn < entry.vout)
```

test/functional/rpc_listaccounthistory.py
```diff
@@ -72,6 +73,28 @@ def run_test(self):
         for txs in results:
             assert(hasattr(txs['amounts'], '__len__') and (not isinstance(txs['amounts'], str)))
 
+        # list {"maxBlockHeight":103, "txn":1}, should list without blockheight = 103, txn=2. i.e without MintToken
+        results = self.nodes[0].listaccounthistory(collateral_a, {"maxBlockHeight":103, "txn":1})
+        #Note(surangap): the results here has 2 extra send receive entries with different txids. not sure where they come from.
```
surangap, line 78: @bvbfan any idea where these extra entries have come from?

bvbfan: `Auto Auth` is used to ensure auth on specific address.

bvbfan (dismissed):

test/functional/rpc_listaccounthistory.py
surangap, line 78:
```suggestion
```

surangap (approved):
`txn` is an input parameter added in this PR. So there can't be an already existing use-case that uses `txn` input parameter at Cake side. If they also do not depend on `txn` value in the response, then should be good to go.

PR #1111, opened 2022-02-17
Generic DFI payback capability

Mixa84:

/kind feature

Enable payback of all dTokens with DFI and not just dUSD.


bvbfan (approved):

kuegi:
Do I understand correctly, that this feature will technically allow to payback any loanToken with any other dToken. But it will only be activated for paying back DUSD with DFI (as it is right now)? Cause the plans to payback dToken with DFI or DUSD have been abandoned, right? Do we think we will ever activate that feature?

I ask cause I think activating this would destroy one of the major usecases of the dTokens: being open 24/7 and therefore being able to diverge from the oracle price of closed markets. So the question is why to even prepare it technically?

prasannavl:
Correct. No plans to activate beyond DUSD. One of the important motivations behind many of these initiatives in the engineering aspects of DeFiChain, is to expand it's platform capability and as such where it makes sense, not to couple things strictly on the governance direction - as this narrows down ability to test the economy, run things on devnet, testnet and restricts long term experimentation and research to innovate.

From a codebase perspective, it's easier to be flexible when things are agnostic of community decisions and generic - as the community is free to decide on the governance approach and move as needed without constantly being blocked on engineering and vice-versa. Engineering is about enabling the capability. The direction is decided over DFIPs by governance.

The current direction on the particular issue you're referring to is indeed: https://github.com/DeFiCh/dfips/issues/127, which is unrelated to this.

kuegi:
> Correct. No plans to activate at the moment, beyond DUSD as far as I know. 

thanks for your answer. makes perfect sense when you put it like that.

PR #1115, opened 2022-02-22
Use direct path in testpoolswap when available

### LINKED ISSUE #1114: `testpoolswap` RPC should check for a direct swap first when `auto` is specified.
#### What happened:
1. add a CAT-DFI pool.
2. call `testpoolswap` from CAT -> DFI with `path = auto`, `verbose = true`
3. response have as follows. note `pools` field.
`{ path: 'auto', pools: [ '2', '2', '2' ], amount: '48.48437068@0' }`

The above should be a direct swap since `CAT-DFI`  pool is present.
RCA
even in `compositeswap` check for a direct swap is done first and then goes to the `CalculateSwaps()` and `compositeswap()`
ref -> https://github.com/DeFiCh/ain/blob/c0a25f727f32c1b2c32a5120059276ec97e38bd2/src/masternodes/rpc_poolpair.cpp#L919

but in the scenario mentioned above, it will call `CalculateSwaps()` straight. ref -> https://github.com/DeFiCh/ain/blob/c0a25f727f32c1b2c32a5120059276ec97e38bd2/src/masternodes/rpc_poolpair.cpp#L1091

it could be that the `CalculateSwaps()` is unable to handle this scenario well.
But simply adding a check for a direct swap first should solve the issue(same as the `compositeswap` flow. ).
https://github.com/DeFiCh/ain/blob/c0a25f727f32c1b2c32a5120059276ec97e38bd2/src/masternodes/rpc_poolpair.cpp#L1091

Jouzo:

/kind fix

Uses direct path in testpoolswap when available

Fixes #1114

### DIFF
--- a/src/masternodes/rpc_poolpair.cpp
+++ b/src/masternodes/rpc_poolpair.cpp
@@ -997,6 +997,7 @@ UniValue testpoolswap(const JSONRPCRequest& request) {
                        "One of auto/direct (default = direct)\n"
                        "auto - automatically use composite swap or direct swap as needed.\n"
                        "direct - uses direct path only or fails.\n"
+                       "composite - uses composite path only or fails.\n"
                        "Note: The default will be switched to auto in the upcoming versions."
                    },
                    {
@@ -1051,7 +1052,7 @@ UniValue testpoolswap(const JSONRPCRequest& request) {
         int targetHeight = ::ChainActive().Height() + 1;
 
         auto poolPair = mnview_dummy.GetPoolPair(poolSwapMsg.idTokenFrom, poolSwapMsg.idTokenTo);
-        if (poolPair) path = "direct";
+        if (poolPair && path == "auto") path = "direct";
 
         // If no direct swap found search for composite swap
         if (path == "direct") {
@@ -1089,7 +1090,7 @@ UniValue testpoolswap(const JSONRPCRequest& request) {
             auto compositeSwap = CPoolSwap(poolSwapMsg, targetHeight);
 
             std::vector<DCT_ID> poolIds;
-            if (path == "auto") {
+            if (path == "auto" || path == "composite") {
                 poolIds = compositeSwap.CalculateSwaps(mnview_dummy, true);
             } else {
                 path = "custom";
--- a/src/masternodes/rpc_poolpair.cpp
+++ b/src/masternodes/rpc_poolpair.cpp
@@ -1050,9 +1050,11 @@ UniValue testpoolswap(const JSONRPCRequest& request) {
 
         int targetHeight = ::ChainActive().Height() + 1;
 
+        auto poolPair = mnview_dummy.GetPoolPair(poolSwapMsg.idTokenFrom, poolSwapMsg.idTokenTo);
+        if (poolPair) path = "direct";
+
         // If no direct swap found search for composite swap
         if (path == "direct") {
-            auto poolPair = mnview_dummy.GetPoolPair(poolSwapMsg.idTokenFrom, poolSwapMsg.idTokenTo);
             if (!poolPair)
                 throw JSONRPCError(RPC_INVALID_REQUEST, std::string{"Direct pool pair not found. Use 'auto' mode to use composite swap."});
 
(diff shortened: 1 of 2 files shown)


Bushstar (dismissed):

eli-lim:
Just curious, which of the following are we looking to achieve for `testpoolswap`?
- Always use direct path if available
- Use _best_ path (implying some composite path may yield an amount greater than direct - assuming that's possible)

Jouzo:
> Just curious, which of the following are we looking to achieve for `testpoolswap`?
> 
>     1. Always use direct path if available
> 
>     2. Use _best_ path (implying some composite path may yield an amount greater than direct - assuming that's possible)
> 
> 
> Asking because if (2) is preferred, the logic could instead be in `CPoolSwap::CalculatePoolPaths`.

Good question. Here I addressed the issue by going for the first option but that might not always be the best scenario. 
Another option would be to add "composite" to the accepted path values which would always test for `compositeswap`.
Then we would have:
-  `direct`: Always test for a `poolswap`
- `composite`: Always test for a `compositeswap`
- `auto`: Use a direct swap via `poolswap` if available (as currently in this PR) or go through `compositeswap`
- [poolpairs_ids]: Try through a custom path

With this additional path, a user could try for `direct` vs `composite` and choose the best of two paths (scenario 2).

surangap (approved):

surangap:
> 




> > Just curious, which of the following are we looking to achieve for `testpoolswap`?
> > ```
> > 1. Always use direct path if available
> > 
> > 2. Use _best_ path (implying some composite path may yield an amount greater than direct - assuming that's possible)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Asking because if (2) is preferred, the logic could instead be in `CPoolSwap::CalculatePoolPaths`.
> 
> Good question. Here I addressed the issue by going for the first option but that might not always be the best scenario. Another option would be to add "composite" to the accepted path values which would always test for `compositeswap`. Then we would have:
> 
> * `direct`: Always test for a `poolswap`
> * `composite`: Always test for a `compositeswap`
> * `auto`: Use a direct swap via `poolswap` if available (as currently in this PR) or go through `compositeswap`
> * [poolpairs_ids]: Try through a custom path
> 
> With this additional path, a user could try for `direct` vs `composite` and choose the best of two paths (scenario 2).

saw this later. I guess it could be possible that a composite swap yields more than the direct swap 🤔
I think the main problem is that `CalculateSwaps ` is not able to handle the scenario where only the direct pool is available and no other path. need to write a unit test and confirm it.

eli-lim:
> > 
> 
> > > Just curious, which of the following are we looking to achieve for `testpoolswap`?
> > > ```
> > > 1. Always use direct path if available
> > > 
> > > 2. Use _best_ path (implying some composite path may yield an amount greater than direct - assuming that's possible)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Asking because if (2) is preferred, the logic could instead be in `CPoolSwap::CalculatePoolPaths`.
> > 
> > 
> > Good question. Here I addressed the issue by going for the first option but that might not always be the best scenario. Another option would be to add "composite" to the accepted path values which would always test for `compositeswap`. Then we would have:
> > 
> > * `direct`: Always test for a `poolswap`
> > * `composite`: Always test for a `compositeswap`
> > * `auto`: Use a direct swap via `poolswap` if available (as currently in this PR) or go through `compositeswap`
> > * [poolpairs_ids]: Try through a custom path
> > 
> > With this additional path, a user could try for `direct` vs `composite` and choose the best of two paths (scenario 2).

`composite` here would mean "use the best possible composite swap path"? I would think that's a pretty niche use-case 😮.
Regarding the user trying both options - we probably want to avoid having to make 2 RPCs as it would be a performance / UX impediment for wallet users.

> saw this later. I guess it could be possible that a composite swap yields more than the direct swap 🤔 I think the main problem is that `CalculateSwaps ` is not able to handle the scenario where only the direct pool is available and no other path. need to write a unit test and confirm it.

Agree. Deeper - `CalculatePoolPaths` doesn't ever consider a direct path. So we could just have the logic there instead.
Then `auto` would return the "best" among the direct and possible composite paths, since the direct path is also included in the `CalculateSwaps` evaluation.

prasannavl (approved):

PR #1124, opened 2022-03-01
Fix number of totalAuctions in getloaninfo RPC

Jouzo:

/kind fix

Calculate totalAuctions by adding each liquidated vault's batch count instead of counting the number of liquidated vaults. 
This will now match the number of open auctions displayed at https://defiscan.live/auctions


src/masternodes/rpc_loan.cpp
```diff
@@ -1288,7 +1288,11 @@ UniValue getloaninfo(const JSONRPCRequest& request) {
     });
 
     pcustomcsview->ForEachVaultAuction([&](const CVaultId& vaultId, const CAuctionData& data) {
-        totalAuctions++;
+        if (auto data = pcustomcsview->GetAuction(vaultId, height)) {
+            totalAuctions += data->batchCount;
```
bvbfan, line 1292:
```
ForEachVaultAuction([&](const CVaultId& vaultId, const CAuctionData& data)
                                                                     ^^^^
````
Just use it

bvbfan (changes requested):

hidiego (approved):

bvbfan (approved):

PR #1128, opened 2022-03-07
DUSD contributes  to the 50% minimum required collateral

Jouzo:

/kind feature

DUSD is moved to the DFI side of the 50% collateral ratio.
This gives DUSD more utility and backs up its value.


prasannavl:
Let's refactor this a bit to make the branches simpler. Merging for now.

PR #1151, opened 2022-03-18
Compute vaults in parallel

bvbfan:
/kind feature

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -27,6 +27,7 @@
 #include <masternodes/masternodes.h>
 #include <masternodes/mn_checks.h>
 #include <masternodes/vaulthistory.h>
+#include <parallel_cache.h>
 #include <policy/fees.h>
 #include <policy/policy.h>
 #include <policy/settings.h>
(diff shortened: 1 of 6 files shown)


src/validation.cpp
```diff
@@ -3129,19 +3138,23 @@ void CChainState::ProcessLoanEvents(const CBlockIndex* pindex, CCustomCSView& ca
...
 
             // Store state in vault DB
-            if (pvaultHistoryDB) {
-                pvaultHistoryDB->WriteVaultState(cache, *pindex, vaultId, collateral.val->ratio());
+            if (vaultHistory) {
+                vaultHistory->WriteVaultState(cache, *pindex, vaultId, collateral.val->ratio(), batches);
             }
+        };
 
+        cache.ForEachVaultCollateral([&](const CVaultId& vaultId, const CBalances& collaterals) {
```
prasannavl, line 3149: Is it possible to just shard the vaults by the number of hardware concurrency, and provide it a prefix and end key to each before hand, and just let the thread run, and then join them all in the end? 

That would be a faster implementation, since you don't have so much message passing between each threads. It's just one set in the beginning and end.

prasannavl:
2 things to ensure we derive the most of out this parallization:  

- Reusable thread pool without creating and destroying threads. 
- Shard the data, and provide it to to the threads in one-shot, so there's less need for cross thread messaging and sync primitives either futexes or mutex - both significantly expensive actions.

PR #1154, opened 2022-03-22
Update Boost to 1.75

Bushstar:


bvbfan (dismissed):

prasannavl:
What are your thoughts to lower it to 1.75 for now? They resolve the header but also don't break anything else - and then move to 1.77/1.78 for v3 resolving these.

bvbfan:
`v3` is moved to 1.77, you can lowered as you want.

prasannavl:
Also might need to do something like: 

```
$(package)_version_dot=$(subst _,.,$($(package)_version))
$(package)_download_path=https://boostorg.jfrog.io/artifactory/main/release/$($(package)_version_dot)/source/
```

The current ones fails due to the `_` / `.` difference in versions, and fallsback to bitcoin org repository.

prasannavl:
Lowered to 1.75 for now. Let's move to 1.77 on v3 after resolving the warnings.

bvbfan (approved):

PR #1155, opened 2022-03-22
DFI2203 Futures

Bushstar:


bvbfan (dismissed):

src/masternodes/govvariables/attributes.cpp
```diff
@@ -91,6 +93,7 @@ const std::map<uint8_t, std::map<std::string, uint8_t>>& ATTRIBUTES::allowedKeys
                 {"loan_payback_fee_pct",TokenKeys::LoanPaybackFeePCT},
                 {"dex_in_fee_pct",      TokenKeys::DexInFeePct},
                 {"dex_out_fee_pct",     TokenKeys::DexOutFeePct},
+                {"dfip2203_disabled",   TokenKeys::DFIP2203Disabled},
```
Jouzo, line 96:
```suggestion
                {"dfip2203",   TokenKeys::DFIP2203},
```

For consistency with `payback_dfi` and `loan_payback`

src/masternodes/rpc_accounts.cpp
```diff
@@ -2025,6 +2031,231 @@ UniValue executesmartcontract(const JSONRPCRequest& request) {
...
 
+UniValue futureswap(const JSONRPCRequest& request) {
+    auto pwallet = GetWallet(request);
+
+    RPCHelpMan{"futureswap",
+               "\nCreates and submits to the network a futures contract" +
+               HelpRequiringPassphrase(pwallet) + "\n",
+               {
+                       {"address", RPCArg::Type::STR, RPCArg::Optional::NO, "Address to fund contract and receive resulting token"},
+                       {"amount", RPCArg::Type::STR, RPCArg::Optional::NO, "Amount to send in amount@token format"},
```
Jouzo, line 2042: Could we accept multiple non-DUSD amounts ?

src/masternodes/rpc_accounts.cpp
```diff
@@ -2045,6 +2276,9 @@ static const CRPCCommand commands[] =
     {"accounts",    "sendtokenstoaddress",   &sendtokenstoaddress,   {"from", "to", "selectionMode"}},
     {"accounts",    "getburninfo",           &getburninfo,           {}},
     {"accounts",    "executesmartcontract",  &executesmartcontract,  {"name", "amount", "inputs"}},
+    {"accounts",    "futureswap",            &futureswap,            {"name", "amount", "destination", "inputs"}},
```
Jouzo, line 2279:
```suggestion
    {"accounts",    "futureswap",            &futureswap,            {"address", "amount", "destination", "inputs"}},
```

src/masternodes/rpc_oracles.cpp
```diff
@@ -1124,7 +1126,106 @@ UniValue listfixedintervalprices(const JSONRPCRequest& request) {
...
+        obj.pushKV("discountPrice", ValueFromAmount(futuresPrices.discount));
+        obj.pushKV("premiumPrice", ValueFromAmount(futuresPrices.premium));
+        listPrice.push_back(obj);
+
+        return true;
+    });
+    return listPrice;
+}
+
+UniValue getfutures(const JSONRPCRequest& request) {
```
Jouzo, line 1164: Might be clearer with `getfuturesprices`

kuegi:
Sorry for being late to the party/review, but why call it dfip2203 everywhere in the code? wouldn't a more descriptive name like "futures" be better? dfip2202 was a specific smart contract for this purpose, so I get the name. But here we have a nice functionality that can be named instead of a "generic" dfip number.

kuegi:
sorry if i misunderstand it completly, but from the code, the logic feels totally off.

What I understand from the code how it works:
* Users can create a future contract basically saying "I will buy TSLA for X DUSD at the next settlement" but without specifying a current price. So when I do this contract, I need to guess what the futures price will be, (up to a week from now). Seems like there is no reason whatsoever to do that more than a few blocks before the settlement (when I know the current price). (In combination with the next point, this kinda "makes sense", but in the way a future should work, this makes no sense)
* The price where the future settles seems to get set on the previous futures block (to the active oracle price +- rewardPct). So you would say "In one week time, the future will settle for the current price +5%" right? But this contradicts the whole idea behind the futures. They should settle at the current active oracle price (+- 5%) at the time of settlement, not from the active price 1 week before the settlement. Otherwise this whole thing will lock the price of a dToken within a 5% range of the current price, for the next week.

Shouldn't the future be a pair that can be traded normally and its just fixed that all positions will be settled at (currrent-active-oracleprice +-5%) at the settlement time?

src/masternodes/govvariables/attributes.cpp
```diff
@@ -282,6 +302,20 @@ Res ATTRIBUTES::ProcessVariable(const std::string& key, const std::string& value
 
     auto typeKey = itype->second;
 
+    if (type == AttributeTypes::Param) {
+        if (typeId == ParamIDs::DFIP2201) {
+            if (typeKey == DFIPKeys::RewardPct ||
+                typeKey == DFIPKeys::BlockPeriod) {
+                return Res::Err("Unsupported type for DFIP2201 {%d}", typeKey);
+            }
+        } else {
```
Krysh90, line 311: It might not change current functionality...

I would have expected an `else if` which checks `typeId` against `ParamIDs::DFIP2203`, in order to avoid unexpected early returns.

uzyn:
Thanks @kuegi 

I haven't read the code yet. Just responding based on your reviews. 

> * Users can create a future contract basically saying `I will buy TSLA for X DUSD at the next settlement` but without specifying a current price. So when I do this contract, I need to guess what the futures price will be, (up to a week from now). Seems like there is no reason whatsoever to do that more than a few blocks before the settlement (when I know the current price). (In combination with the next point, this kinda "makes sense", but in the way a future should work, this makes no sense)

This is correct. We are not exactly creating a Future here. More like a "future swap".

This serves only to bound dToken to a range. It's also widely discussed that this Future will likely ended up not in use – acting pretty much just like a range guarantee every week.

> * The price where the future settles seems to get set on the previous futures block (to the active oracle price +- rewardPct). So you would say `In one week time, the future will settle for the current price +5%` right? But this contradicts the whole idea behind the futures. They should settle at the current active oracle price (+- 5%) at the time of settlement, not from the active price 1 week before the settlement. Otherwise this whole thing will lock the price of a dToken within a 5% range of the current price, for the next week.
> 
> Shouldn't the future be a pair that can be traded normally and its just fixed that all positions will be settled at (currrentActiveOracleprice +-5%) at the settlement block?

This would be a bug if you're right. I haven't checked the code yet to confirm.


> 
> Also I didn't find a way to get out of the futures-contract once I opened it. So you can't close your position? which makes sense, since I already know exactly what it will settle on. Like looking 1 week into the future, but not the way that I know where the dToken will stand in 1 week, but by locking the price of the future in 1 week to the currentprice+5%.

You should be able to get out by getting back your original input before the swap block. I haven't checked the code yet to see if it's implemented, but specification wise you should be able to get back your original input.

src/masternodes/govvariables/attributes.cpp
Bushstar: I think this is different, it is more of a blacklist entry, the others are enabled when true, this is disabled when true hence the explicit disabled in the name. Could invert it but this may be more clear this way.

src/masternodes/rpc_accounts.cpp
Bushstar: Correct. You can swap dTokens for DUSD.

Bushstar:
@kuegi many thanks for your feedback. You are correct, the implementation does need to be updated to use the Oracle price at the point of execution.

kuegi:
> @kuegi many thanks for your feedback. You are correct, the implementation does need to be updated to use the Oracle price at the point of execution.

Thx for the confirmation. If I understand it correctly, this would mean that we don't need the new "futurePrices" at all, right? (so no datastructure, not storing them, no rpc calls) Cause we only use the active oracle price at the settlement.
Would reduce the size of the change dramatically. And since this is a "safeguard" feature that is probably/hopefully not really used alot, less change is good.

src/masternodes/rpc_accounts.cpp
Jouzo: I was not clear, I meant accepting multiple amounts at the same time. As in changing `CDFIP2203Message`: `CBalances amounts;` instead of `CTokenAmount source{};` to use a single TX

Bushstar: Got it, the TX allows creating a contract for a single source token only.

Bushstar:
@kuegi Correct, but I was thinking of perhaps keeping the `listfuturesprices` and `getfuturesprices` and apply the 5% to the current price, this gives people an idea of what to expect, like a helper function.

src/masternodes/rpc_accounts.cpp
Bushstar: The rationale was because for DUSD source you need to set the destination, with dToken source this is not an issue. Wanted to keep a consistent interface.

kuegi:
> @kuegi Correct, but I was thinking of perhaps keeping the `listfuturesprices` and `getfuturesprices` and apply the 5% to the current price, this gives people an idea of what to expect, like a helper function.

Honestly I wouldn't do that, as this opens a wide door for misunderstandings. Users seeing this rpc call, thinking that those prices are fixed already. then they open futures positions based on this assumptions. And have a possibly really bad awakening on the settlement block.
Yes, docu would state it, you can add warnings, but we know how this usually goes. So I wouldn't open that door. Risk/Reward ratio is just not worth it IMHO.

src/masternodes/govvariables/attributes.cpp
Jouzo: You're right, it's clearer this way for a blacklist

Bushstar:
@kuegi I'll remove them to save misunderstandings.

src/masternodes/govvariables/attributes.cpp
```diff
@@ -282,6 +312,26 @@ Res ATTRIBUTES::ProcessVariable(const std::string& key, const std::string& value
 
     auto typeKey = itype->second;
 
+    if (type == AttributeTypes::Param) {
+        if (typeId == ParamIDs::DFIP2201) {
+            if (typeKey == DFIPKeys::RewardPct ||
+                typeKey == DFIPKeys::BlockPeriod) {
```
kuegi, line 318: since this block in general is prepared for other typeIds etc. Why check here for the "other" values?
This if needs to be adapted whenever a new typeId (with new typeKeys) come along. Would be better to check here for the wanted values and return the error if typeKey is not a wanted one, right?

src/masternodes/govvariables/attributes.cpp
```diff
@@ -312,6 +362,65 @@ Res ATTRIBUTES::ProcessVariable(const std::string& key, const std::string& value
...
+                userFuturesValues[key] = futuresValues;
+            }
+        } else {
+            userFuturesValues[key] = futuresValues;
+        }
+
+        return true;
+    }, {height, {}, std::numeric_limits<uint32_t>::max()});
+
+    const auto resVal = GetFutureSwapContractAddress();
```
kuegi, line 392: could we give this a better name than "resVal"? like "contractAddress" ?

src/masternodes/mn_checks.cpp
```diff
@@ -1452,6 +1461,137 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            }
+
+            if (totalFutures.nValue > 0) {
+                auto res = mnview.StoreFuturesUserValues({height, obj.owner, txn}, {totalFutures, obj.destination});
+                if (!res) {
+                    return res;
+                }
+            }
+
+            res = TransferTokenBalance(obj.source.nTokenId, obj.source.nValue, *resVal, obj.owner);
```
kuegi, line 1565: missing check that obj.source.nValue <= totalFutures.nValue.
Otherwise it would withdraw more than the user send to the contract, right?

src/masternodes/govvariables/attributes.cpp
```diff
@@ -312,6 +362,65 @@ Res ATTRIBUTES::ProcessVariable(const std::string& key, const std::string& value
...
+
+        auto res = mnview.SubBalance(*resVal, value.source);
+        if (!res) {
+            return res;
+        }
+
+        res = mnview.AddBalance(key.owner, value.source);
+        if (!res) {
+            return res;
+        }
```
kuegi, line 411: why not use transferTokenBalance?

src/masternodes/mn_checks.cpp
```diff
@@ -1452,6 +1461,137 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+            }
+
+            CDataStructureV0 tokenKey{AttributeTypes::Token, obj.source.nTokenId.v, TokenKeys::DFIP2203Disabled};
+            const auto disabled = attributes->GetValue(tokenKey, false);
+            if (disabled) {
+                return Res::Err("DFIP2203 currently disabled for token %s", obj.source.nTokenId.ToString());
+            }
+        }
+
+        const auto resVal = GetFutureSwapContractAddress();
```
kuegi, line 1518: could we name it "contractAddressValue"? improves readability later one a lot.

src/masternodes/rpc_accounts.cpp
```diff
@@ -2025,6 +2031,314 @@ UniValue executesmartcontract(const JSONRPCRequest& request) {
...
+    }, {std::numeric_limits<uint32_t>::max(), {}, std::numeric_limits<uint32_t>::max()});
+
+    return listFutures;
+}
+
+UniValue getpendingfutureswaps(const JSONRPCRequest& request) {
+    RPCHelpMan{"getpendingfutureswaps",
+               "Get specific pending futures.\n",
+                {
+                       {"address", RPCArg::Type::STR, RPCArg::Optional::NO, "Address to get futures prices for"},
```
kuegi, line 2270: not "futures prices" but "pending future swaps", right?

src/masternodes/mn_rpc.h
```diff
@@ -61,5 +61,6 @@ CAccounts SelectAccountsByTargetBalances(const CAccounts& accounts, const CBalan
 void execTestTx(const CTransaction& tx, uint32_t height, CTransactionRef optAuthTx = {});
 CScript CreateScriptForHTLC(const JSONRPCRequest& request, uint32_t &blocks, std::vector<unsigned char>& image);
 CPubKey PublickeyFromString(const std::string &pubkey);
+std::optional<std::pair<CAmount, CAmount>> GetFuturesBlockAndReward();
```
kuegi, line 64: general question: why call it "reward" everywhere? isn't it a "offset", "fee" or something like that? definitly not a "reward" right?

src/validation.cpp
```diff
@@ -3284,6 +3288,121 @@ void CChainState::ProcessLoanEvents(const CBlockIndex* pindex, CCustomCSView& ca
...
+        if (source->symbol == "DUSD") {
+            const DCT_ID destId{futuresValues.destination};
+            const auto destToken = cache.GetLoanTokenByID(destId);
+            assert(destToken);
+
+            try {
+                const auto& premiumPrice = futuresPrices.at(destId).premium;
+                if (premiumPrice > 0) {
+                    const auto total = DivideAmounts(futuresValues.source.nValue, premiumPrice);
+                    cache.AddBalance(key.owner, {destId, total});
```
kuegi, line 3364: are we missing EraseFuturesUserValues here? (its called on the unpaidContracts, but not on the paid ones it seems).
same in the DUSD case.

src/validation.cpp
```diff
@@ -3284,6 +3288,121 @@ void CChainState::ProcessLoanEvents(const CBlockIndex* pindex, CCustomCSView& ca
...
+        }
+
+        return true;
+    }, {static_cast<uint32_t>(pindex->nHeight), {}, std::numeric_limits<uint32_t>::max()});
+
+    const auto resVal = GetFutureSwapContractAddress();
+    assert(resVal);
+
+    CDataStructureV0 liveKey{AttributeTypes::Live, ParamIDs::Economy, EconomyKeys::DFIP2203Tokens};
+    auto balances = attributes->GetValue(liveKey, CBalances{});
```
kuegi, line 3390: currently we are only tracking the amounts that are send into the contract right? no tracking of paid out funds.
Might risk another AtomicSwap-like-leak?

kuegi (changes requested):

src/masternodes/mn_checks.cpp
Bushstar: That is checked here.

https://github.com/DeFiCh/ain/pull/1155/files#diff-5b5f5b126aad5172f5e15066f2926f4afb57e75eb624b509f56233c14cd4b5c9R1553

src/masternodes/govvariables/attributes.cpp
Bushstar: It is not available there, currently a member of CCustomTxVisitor which uses members of its class. Would have to be moved out and all current uses of it updated.

src/validation.cpp
Bushstar: Only unpaid contracts get removed, we want to keep the history of paid contracts.

src/masternodes/rpc_accounts.cpp
```diff
@@ -2025,6 +2031,314 @@ UniValue executesmartcontract(const JSONRPCRequest& request) {
...
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
+    }
+
+    CFutureSwapMessage msg{};
+    msg.owner = GetScriptForDestination(dest);
+    msg.source = DecodeAmount(pwallet->chain(), request.params[1], "");
+
+    if (!request.params[2].isNull()) {
+        msg.destination = request.params[2].get_int();
+    }
```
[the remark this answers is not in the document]
Bushstar: Those checks happen in mn_checks.cpp when parsing the CFutureSwapMessage TX, this happens at the end of the RPC call via execTestTx.

src/masternodes/mn_checks.cpp
kuegi: Ah right, there is a check within Sub. thx.

bvbfan (dismissed):

src/validation.cpp
Bushstar: History of paid out funds added. Will add an RPC call in a new PR.

kuegi: awesome. thank you. (I cant find how to resolve discussions. I think I don't have the rights....)

bvbfan (dismissed):

src/masternodes/mn_rpc.h
prasannavl: This is correct. A bit of obsolete terminology that slipped through and got shipped. Internal code can be easily refactored. But change on attrs is probably now have to be there for a bit.

prasannavl: Resolving this at the moment, since there's no action points unfortunately - and was a bit constrained on bandwidth to correct this in time.

kuegi: thx. could I do anything different on reporting the next time? so that it gets attention before getting shipped?

PR #1181, opened 2022-04-01
Invalidate block at any time

bvbfan:
/kind feature


Mixa84 (approved):
Finally, so we don't need to wait invalidate/reconsider to finish and we can invalidate another block before it syncs to tip.

PR #1197, opened 2022-04-12
Schedule wallet backup every 2 hours

bvbfan:
/kind feature

Doing wallet backup every 2 hours per default, in case of corruption during sync.
Configuration `-walletbackup=0` disable backup otherwise indicate hours to be backup performed
Backup file is placed next to wallet dat file with suffix `.back`

### DIFF
--- a/src/wallet/load.cpp
+++ b/src/wallet/load.cpp
@@ -3,12 +3,17 @@
 // Distributed under the MIT software license, see the accompanying
 // file LICENSE or http://www.opensource.org/licenses/mit-license.php.
 
-#include <wallet/load.h>
-
+#include <chrono>
+#include <condition_variable>
 #include <interfaces/chain.h>
+#include <logging.h>
 #include <scheduler.h>
+#include <shutdown.h>
+#include <sync.h>
 #include <util/system.h>
+#include <util/time.h>
 #include <util/translation.h>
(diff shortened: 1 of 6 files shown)


prasannavl (approved):

src/wallet/load.cpp
```diff
@@ -77,27 +82,88 @@ bool LoadWallets(interfaces::Chain& chain, const std::vector<std::string>& walle
...
+                if (!fs::exists(walletFile)) {
+                    continue;
+                }
+                auto now = GetTime();
+                auto prevBackup = strprintf("%s.auto.%d.bak1", walletFile, now);
+                if (fs::exists(prevBackup)) {
+                    fs::remove(prevBackup);
+                }
+                if (pwallet->BackupWallet(prevBackup)) {
+                    auto currBackup = strprintf("%s.auto.%d.bak", walletFile, now);
```
prasannavl, line 122: This just renames `prevBackup` into `currBackup`, but is there a reason to do this, since prevbackup is a constructed name anyway with the timestamp. 

In order to remove the oldest file, we need to scan the fs, and remove the oldest one, leaving 2 intact.

bvbfan: It's just tries to prevent someone else to inject a file to screw up the process.

PR #1201, opened 2022-04-15
Calculate owner rewards for futures

bvbfan:
/kind fix


src/masternodes/consensus/smartcontracts.cpp
```diff
@@ -172,6 +172,10 @@ Res CSmartContractsConsensus::operator()(const CFutureSwapMessage& obj) const {
         if (!res)
             return res;
     } else {
+        // some txs might be rejected due to not enough owner amount
+        if (static_cast<int>(height) >= consensus.GreatWorldHeight)
```
prasannavl, line 176: Can we add a comment to remove this branch after v3 hard fork? Should be safe to remove after all nodes upgrade.

prasannavl (approved):

Issue #1202, opened 2022-04-18
Update `CFuturesUserKey` to have `owner` as the first


surangap:
#### What would you like to be added:
currently
```
struct CFuturesUserKey {
    uint32_t height;
    CScript owner;
    uint32_t txn;
```

update it to
```
struct CFuturesUserKey {
    CScript owner;
    uint32_t height;
    uint32_t txn;
```

#### Why is this needed:

Improves perf for `getpendingfutureswaps` RPC.

currently `ForEachFuturesUserValues ` is used in following locations
https://github.com/DeFiCh/ain/blob/e63ae12e399b846c5f97dbf3cb1e771f75263255/src/validation.cpp#L3354
https://github.com/DeFiCh/ain/blob/e63ae12e399b846c5f97dbf3cb1e771f75263255/src/validation.cpp#L3407
https://github.com/DeFiCh/ain/blob/e63ae12e399b846c5f97dbf3cb1e771f75263255/src/validation.cpp#L3434

should not be a problem if we change the index order.

PR #1205, opened 2022-04-20
[DO NOT MERGE] Address based lookup for future swaps

chanakasameera:

/kind chore

Performance improvement for owner address based lookup for future swaps

Fixes https://github.com/DeFiCh/ain/issues/1202

### DIFF
--- a/src/masternodes/accounts.cpp
+++ b/src/masternodes/accounts.cpp
@@ -109,6 +109,28 @@ uint32_t CAccountsView::GetBalancesHeight(CScript const & owner)
     return ok ? height : 0;
 }
 
+void CAccountsView::CreateFuturesMultiIndexIfNeeded()
+{
+    CFuturesUserKeyOwner anyOwnerKey{{}, ~0u, ~0u};
+    if (auto it = LowerBound<ByFuturesSwapKeyOwner>(anyOwnerKey); it.Valid()) {
+        return;
+    }
+
+    LogPrint(BCLog::BENCH, "FuturesSwap - Adding multi index in progress...\n");
+
+    auto startTime = GetTimeMillis();
+
+    CFuturesUserKey startKey{~0u, {}, ~0u};
+    auto it = LowerBound<ByFuturesSwapKey>(startKey);
+    for (; it.Valid(); it.Next()) {
+        WriteBy<ByFuturesSwapKeyOwner>(Convert(it.Key()), '\0');
+    }
+
+    Flush();
+
+    LogPrint(BCLog::BENCH, "FuturesSwap - Multi index took: %dms\n", GetTimeMillis() - startTime);
+}
+
 Res CAccountsView::StoreFuturesUserValues(const CFuturesUserKey& key, const CFuturesUserValue& futures)
 {
     if (!WriteBy<ByFuturesSwapKey>(key, futures)) {
--- a/src/masternodes/accounts.cpp
+++ b/src/masternodes/accounts.cpp
@@ -126,6 +126,9 @@ void CAccountsView::CreateFuturesMultiIndexIfNeeded()
         WriteBy<ByFutureSwapOwnerKey>(TranslateKeyToOwnerPrefix(it.Key()), NON_SERIALIZED_EMPTY_VALUE);
     }
 
+    // Flush the cache
+    Flush();
+
     LogPrint(BCLog::BENCH, "FuturesSwap - Multi index took: %dms\n", GetTimeMillis() - startTime);
 }
 
--- a/src/masternodes/accounts.cpp
+++ b/src/masternodes/accounts.cpp
@@ -104,6 +104,9 @@ Res CAccountsView::StoreFuturesUserValues(const CFuturesUserKey& key, const CFut
     if (!WriteBy<ByFuturesSwapKey>(key, futures)) {
         return Res::Err("Failed to store futures");
     }
+    if (!WriteBy<ByFuturesSwapKeyOwner>(std::make_pair(key.owner, key), '\0')) {
+        return Res::Err("Failed to store futures by owner");
+    }
 
     return Res::Ok();
 }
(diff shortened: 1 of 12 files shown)


surangap (commented):
LGTM. Add tests

src/masternodes/accounts.cpp
```diff
@@ -99,25 +109,60 @@ uint32_t CAccountsView::GetBalancesHeight(CScript const & owner)
     return ok ? height : 0;
 }
 
+void CAccountsView::CreateFuturesMultiIndexIfNeeded()
```
surangap, line 112: checking if we can add a boost test to check this function
there are some storage tests in here ->`src/test/storage_tests.cpp`
or else can run the new binary on old data and check if `CreateFuturesMultiIndexIfNeeded` works as intended.

Jouzo (approved):

src/masternodes/accounts.cpp
chanakasameera: Thanks for pointing out, Updated and added testing for `CreateFuturesMultiIndexIfNeeded`.
Wanted to remove `Flush()` command from `CreateFuturesMultiIndexIfNeeded`, otherwise the test will fail.

surangap: hmm. Currently `CDBWrapper::Flush()` does nothing 🤔  But I see @bvbfan has updated the code in v3.
Flush() is supposed to write to the flat file if any.
but if that's already happening at write, I guess not having `Flush` here is fine.
@prasannavl if you have more insight into this.

surangap: Update 
further reading of the code, final resolution of the Flush should be -> https://github.com/DeFiCh/ain/blob/e63ae12e399b846c5f97dbf3cb1e771f75263255/src/flushablestorage.h#L157

src/test/storage_tests.cpp
```diff
@@ -465,4 +466,51 @@ BOOST_AUTO_TEST_CASE(LowerBoundTest)
...
+    }
+
+    // CreateFuturesMultiIndex
+    pcustomcsview->CreateFuturesMultiIndexIfNeeded();
+
+    // After CreateFuturesMultiIndex, There should be key value pairs CFuturesUserOwnerPrefixKey
+    for(int i = 0; i < 4; ++i) {
+        char c;
+        const CFuturesUserOwnerPrefixKey ownerKey = {key[i].owner, key[i].height, key[i].txn};
+        BOOST_CHECK(pcustomcsview->ReadBy<CAccountsView::ByFutureSwapOwnerKey>(ownerKey, c));
```
surangap, line 501: @prasannavl @Bushstar Adding a `Flush()` at the end of  `CreateFuturesMultiIndexIfNeeded` will cause this line to fail.

chanakasameera: The Flush() endpoint is https://github.com/DeFiCh/ain/blob/master/src/flushablestorage.h#L300

surangap (approved):

src/masternodes/accounts.h
```diff
@@ -35,11 +35,40 @@ struct CFuturesUserKey {
...
+            uint32_t height_ = ~height;
+            READWRITE(WrapBigEndian(height_));
+            uint32_t txn_ = ~txn;
+            READWRITE(WrapBigEndian(txn_));
+        }
+    }
+
+    bool operator<(const CFuturesUserHeightPrefixKey& o) const {
+        return std::tie(owner, height, txn) < std::tie(o.owner, o.height, o.txn);
+    }
```
bvbfan, line 69: Unneeded

src/serialize.h
```diff
@@ -256,8 +260,9 @@ template<typename Stream> inline void Unserialize(Stream& s, Span<unsigned char>
 template<typename Stream> inline void Serialize(Stream& s, bool a)    { char f=a; ser_writedata8(s, f); }
 template<typename Stream> inline void Unserialize(Stream& s, bool& a) { char f=ser_readdata8(s); a=f; }
 
-
-
+// Non serialized empty value
+template<typename Stream> inline void Serialize(Stream& s, NonSerializedEmptyValue a) { }
+template<typename Stream> inline void Unserialize(Stream& s, NonSerializedEmptyValue& a) { a = NonSerializedEmptyValue{}; }
```
bvbfan, line 265: prefer empty body, now it's pure non-sense

src/masternodes/rpc_accounts.cpp
```diff
@@ -2297,7 +2297,7 @@ UniValue getpendingfutureswaps(const JSONRPCRequest& request) {
     LOCK(cs_main);
 
     std::vector<CFuturesUserValue> storedFutures;
-    pcustomcsview->ForEachFuturesUserValues([&](const CFuturesUserKey& key, const CFuturesUserValue& futuresValues) {
+    pcustomcsview->ForEachFuturesUserValues([&](const CFuturesUserHeightPrefixKey& key, const CFuturesUserValue& futuresValues) {
 
         if (key.owner == owner) {
```
bvbfan, line 2302: Make it
```
if (key.owner != owner) {
    return false;
}
```

bvbfan (changes requested):
Users of `ForEachFuturesUserValues`, when owner is set, should early return when he differs

bvbfan:
Overall `CFuturesUserKey` remaned to `CFuturesUserHeightPrefixKey` hurts the code, it does not provide any meaningful information since `CFuturesUserOwnerPrefixKey` is private class. Same for `TranslateKeyTo...`, it's over-engineering.

src/serialize.h
prasannavl: @bvbfan - it doesn't prevent someone from accidentally modifying the value (or worse the static values) . 
This resets always to the expected value post serialisation attempt to avoid surprises.

Not opinionated -- but I'm curious on your how you measure "purity" of non-sense ;)

prasannavl:
Better naming of variables to provide intuitive sense to third party readers, I'd argue is stronger principles of engineering - not over engineering - I particularly disagree with the assertion on `Translate..`. I find that to be far more sensible. 

That said, this PR is not a merge candidate. To provide interim hotfix builds for high performance future swap query only. 
@bvbfan is there a blocker for hotfix merge? Otherwise, I'm inclined to proceed.

src/masternodes/accounts.h
prasannavl: Sure. Can be removed.

bvbfan:
Right now PR has no meaning, you miss the entire purpose
`ForEachFuturesUserValues` should stops on different owne.

chanakasameera:
> Right now PR has no meaning, you miss the entire purpose `ForEachFuturesUserValues` should stops on different owne.

Will add owner checking https://github.com/DeFiCh/ain/pull/1205#discussion_r859407446

src/serialize.h
prasannavl: Indeed, that's the intention. But forces someone to think about it when someone adds something that requires init into the class for whatever reason instead of it silently being ignored. 

Either way, compiler will likely consider it empty block until precursors change. Remove if that's make it appear cleaner - not opinionated on this.

prasannavl:
Agree on https://github.com/DeFiCh/ain/pull/1205#issuecomment-1110628177. Thanks.

prasannavl:
Closing, as this is already in `hotfix branch as well `hotfixes/2.7.0-o1` and `hotfixes/2.7.0-o2`

PR #1207, opened 2022-04-20
Beautify consensus logic

bvbfan:
/kind refactor


Mixa84 (approved):
Great readability of the code and the condition of error.

PR #1214, opened 2022-04-27
Expose live dex statistics

bvbfan:
/kind feature


src/masternodes/mn_checks.cpp
```diff
@@ -3989,6 +3989,14 @@ Res CPoolSwap::ExecuteSwap(CCustomCSView& view, std::vector<DCT_ID> poolIDs, boo
         mnview.Flush();
     }
 
+    auto attributes = view.GetAttributes();
+    if (!attributes) {
+        attributes = std::make_shared<ATTRIBUTES>();
+    }
```
prasannavl, line 3995:
```suggestion
    auto attributes = view.GetAttributes();
    assert(attributes);
```

PR #1224, opened 2022-05-11
devnet v3

hidiego:

CI will create a docker image from this PR and push it to docker hub. Facilitates the deployment of devnet.

/kind fix


prasannavl:
PR reason was to add change devnet keys, if I recall correctly. Better to do this fresh on master, since this seems to be based off v3.

PR #1232, opened 2022-05-13
Don't check merkle root temporary

bvbfan:
/kind fix


prasannavl:
Test only PR. Related fixes in #1214 now.

PR #1242, opened 2022-05-17
Simplify caches

prasannavl:
- Instead of scaled cache from dbcache, simplify caches to use the same value for all the DBs. 
- This has the implication of higher memory requirements, but provide better intuition on caching and simpler semantics.
- Make default writes to do `sync=false`. This should massively speed-up - this should offload the write to OS, while reducing the overall locked time in the application. sync=false moves the responsibility of rest of the write to the OS, so it is resistant to process crash. Only machine crash dynamics are changed. This should be an acceptable compromise for the benefits gained.
- Add additional bloom filter bits, since we do end up with very large set, eg. enhancedcs db.

### DIFF
--- a/src/dbwrapper.cpp
+++ b/src/dbwrapper.cpp
@@ -101,7 +101,7 @@ static leveldb::Options GetOptions(size_t nCacheSize)
 {
     leveldb::Options options;
     options.block_cache = leveldb::NewLRUCache(nCacheSize);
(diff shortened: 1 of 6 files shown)


prasannavl:
Don't need this for now. https://github.com/DeFiCh/ain/pull/1269 is in master.

PR #1250, opened 2022-05-19
Move consortium to separate key in attributes instead under token key.

Mixa84:

/kind refactor

Put consortium members in `v0/consortium/<token_id>/...` instead of `v0/token/<token_id>/consortium_members/...`
Also made live economy consortium accounting to be per token: `v0/live/economy/consortium_global/1/minted`


prasannavl (commented):
v3 is C++ 17 enabled :) -- so we can simplify these with [key, value]! Though, just marking as a remark. Nothing to be blocked on.

prasannavl (approved):

PR #1259, opened 2022-05-22
WIP Token splits changes from master

Bushstar:


src/masternodes/mn_checks.cpp
```diff
@@ -953,9 +953,11 @@ bool IsVaultPriceValid(CCustomCSView& mnview, const CVaultId& vaultId, uint32_t
     if (auto loans = mnview.GetLoanTokens(vaultId))
         for (const auto& loan : loans->balances)
             if (auto loanToken = mnview.GetLoanTokenByID(loan.first))
-                if (auto fixedIntervalPrice = mnview.GetFixedIntervalPrice(loanToken->fixedIntervalPriceId))
+                if (auto fixedIntervalPrice = mnview.GetFixedIntervalPrice(loanToken->fixedIntervalPriceId)) {
                     if (!fixedIntervalPrice.val->isLive(mnview.GetPriceDeviation()))
                         return false;
-
+                } else {
```
prasannavl, line 959: This appears correct, and probably should also branch more like this short-circuit. However, we might want to test this with a sync.

Bushstar: Indeed. Already mentioned this to Jeremy. He is performing a full sync of #1243 with the same change.

PR #1262, opened 2022-05-23
Add filtering mechanism for listgovs

prasannavl:
/kind feature

- `listgovs` can get very large after 2.8.0. 
- This introduces 
  - `listgovs all/live/gov/legacy/<any-prefix-for-attrs>`
  - `all` - Displays everything
  - `legacy` - Meant to keep the similar surface area as pre 2.8.0 (NOT IMPLEMENTED YET)
  - `gov` - List all gov vars only without any attributes
  - `live` - Display live attributes only. Useful for quickly getting stats
  - `<any-prefix`>: Eg: `v0/token/15`, `v0/token`, `token/15`, `tok`, etc to filter attributes and display those attributes only.


src/rpc/client.cpp
```diff
@@ -314,6 +314,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "setgovheight", 0, "variables" },
     { "setgovheight", 1, "height" },
     { "setgovheight", 2, "inputs" },
+    { "listgovs", 0, "prefix" },
```
Bushstar, line 317: No need to convert.

Bushstar (approved):

src/rpc/client.cpp
prasannavl: Resolved

PR #1271, opened 2022-05-23
Add legacy filtering mode

prasannavl:
/kind feature

- This completes the missing legacy filtering impl to make sure the default output is compatible with previous versions. 
- all / live / gov / <prefix> can be used for the rest.

### DIFF
--- a/test/functional/feature_setgov.py
+++ b/test/functional/feature_setgov.py
@@ -620,7 +620,7 @@ class GovsetTest (DefiTestFramework):
         assert_equal(self.nodes[0].listgovs("live"), [[{'ATTRIBUTES': {}}]])
 
         result_all = self.nodes[0].listgovs("all")
-        result_legacy = self.nodes[0].listgovs("legacy")
+        result_legacy = self.nodes[0].listgovs()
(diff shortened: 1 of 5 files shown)


test/functional/feature_setgov.py
```diff
@@ -619,8 +619,8 @@ def run_test(self):
         # live test. No live on this. For now, just ensure, no other attrs are listed
         assert_equal(self.nodes[0].listgovs("live"), [[{'ATTRIBUTES': {}}]])
 
-        result_all = self.nodes[0].listgovs("all")
-        result_legacy = self.nodes[0].listgovs("legacy")
+        result_all = self.nodes[0].listgovs("attrs")
```
prasannavl, line 622: This is fail, however.

prasannavl: Keep this all.

prasannavl:
Once "attrs" are used. It changes the index. Need to adapt.

prasannavl:
The change appear to have been revert. Added the correct fix, which is to change the index. Also change defaults and added a bit more.

Bushstar (approved):

PR #1273, opened 2022-05-24
Add mocknet helpers

prasannavl:
/kind feature

Adds the following undocumented flags:

- `-mocknet` - Enables mocked mainnet for testing. Useless without further patch.
- `-mocknet-blocktime=<secs>` - Block time option when mocknet is enabled
- `-mocknet-key=<pubkey>` - Adds an optional foundation key into mocknet for testing. 

Has the patch file: `contrib/devtools/mocknet.patch`. 

- The patch nullifies the difficulty of mocknet, so that it's easier to mine with any MNs that's carried over into mocknet. 
- The patch is technically not required and can just be a part of the main code, as it is gated. However, just to ensure there is no accidental or potential exploit that can utilize it, it is for now kept completely off the main code and carried as an optional patch. 

Procedure for mock testing mainnet:

- `git apply  contrib/devtools/mocknet.patch`
- Change the fork height as desired on chainparams. 
- Start with `-mocknet [-mocknet-blocktime=<secs>] [-mocknet-key="pubkey"]

This allows easier testing, and further automated mocked testing of mainnet. It is possible to fully bake in a MN into mocknet as well, but requires further work. For now, MNs need to carried over.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ src/univalue/gen
 # Only ignore unexpected patches
 *.patch
 !depends/patches/**/*.patch
+!contrib/devtools/*.patch
 
 #libtool object files
 *.lo
--- a/contrib/devtools/mocknet.patch
+++ b/contrib/devtools/mocknet.patch
@@ -1,0 +1,38 @@
+diff --git a/src/pos.cpp b/src/pos.cpp
+index 1ad8d066d..01b13b8aa 100644
+--- a/src/pos.cpp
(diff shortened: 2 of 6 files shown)


prasannavl:
I don't think it matters, as the difficulty is unchanged. So they will never be able to mine on the new one unless they have the mining power - it's not theoretically possible for the fork should it ever happen to catch up.

Bushstar (approved):

PR #1288, opened 2022-05-25
Add token split logging context

ghost:
/kind chore

Adds `TOKEN_SPLIT` logging context and moves some logging from loops under the flag.


prasannavl:
This is neat! Let's merge this after 2.8.0 is released (to keep the current verbosity for for 2.8.0 at launch).

prasannavl:
@wafflespeanut are you ok if we close this. This was super useful to during the time - and I think we got many of these into main logs. Is there anything specific from here, you'd like to add, or should we close this PR? 

Think it would be easier to pick what we want than resolve the conflicts.

wafflespeanut:
Let's close this.

PR #1307, opened 2022-05-30
Uses sysctl -n hw.logicalcpu instead of nproc on macOS

Jouzo:

/kind fix

Fixes make.sh on macOS by using `sysctl -n hw.logicalcpu` as default MAKE_JOBS value.


prasannavl (approved):

prasannavl:
Merging for now. But note however: https://gitlab.haskell.org/ghc/ghc/-/issues/8594

Appears that ncpu is deprecated.

PR #1308, opened 2022-05-30
Disable building benchmarks on CI dev builds temporarily

Jouzo:

/kind feature

Disable bench on CI `Dev Build` while bench is unused.

Fixes https://github.com/DeFiCh/ain/blob/5fff31befd37b35581702ae9b9d934ed6ee0721e/src/bench/bench_defi.cpp#L36


prasannavl:
Merging. To be enabled again after re-enabling benchmarks. Curious how long it takes. 
Nice-eye. Let's see.

PR #1311, opened 2022-05-30
Return -1 value on getvault when tokens are locked

hidiego:

/kind fix

Make `getvault` return -1 values instead of 0 when any of the tokens in the vault is frozen.

Added functional tests.

### DIFF
--- a/src/masternodes/rpc_vault.cpp
+++ b/src/masternodes/rpc_vault.cpp
@@ -1,3 +1,4 @@
+#include "masternodes/masternodes.h"
 #include <masternodes/accountshistory.h>
 #include <masternodes/auctionhistory.h>
 #include <masternodes/mn_rpc.h>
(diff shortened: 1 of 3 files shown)


src/masternodes/rpc_vault.cpp
```diff
@@ -141,6 +142,14 @@ namespace {
             collateralRatio = int(rate.val->ratio());
         }
 
+        bool isVaultTokenLocked {false};
+        for (const auto& collateral : collaterals->balances) {
```
Bushstar, line 146: Vault will only lock on a loan token being split.

Bushstar (approved):

prasannavl (approved):

src/masternodes/rpc_vault.cpp
prasannavl: Good call out. Should be harmless for now.

PR #1324, opened 2022-06-03
Remove irrelevant height branching after FCC

prasannavl:
/kind chore

- Remove the unnecessary condition since FCC has kicked in.


Mixa84:
This shouldn't be removed from consensus as it is possible to modify prefork state (for example from bad snapshot or etc.).

The only fork checks that are safe to remove after hardfork kicks in is from RPC and part that are not in consensus!

prasannavl:
How so? The mainnet does not have these TXs. Corrupt snapshot are exactly that - corrupt (different network/fork). It's irrelevant. It's just dead code and extra branches, that's unnecessary to maintain.

src/masternodes/mn_checks.cpp
```diff
@@ -1547,11 +1547,7 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
         CDataStructureV0 liveKey{AttributeTypes::Live, ParamIDs::Economy, EconomyKeys::DFIP2203Current};
         auto balances = attributes->GetValue(liveKey, CBalances{});
 
-        // Can be removed after the hard fork, since it will be backward compatible
-        // but have to keep it around for pre 2.8.0 nodes for now
-        if (height >= static_cast<uint32_t>(consensus.FortCanningCrunchHeight)) {
-            CalculateOwnerRewards(obj.owner);
-        }
+        CalculateOwnerRewards(obj.owner);
```
Bushstar, line 1550: This needs to remain guarded. Otherwise full sync nodes will end up with some accounts that performed a FutureSwap before FCC having a higher balance than others which could lead to consensus issues.

prasannavl: But each action that does require a balance in consensus (expect for this bug), should already be doing is expected to do this action anyway?

Is there anything else I'm missing - otherwise, let's just do a confirmation anyway with full a sync and proceed to remove?

PR #1326, opened 2022-06-04
Add flag to skip token is locked check

hidiego:

/kind fix

Add "skipLockedCheck" optional parameter to `listvaults` rpc, when set to True, return informative collateralValue and loanValue even if token is locked in vault. Defaults to False.


hidiego:
Need to do some fixes on Interest Per Block calculation.

hidiego:
Worked continued in #1552 with cleaner history.

PR #1330, opened 2022-06-06
Add optional `format` flag for `listaccounthistory`

hidiego:

/kind feature

Adds optional parameter `format` to `listaccounthistory` with following possible values:
- 'id' -> "amounts@id" (i.e "10@0") 
- 'symbol' (default) -> "amounts@symbol" (i.e "10@DFI")


src/masternodes/rpc_accounts.cpp
```diff
@@ -968,6 +984,9 @@ UniValue listaccounthistory(const JSONRPCRequest& request) {
                                   "Maximum number of records to return, 100 by default"},
                                  {"txn", RPCArg::Type::NUM, RPCArg::Optional::OMITTED,
                                   "Order in block, unlimited by default"},
+                                 {"amountFormat", RPCArg::Type::STR, RPCArg::Optional::OMITTED,
+                                  "Return amounts with the following: (default)'id' -> <amount>@id; 'symbol' -> <amount>@symbol"},
```
chanakasameera, line 988: Here default is 'symbol', right ?

hidiego: That is right, nice catch!

src/masternodes/mn_rpc.h
```diff
@@ -35,6 +35,16 @@ typedef enum {
     SelectionPie,
 } AccountSelectionMode;
 
+enum class AmountFormat : uint8_t {
```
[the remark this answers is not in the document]
prasannavl: Don't complicate it when there's no need.

prasannavl (approved):

PR #1338, opened 2022-06-09
[WIP] Add index to pagination in listauctionhistory

hidiego:

/kind fix

### DIFF
--- a/src/masternodes/rpc_vault.cpp
+++ b/src/masternodes/rpc_vault.cpp
@@ -1214,7 +1214,9 @@ UniValue listauctionhistory(const JSONRPCRequest& request) {
         if (isMine && !(IsMineCached(*pwallet, key.owner) & ISMINE_SPENDABLE)) {
             return true;
         }
-
+        if (key.index < start.index){
+            return true;
+        }
         ret.push_back(auctionhistoryToJSON(key, valueLazy.get()));
 
         return --limit != 0;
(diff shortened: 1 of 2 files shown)


src/masternodes/rpc_vault.cpp
```diff
@@ -1217,7 +1217,9 @@ UniValue listauctionhistory(const JSONRPCRequest& request) {
         if (isMine && !(IsMineCached(*pwallet, key.owner) & ISMINE_SPENDABLE)) {
             return true;
         }
-
+        if (key.index < start.index){
+            return true;
+        }
```
surangap, line 1222: is this necessary? this will create a filter effect. https://github.com/JellyfishSDK/jellyfish/pull/1526#issuecomment-1147116680
listauctionhistory should provide filtration only for `owner|vaultId` from a given pagination point.
so `blockHeight, index` should only be for pagination.

test/functional/feature_loan.py
```diff
@@ -231,7 +231,9 @@ def run_test(self):
         assert_equal(auctions[1]['auctionBid'], "515.00000000@TSLA")
         assert_equal(auctions[2]['auctionBid'], "555.50000000@TSLA")
 
-        auctionsV = self.nodes[0].listauctionhistory(vaultId1)
+        auctionsV = self.nodes[0].listauctionhistory(vaultId1, {"index": 1})
```
surangap, line 234: related to https://github.com/DeFiCh/ain/pull/1338/files#r899757282,
This query asks for all the subsequent records from the pagination start point AuctionHistoryKey(0,{},{},1)

surangap (changes requested):
https://github.com/DeFiCh/ain/pull/1338/files#r899757282
better to keep it the way it is if this change is not required for any special requirement.

PR #1364, opened 2022-06-30
Consortium for minting DAT tokens backport

Mixa84:

/kind feature


shohamc1:
Superseded by #1504

PR #1375, opened 2022-07-07
Change `interetPerBlock` and `interestPerBlockValue` to precision string

hidiego:
#### What kind of PR is this?:

/kind fix

#### What this PR does / why we need it:

Adds precision string value to interestPerBlock introduced after FCH update in `getvault` + verbose flag enabled

PR #1386, opened 2022-07-18
Negative interest rates

hidiego:

/kind feature

Allows negative interest on loan tokens

#### Additional comments?:
This PR has merged commit 7ce7530c2f9a18b499856ab2b0c0bfec53f8babf  by @Bushstar
Also has PR #1375 that fixes interest per block display bug on `getvault` (needed for testing)

### DIFF
--- a/src/masternodes/loan.cpp
+++ b/src/masternodes/loan.cpp
@@ -210,6 +210,9 @@ inline T InterestPerBlockCalculationV1(CAmount amount, CAmount tokenInterest, CA
 inline base_uint<128> InterestPerBlockCalculationV2(CAmount amount, CAmount tokenInterest, CAmount schemeInterest)
 {
     auto netInterest = (tokenInterest + schemeInterest) / 100; // in %
+    if (netInterest < 0) {
+        return 0;
+    }
     static const auto blocksPerYear = 365 * Params().GetConsensus().blocksPerDay();
     return arith_uint256(amount) * netInterest * COIN / blocksPerYear;
 }
--- a/src/masternodes/loan.cpp
+++ b/src/masternodes/loan.cpp
@@ -253,20 +253,7 @@ CNegativeInterest TotalInterestCalculation(const CInterestRateV3& rate, const ui
 {
     const auto totalInterest = (height - rate.height) * rate.interestPerBlock.amount;
 
-    CNegativeInterest interest;
-    if ((!rate.interestToHeight.negative && !rate.interestPerBlock.negative) ||
-        (rate.interestToHeight.negative && rate.interestPerBlock.negative)) {
-        interest.amount = rate.interestToHeight.amount + totalInterest;
-        interest.negative = rate.interestPerBlock.negative && rate.interestToHeight.negative;
-    } else {
-        if (rate.interestToHeight.amount > totalInterest) {
-            interest.amount = rate.interestToHeight.amount - totalInterest;
-            interest.negative = rate.interestToHeight.negative;
-        } else {
-            interest.amount = totalInterest - rate.interestToHeight.amount;
-            interest.negative = !rate.interestToHeight.negative;
-        }
-    }
+    const auto interest = InterestAddition(rate.interestToHeight, {rate.interestPerBlock.negative, totalInterest});
 
     LogPrint(BCLog::LOAN, "%s(): CInterestRate{.height=%d, .perBlock=%d, .toHeight=%d}, height %d - totalInterest %d\n",
         __func__,
--- a/src/masternodes/loan.cpp
+++ b/src/masternodes/loan.cpp
@@ -186,7 +186,7 @@ void CLoanView::EraseDelayedDestroyScheme(const std::string& loanSchemeID)
     EraseBy<DestroyLoanSchemeKey>(loanSchemeID);
 }
 
-std::optional<CInterestRateV3> CLoanView::GetInterestRate(const CVaultId& vaultId, DCT_ID id, uint32_t height)
+std::optional<CInterestRateV3> CLoanView::GetInterestRate(const CVaultId& vaultId, const DCT_ID id, const uint32_t height)
 {
     if (height >= static_cast<uint32_t>(Params().GetConsensus().GreatWorldHeight)) {
         return ReadBy<LoanInterestV3ByVault, CInterestRateV3>(std::make_pair(vaultId, id));
--- a/src/masternodes/loan.cpp
+++ b/src/masternodes/loan.cpp
@@ -254,16 +254,17 @@ CNegativeInterest TotalInterestCalculation(const CInterestRateV3& rate, const ui
     const auto totalInterest = (height - rate.height) * rate.interestPerBlock.amount;
 
     CNegativeInterest interest;
-    if ((!rate.interestPerBlock.negative && !rate.interestToHeight.negative) ||
-        (rate.interestPerBlock.negative && rate.interestToHeight.negative)) {
+    if ((!rate.interestToHeight.negative && !rate.interestPerBlock.negative) ||
+        (rate.interestToHeight.negative && rate.interestPerBlock.negative)) {
         interest.amount = rate.interestToHeight.amount + totalInterest;
         interest.negative = rate.interestPerBlock.negative && rate.interestToHeight.negative;
     } else {
         if (rate.interestToHeight.amount > totalInterest) {
             interest.amount = rate.interestToHeight.amount - totalInterest;
+            interest.negative = rate.interestToHeight.negative;
         } else {
             interest.amount = totalInterest - rate.interestToHeight.amount;
-            interest.negative = !rate.interestToHeight.negative && rate.interestPerBlock.negative;
+            interest.negative = !rate.interestToHeight.negative;
         }
     }
 
--- a/src/masternodes/loan.cpp
+++ b/src/masternodes/loan.cpp
@@ -232,7 +232,7 @@ CNegativeInterest InterestPerBlockCalculationV3(CAmount amount, CAmount tokenInt
 
 CAmount CeilInterest(const base_uint<128>& value, uint32_t height)
 {
-    if (int(height) >= Params().GetConsensus().FortCanningHillHeight) {
+    if (height >= static_cast<uint32_t>(Params().GetConsensus().FortCanningHillHeight)) {
         CAmount amount = (value / base_uint<128>(HIGH_PRECISION_SCALER)).GetLow64();
         amount += CAmount(value != base_uint<128>(amount) * HIGH_PRECISION_SCALER);
         return amount;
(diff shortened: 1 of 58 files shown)


src/masternodes/loan.cpp
```diff
@@ -227,11 +240,29 @@ CAmount CeilInterest(const base_uint<128>& value, uint32_t height)
...
+CNegativeInterest TotalInterestCalculation(const CInterestRateV3& rate, uint32_t height)
+{
+    CNegativeInterest interest;
+    if (rate.interestPerBlock.negative) {
+        // TODO this formula is inaccurate, it assumes that the base amount will remain the same,
+        // when with a negative interest rate the loan amount should reduce on each block. Required
+        // is a formula to reduce the loan amount on each height by the negative interest per block,
+        // and then update the interest per block for the next height, repeat until we are at the
+        // current height. This requires that the loan amount be passed to this function which was
+        // until now not required as the loan amount was static before and so therefore was the interest.
```
kuegi, line 252: noob question: why? positive interest didn't increase the base amount, so why should negative interest decrease it?
why not keep it simple and have negative interest in the vault. if `loan + interest <= 0` the loan is removed, so only this check is necessary, but can be done outside.

hidiego: Firstly interests are reduced and only after interests get to 0 loan starts reducing. This cannot happen with positive interest as loan and interests of the loan are considered differently. 

Examples:

- Positive interest

>  Lets asume +0,1 positive interest, then;
>  Loan amount -> 1Token
>  After 1 block -> Loan amount = 1Token; Interest of loan = 0,1Token; Total loan = 1,1Token

- Negative interest

> Asume -0,1 positive interest in a vault with already 0,2Token of interest accrued in the past
> Loan amount -> 1Token; Total loan -> 1,2Token
> After 1 block -> Loan amount = 1Token; Interest of loan = 0,2-0,1Token = 0,1Token; Total loan = 1,1Token
> After 1 block -> Loan amount = 1Token; Interest of loan = 0,1-0,1Token = 0Token; Total loan = 1Token
> After 1 block -> Loan amount = 1-0,1Token = 0,9Token; Interest of loan = 0Token; Total loan = 0,9Token
> After 1 block -> Loan amount = 0,9-0,1Token = 0,8Token; Interest of loan = 0Token; Total loan = 0,8Token

hidiego: Will write a set of tests illustrating the above

kuegi: imho the interest value should become negative and not the loan value be reduced. in the positive interest rate we also don't increase the loan when interest accumulates.
Would also be far easier for users to understand when loan value stays the same (unless there is a payback or takeloan) but interest value increases or decreases (or even goes negative).

(hmm i got notification of a comment, but github is not showing it. my internet bad?)

src/masternodes/mn_checks.cpp
```diff
@@ -2926,8 +2925,26 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                if (!rate) {
+                    return Res::Err("Cannot get interest rate for this token (%d)", tokenId.v);
+                }
+                const auto scheme = mnview.GetLoanScheme(vault->schemeId);
+                assert(scheme);
+                const auto tokenInterest = mnview.GetLoanTokenByID(tokenId);
+                assert(tokenInterest);
+                const auto totalInterest = TotalInterest(*rate, height, amount, tokenInterest->interest, scheme->rate);
+                if (totalInterest < 0 && amount + totalInterest <= 0) {
+                    res = mnview.SubMintedTokens(tokenId, amount);
```
kuegi, line 2940: This is not correct. The token was not taken out of circulation, so it should not be reduced from the minted amount.
negative interest rates reduce the loans, but not the minted token amount.

src/masternodes/mn_checks.cpp
```diff
@@ -2992,6 +3008,38 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+                        const auto loanToken = mnview.GetLoanTokenByID(tokenId);
+                        assert(loanToken);
+                        if (const auto totalInterest = TotalInterest(*rate, height, tokenAmount, loanToken->interest, scheme->rate); totalInterest < 0) {
+                            const auto subAmount = tokenAmount > std::abs(totalInterest) ? std::abs(totalInterest) : tokenAmount;
+                            res = mnview.SubLoanToken(obj.vaultId, CTokenAmount{tokenId, subAmount});
+                            if (!res) {
+                                return res;
+                            }
+
+                            res = mnview.SubMintedTokens(tokenId, subAmount);
```
kuegi, line 3029: those tokens are still in circulation, so they must not be substracted from the minted amount. we only reduced the loan amount. reducing the minted tokens here would add "algorithmic" DUSD that are not accounted for anywhere.

src/masternodes/mn_checks.cpp
```diff
@@ -3361,56 +3445,61 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
                         return Res::Err("Cannot payback this amount of loan for %s, either payback full amount or less than this amount!", loanToken->symbol);
                 }
 
                 CalculateOwnerRewards(obj.from);
 
                 if (paybackTokenId == loanTokenId)
                 {
-                    res = mnview.SubMintedTokens(loanTokenId, subLoan);
+                    // If interest was negative remove it from sub amount
+                    res = mnview.SubMintedTokens(loanTokenId, subInterest > 0 ? subLoan : subLoan + subInterest);
```
kuegi, line 3502: again: negative interest does not reduce minted tokens. only the subLoan has to be substracted from minted tokens!

kuegi, line 3502: only the paybackAmount (coins that the user is sending in) is to be substracted from the minted tokens. interest was not minted and negative interest does not substract minted tokens.

src/masternodes/loan.cpp
```diff
@@ -241,66 +249,119 @@ static base_uint<128> ToHigherPrecision(CAmount amount, uint32_t height)
...
+            if (loanAmountHP > reductionCeil) {
+                loanAmountHP -= reductionCeil;
+                reduction += reductionCeil;
+            } else {
+                // Loan amount reduced to zero
+                reduction += loanAmountHP;
+                break;
+            }
+            reductionRate = InterestPerBlockCalculationV3(CeilInterest(loanAmountHP, std::numeric_limits<uint32_t>::max()), tokenInterest, schemeInterest);
+        }
```
kuegi, line 285: I have to ask again: why doing this complicated calculation which makes everything harder (calculations, explaining it to users) and acts as the worst case for users?

positive interest is on purpose (for performance reasions) not compounding:
loan of 1 token, interestRatePerBlock = 0.1 (based on this 1 token loan) we get 
block 0 : loan 1 , interest 0
block 1: loan 1, interest 0.1
block 2: loan 1, interest 0.2
... easy to explain, easy to calculate

negative interest should (IMHO) be the same logic. interestPerBlock = -0.1
block 0: loan 1, interest 0
block 1: loan 1, interest -0.1
block 2: loan 1, interest -0.2
....
at block 10 we see loan + interest <= 0 and we auto close the loan. But until then we accumulate the negative interest based on the loan. 

The way it currently is implemented, there is a different interest effect if you opened a 1 token loan 1 month ago and accumulated positive interest and a fresh loan of 1 token:
with existing interest on the loan you profit, because the negative rate is calculated from the loan, not loan + interest, but the negative interest first reduces your existing interest (so for longer time you get higher negative interestPerBlock).
while for the fresh loan, the negative interest reduces your loan value and therefore the negativeInterestPerBlock.

This will be super hard to explain, specially since its bad for the user and now different to how it was done for positive interest.

src/masternodes/mn_checks.cpp
Bushstar: Thanks for spotting this, you are correct.

src/masternodes/mn_checks.cpp
Bushstar: Resolved.

src/masternodes/mn_checks.cpp
Bushstar:
```
auto subLoan = paybackAmount - subInterest;
```

In the case of subInterest being negative it is added to the subLoan amount increasing the subLoan amount. The above commented line...

```
mnview.SubMintedTokens(loanTokenId, subInterest > 0 ? subLoan : subLoan + subInterest);
```

Then reduces subLoan by adding the negative subInterest amount so that negative interest is not reduced from the minted amount.

src/masternodes/loan.cpp
Bushstar: This will make life a lot easier, not just to explain to the users but on the implementation front. Spoke to the team about this and will start updating now.

src/masternodes/loan.h
```diff
@@ -260,17 +300,62 @@ inline CInterestRateV2 ConvertInterestRateToV2(const CInterestRate& rate1)
...
+{
+    CInterestRateV3 rate3{};
+    rate3.height = rate2.height;
+    rate3.interestPerBlock = {false, rate2.interestPerBlock};
+    rate3.interestToHeight = {false, rate2.interestToHeight};
+
+    return rate3;
+}
+
+inline const auto InterestCalculation = [](CNegativeInterest &a, const CNegativeInterest &b) {
```
kuegi, line 333: minor: "InterestCalculation" sounds like a bad wording here. Its "sum interest into first argument" right?

src/masternodes/loan.h
```diff
@@ -260,17 +300,62 @@ inline CInterestRateV2 ConvertInterestRateToV2(const CInterestRate& rate1)
...
+    if ((!a.negative && !b.negative) ||
+        (a.negative && b.negative)) {
+        a.amount += b.amount;
+    } else {
+        if (a.amount > b.amount) {
+            a.amount -= b.amount;
+        } else {
+            a.amount = b.amount - a.amount;
+            a.negative = !a.negative && b.negative;
```
kuegi, line 342: effectively `a.negative = b.negative`  would make it also easier to read

src/masternodes/loan.h
```diff
@@ -260,17 +300,62 @@ inline CInterestRateV2 ConvertInterestRateToV2(const CInterestRate& rate1)
...
+    rate3.height = rate2.height;
+    rate3.interestPerBlock = {false, rate2.interestPerBlock};
+    rate3.interestToHeight = {false, rate2.interestToHeight};
+
+    return rate3;
+}
+
+inline const auto InterestCalculation = [](CNegativeInterest &a, const CNegativeInterest &b) {
+    if ((!a.negative && !b.negative) ||
+        (a.negative && b.negative)) {
```
kuegi, line 335: why not `a.negative == b.negative` ?

src/masternodes/loan.h
Bushstar: Right.

src/masternodes/loan.h
Bushstar: Agreed.

src/masternodes/loan.h
Bushstar: Also agreed. I wrote this thing out in full first for all cases and then condensed it but failed to simplify it further as you've noted. Many thanks for taking the time to review the changes so far.

PR #1392, opened 2022-07-19
Fix swap when direct path disabled

Bushstar:


kuegi:
@Bushstar same story for composite, right? also doesn't check for active pool and easily chooses an disabled path as "best"

PR #1398, opened 2022-07-26
Prevent SPV closing sockets in use

Bushstar:


src/spv/bitcoin/BRPeer.cpp
```diff
@@ -833,142 +826,37 @@ static int _BRPeerAcceptMessage(BRPeer *peer, const uint8_t *msg, size_t msgLen,
...
-    if (error && err) *error = err;
-    return r;
-}
-
 static int _peerCheckAndGetSocket (BRPeerContext *ctx, SOCKET *socket) {
     int exists;
 
     ctx->lock.lock();
-    exists = ctx->socket != INVALID_SOCKET;
+    exists = ctx->socket >= 0;
```
prasannavl, line 859: On Windows, invalid socket is 0, while on other platforms, it's >=0.

prasannavl: Since windows does not real use fds, rather kernel objects and handles. Believe 0 is error and anything non zero is valid.

prasannavl: Looks like Bitcoin code, the hypothesis I'm forming from a quick glance is that, it assumes that Linux fds for these sockets will never be 0 (since that's stdin) and as such uses Window's 0 as the SOCKET_INVALID, and then on each failure, sets it manually to SOCKET_INVALID to ensure it's consistent. Perhaps, we should do the same?

Jouzo (approved):

src/spv/bitcoin/BRPeer.cpp
```diff
@@ -1035,17 +934,10 @@ static void *_peerThreadRoutine(void *arg)
             len = 0;
 
             while (socket != INVALID_SOCKET && ! error && len < HEADER_LENGTH) {
-//                n = read(socket, &header[len], sizeof(header) - len);
                 n = recv(socket, (char*)&header[len], sizeof(header) - len, 0);
                 if (n > 0) len += n;
                 if (n == 0) error = ECONNRESET;
                 if (n < 0 && WSAGetLastError() != WSAEWOULDBLOCK) error = WSAGetLastError();
-                if (error == WSAENOTSOCK) {
```
prasannavl, line 1043: Is there a specific reason we're removing this? 

The situation I have in mind of occurring is if the OS is forced to closed a socket for whatever reason, and then this handle is reused for something else (in which case the right behaviour for us is indeed to ignore).

src/spv/bitcoin/BRPeer.cpp
```diff
@@ -1092,17 +984,10 @@ static void *_peerThreadRoutine(void *arg)
...
                         n = recv(socket, (char *) &payload[len], msgLen - len, 0);
                         if (n > 0) len += n;
                         if (n == 0) error = ECONNRESET;
                         if (n < 0 && WSAGetLastError() != WSAEWOULDBLOCK) error = WSAGetLastError();
-                        if (error == WSAENOTSOCK) {
-                            ctx->lock.lock();
-                            socket = ctx->socket = INVALID_SOCKET;
-                            ctx->lock.unlock();
-                            continue;
-                        }
```
prasannavl, line 1105: Same as above.

prasannavl (commented):
Final points to review. LGTM after resolving this.

src/spv/bitcoin/BRPeer.cpp
prasannavl: Resolving, since worst case, it'll log error and as pointed out by Peter that this is closer to upstream.

PR #1400, opened 2022-07-27
On-chain governance updates

Mixa84:

/kind feature

- Introduce separate attribute type for on-chain governance `(v0/governance)`
- Activate on-chain governance transactions and events on attribute: `v0/governance/global/enabled`
- Added attribute which can disable automatic payout of CFP and leave it to be done manually.
- Added context hash to CFP


Mixa84:
Closing this as it will be added to #1441

PR #1403, opened 2022-08-01
Track values for automation

Bushstar:


src/masternodes/mn_checks.cpp
```diff
@@ -3072,22 +3072,27 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
         if (!collaterals)
             return Res::Err("Vault with id %s has no collaterals", obj.vaultId.GetHex());
 
+        auto attributes = mnview.GetAttributes();
+        assert(attributes);
+        CDataStructureV0 loanKey{AttributeTypes::Live, ParamIDs::Economy, EconomyKeys::Loans};
+        auto balances = attributes->GetValue(loanKey, CBalances{});
```
kuegi, line 3078: for easier reading: could we rename that to "loanAttributeBalances"?

the `balances.add(loanAmount)` looks like there is something added to the account balance etc. Feels like a future-refactor-error in the making.

src/masternodes/mn_checks.cpp
```diff
@@ -3115,14 +3120,18 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+
+            res = mnview.AddMintedTokens(tokenId, tokenAmount);
             if (!res)
                 return res;
 
             const auto& address = !obj.to.empty() ? obj.to
                                                   : vault->ownerAddress;
             CalculateOwnerRewards(address);
-            res = mnview.AddBalance(address, CTokenAmount{kv.first, kv.second});
+            res = mnview.AddBalance(address, CTokenAmount{tokenId, tokenAmount});
```
kuegi, line 3134: why not reuse loanAmount?

src/masternodes/mn_checks.cpp
```diff
@@ -3212,21 +3227,23 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
 
         auto shouldSetVariable = false;
         auto attributes = mnview.GetAttributes();
+        assert(attributes);
+        CDataStructureV0 loanKey{AttributeTypes::Live, ParamIDs::Economy, EconomyKeys::Loans};
+        auto balances = attributes->GetValue(loanKey, CBalances{});
```
kuegi, line 3232: same as above regarding the name.

src/masternodes/mn_checks.cpp
```diff
@@ -3341,13 +3356,17 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
 
                 if (paybackTokenId == loanTokenId)
                 {
+                    if (loanToken->symbol == "DUSD") {
+                        balances.Sub(tokenAmount);
```
kuegi, line 3360: why only for DUSD? I know we only need it for DUSD, but would it be a problem to store all loans right away. I think this would be benefical for statistics etc.

prasannavl: Hi @kuegi - fair question! also the goal exactly as you indicated to enable more and more statistics on the live namespace.  But currently, the attributes subsystem is pending revamp of internal database. Currently, it's hanging off an implementation that wasn't designed to handle the requirements for such large statistics. So, we're just enabling what's needed for the moment, until attributes are moved to dedicated database, which is in the works.

PR #1415, opened 2022-08-10
Update secp256k1 from upstream

Jouzo:

/kind chore

Updates secp256k1 from upstream


Jouzo:
Closed for #1417

PR #1429, opened 2022-08-30
Add `logstoredinterests`

prasannavl:
/kind feature

- To list all interest rates and loans in vaults

### DIFF
--- a/src/masternodes/rpc_vault.cpp
+++ b/src/masternodes/rpc_vault.cpp
@@ -1937,7 +1937,7 @@ UniValue logstoredinterests(const JSONRPCRequest& request) {
     LOCK(cs_main);
 
     auto height = ::ChainActive().Height();
-    typedef std::tuple<DCT_ID,CAmount,CInterestRateV3> VaultInfo;
+    using VaultInfo = std::tuple<DCT_ID,CAmount,CInterestRateV3>;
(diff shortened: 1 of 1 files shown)


src/masternodes/rpc_vault.cpp
```diff
@@ -1911,6 +1911,72 @@ UniValue getstoredinterest(const JSONRPCRequest& request) {
...
+               },
+    }.Check(request);
+
+    if (auto res = GetRPCResultCache().TryGet(request)) return *res;
+
+    LOCK(cs_main);
+
+    auto height = ::ChainActive().Height();
+    using VaultInfo = std::tuple<DCT_ID,CAmount,CInterestRateV3>;
+    std::map<std::string, std::vector<VaultInfo>> items;
```
hidiego, line 1941: Why not add directly to UniValue and avoid iterating again after?

prasannavl: Please don't do that at all, unless there is a specific reason to do so. The extra memory by the raw dataset, compared to UniValue is usually small. And so is the additional computation that they are irrelevant. Code clarity, reusability and maintenance is usually far more important to focus on. This decouples the formatting from the actual logic. 

Second, the format could something else that's not JSON at all (GRPC, for instance). Mixing it up with UniValue makes things messy.

PR #1434, opened 2022-08-31
feat(dmchandler): add DMCHandler Implementation.

DieHard073055:

/kind feature

Adds implementation of the functions to integrate DMC with DefiChain.

- InitializeRPCClient
- AddDMCPayloadToNativeBlock:
To request the DMC node for the new block data which will be added to the native chain block.
- ConnectPayloadToDMC
To request the DMC node to add the given block data as its next block.

Fixes #

#### Additional comments?:
Its still WIP. Currently the code is implemented using RPC, we havent finished the discussion on what should be used on the actual implementation.
Theres still some TODO(s) to be taken care of.

### DIFF
--- a/src/DMCHandler.cpp
+++ b/src/DMCHandler.cpp
@@ -1,16 +1,31 @@
 #include <DMCHandler.h>
 #include <logging.h>
 
+#include "rpc/client.h"
+
 #include <iostream>
 
-bool DMCHandler::AddDMCPayloadToNativeBlock(std::shared_ptr<CBlock> block) {
-  // Dummy imp for now
-  block->dmcPayload = std::vector<unsigned char>({'\x44', '\x4D', '\x43', '\x20', '\x52', '\x4f', '\x43', '\x4b', '\x53'});
+void DMCHandler::InitializeRPCClient(){
+  // Add the correct port number
+  this->DMCNode("localhost", 8080);
+}
+
+bool DMCHandler::AddDMCPayloadToNativeBlock(std::shared_ptr<CBlock> block, std::vector<DMCTx> txN) {
+  // Get DMC node to mint the new block
+  EncodedDMCBlock newDMCBlock = nullptr;
+  newDMCBlock = (this->DMCNode).call("mintBlock", txN);
+  if(newDMCBlock == nullptr){
+    // TODO: add exponential backoff
+    newDMCBlock = (this->DMCNode).call("getBlock");
+  }
+  block->dmcPayload = newDMCBlock;
+  
   return true;
 }
 
 bool DMCHandler::ConnectPayloadToDMC(const std::vector<unsigned char>& payload) {
   // Dummy imp for now
   LogPrintf("DMC Payload: [%s]", std::string(payload.begin(), payload.end()));
+  (this->DMCNode).call("connectBlock", payload);
   return true;
 }
--- a/src/DMCHandler.h
+++ b/src/DMCHandler.h
@@ -3,11 +3,52 @@
 
 #include <iostream>
 #include <primitives/block.h>
+#include "rpc/client.h"
+
+class DFITx {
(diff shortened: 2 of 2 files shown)


src/DMCHandler.cpp
```diff
@@ -1,16 +1,31 @@
...
+#include "rpc/client.h"
+
 #include <iostream>
 
-bool DMCHandler::AddDMCPayloadToNativeBlock(std::shared_ptr<CBlock> block) {
-  // Dummy imp for now
-  block->dmcPayload = std::vector<unsigned char>({'\x44', '\x4D', '\x43', '\x20', '\x52', '\x4f', '\x43', '\x4b', '\x53'});
+void DMCHandler::InitializeRPCClient(){
+  // Add the correct port number
+  this->DMCNode("localhost", 8080);
```
surangap, line 10: can get the params from config.

src/DMCHandler.h
```diff
@@ -3,11 +3,52 @@
...
+};
+
+struct EncodedDMCBlock {
+    std::vector<unsigned char> blockdata;
+} EncodedDMCBlock;
 
 // WIP
 class DMCHandler {
+private:
+    rpc::client DMCNode;
```
surangap, line 47: make this injectable. you can either pass it in the constructor or another method you prefer.
Then we can use a mock object of `DMCNode` to isolate from the DMC side and write more functional tests of its own on the native side.

src/DMCHandler.cpp
```diff
@@ -1,16 +1,31 @@
...
+  if(newDMCBlock == nullptr){
+    // TODO: add exponential backoff
+    newDMCBlock = (this->DMCNode).call("getBlock");
+  }
+  block->dmcPayload = newDMCBlock;
+  
   return true;
 }
 
 bool DMCHandler::ConnectPayloadToDMC(const std::vector<unsigned char>& payload) {
```
surangap, line 26: need to pass `DMCTx`s with this also. refer https://birthday.atlassian.net/wiki/spaces/DMC/pages/41517081/Bridging+Assets+WIP

src/DMCHandler.h
```diff
@@ -3,11 +3,52 @@
...
+
+class DFITx {
+    private:
+    char* from; // TODO: get correct DMC Address type
+    char* to; // TODO: get correct DFI address type
+    int64_t amount;
+    char* signature;
+
+public:
+    DFITx(char* from, char* to, int64_t amount, char* signature){
```
surangap, line 16: maybe check the types used in the native side RPCs to store hash values and use the same

src/DMCHandler.cpp
```diff
@@ -1,16 +1,31 @@
...
-  block->dmcPayload = std::vector<unsigned char>({'\x44', '\x4D', '\x43', '\x20', '\x52', '\x4f', '\x43', '\x4b', '\x53'});
+void DMCHandler::InitializeRPCClient(){
+  // Add the correct port number
+  this->DMCNode("localhost", 8080);
+}
+
+bool DMCHandler::AddDMCPayloadToNativeBlock(std::shared_ptr<CBlock> block, std::vector<DMCTx> txN) {
+  // Get DMC node to mint the new block
+  EncodedDMCBlock newDMCBlock = nullptr;
+  newDMCBlock = (this->DMCNode).call("mintBlock", txN);
```
surangap, line 16: `mintBlock()` should return the `payload` and `DMCTx[]`. and this `DMCTx[]` array should be used by the native side to settle balances for `DMC -> Native` transfers

surangap (commented):
Added few comments.
make the `DMCNode` injectible and check the possibility of using a mock object with boost testing. write a few tests for this module.

we will soon need a cross-chain signature verification module(separate PR) to validate things that are coming from the DMC side. Added it to action items

PR #1442, opened 2022-09-04
Add merge tests and enhance tests framework with helper functions

hidiego:
/kind refactor

### DIFF
--- a/test/functional/feature_poolswap.py
+++ b/test/functional/feature_poolswap.py
@@ -23,24 +23,17 @@ from math import trunc
 
 class PoolPairTest (DefiTestFramework):
     def set_test_params(self):
-        self.num_nodes = 4
+        self.num_nodes = 3
         # node0: main (Foundation)
-        # node3: revert create (all)
-        # node2: Non Foundation
         self.setup_clean_chain = True
         self.extra_args = [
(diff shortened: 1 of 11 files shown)


test/functional/feature_poolswap.py
```diff
@@ -537,14 +529,10 @@ def update_comission_and_fee_to_1pct_pool2(self):
         assert_equal(attributes['v0/live/economy/dex/%s/fee_burn_a'%(self.idBL)], Decimal(str(round(dexoutfee, 8))))
 
     def revert_to_initial_state(self):
```
prasannavl, line 531: Why is something like back to initial state being rolled back 30 blocks? Avoid calling key helpers inaccurately only for a specific context - better to rename to something more accurate or back to reset the node to it's initial position by restarting it.

test/functional/test_framework/test_framework.py
```diff
@@ -404,6 +404,25 @@ def import_deterministic_coinbase_privkeys(self):
             n.importprivkey(privkey=n.get_genesis_keys().ownerPrivKey, label='coinbase', rescan=True)
             n.importprivkey(privkey=n.get_genesis_keys().operatorPrivKey, label='coinbase', rescan=True)
 
+    # rollback one node (Default = node 0)
+    def rollback_to_node(self, block, node=0):
```
prasannavl, line 408: Call it `_rollback_to` so it's understood to be private.

PR #1451, opened 2022-09-08
Allow payback of DUSD with vault's collateral

Jouzo:

/kind feature

Allow payback of DUSD with vault's collateral


prasannavl:
TODO: 

- After the paying back with collateral, add checks if the vault will enter liquidation in the post change state. If it does, TX should be rejected.
- Add interest burn as another gov var. If it's active, it burns/discards negative interests, otherwise, it reduces that from collateral as well. So, this becomes a governance option than hard coded.
- [Pending] Payback TX. 
 
 Looks good otherwise.

src/masternodes/mn_checks.cpp
```diff
@@ -3083,6 +3097,93 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        // vault under liquidation
+        if (vault->isUnderLiquidation)
+            return Res::Err("Cannot payback vault with collateral while vault's under liquidation");
+
+        // owner auth
+        if (!HasAuth(vault->ownerAddress))
+            return Res::Err("tx must have at least one input from token owner");
+
+        if (!IsVaultPriceValid(mnview, obj.vaultId, height))
+            return Res::Err("Cannot payback vault with collateral while any of the asset's price is invalid");
```
kuegi, line 3131: why not? we can always payback loans with same coin, cause no oracle price needed.

src/masternodes/mn_checks.cpp
```diff
@@ -3083,6 +3097,93 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        }
+
+        res = mnview.SubLoanToken(obj.vaultId, subLoanAmount);
+        if (!res)
+            return res;
+
+        res = mnview.SubVaultCollateral(obj.vaultId, subCollateralAmount);
+        if (!res)
+            return res;
+
```
kuegi, line 3182: missing the sub from minted amount. subCollateralAmount is taken out of circulation -> needs to reduce the total minted amount.

src/masternodes/mn_checks.cpp
```diff
@@ -3083,6 +3097,93 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        if (!rate)
+            return Res::Err("Cannot get interest rate for this token (DUSD)!");
+        auto subInterest = TotalInterest(*rate, height);
+
+        // Edge case where interest is greater than collateral
+        if (subInterest > collateralDUSD) {
+            res = mnview.SubVaultCollateral(obj.vaultId, {dUsdToken->first, collateralDUSD});
+            if (!res)
+                return res;
+
```
kuegi, line 3161: missing the sub from minted amount. subCollateralAmount is taken out of circulation -> needs to reduce the total minted amount.

src/masternodes/mn_checks.cpp
```diff
@@ -4348,6 +4393,105 @@ bool IsVaultPriceValid(CCustomCSView& mnview, const CVaultId& vaultId, uint32_t
...
 }
 
+bool IsPaybackWithCollateral(CCustomCSView& view, const std::map<DCT_ID, CBalances>& loans) {
+    auto tokenDUSD = view.GetToken("DUSD");
+    if (!tokenDUSD)
+        return false;
+
+    if (loans.size() == 1
+    && loans.count(tokenDUSD->first)
+    && loans.at(tokenDUSD->first) == CBalances{{{tokenDUSD->first, 999999999999999999LL}}}) {
```
kuegi, line 4403: is this ever true? looks weird.

Jouzo: Yes, this allows two ways to payback with collateral. One with the new `paybackwithcollateral` RPC and one with `paybackloan` RPC and "9999999999.99999999@DUSD" passed as amounts.
This eases and speeds up the integration by downstream users by bypassing the need to support a new RPC and TX message.

kuegi: how can I resolve discussions? do I not have the rights or am I not finding the button? 🤔

src/masternodes/mn_checks.cpp
Jouzo: Fixed in https://github.com/DeFiCh/ain/pull/1451/commits/99161a1ce7889ace1d42f260d4fc90c64774baff

src/masternodes/mn_checks.cpp
Jouzo: Fixed in https://github.com/DeFiCh/ain/pull/1451/commits/99161a1ce7889ace1d42f260d4fc90c64774baff

src/masternodes/mn_checks.cpp
Jouzo: Must be a right issue. Let me resolve it.

src/masternodes/mn_checks.cpp
prasannavl: Let me answer this on the main comments for better visibility later.

sambP:
It's currently not possible to get some information over the RPC about what's really happening in the vault, right? It would be nice to get an entry in the listvaulthistory.

Is this what @prasannavl means with:

> [Pending] Payback TX.

PR #1455, opened 2022-09-11
Burn interest on PaybackWithCollateral

Jouzo:

/kind fix

Burn interest on PaybackWithCollateral


Bushstar (approved):

src/masternodes/mn_checks.cpp
```diff
@@ -4527,12 +4528,23 @@ Res PaybackWithCollateral(CCustomCSView& view, const CVaultData& vault, const CV
...
     } else {
+        // Postive interest: Loan + interest > collateral. 
+        // Negative interest: Loan - abs(interest) > collateral. 
         if (loanDUSD + subInterest > collateralDUSD) {
             subLoanAmount = collateralDUSD - subInterest;
             subCollateralAmount = collateralDUSD;
+
+            // Negative interest safe guard: abs(interest) > collateral. Excess dropped.
+            // Eg: Loan 100, interest -10 and collateral only a 1. SubLoanToken should be 0. 
+            if (subLoanAmount < 0) subLoanAmount = 0;
```
Jouzo, line 4537: This cannot happen. For subLoanAmount to be < 0, subInterest should be > collateralDUSD. Already handled in the if branch above.

prasannavl: Do you mean the `if (subInterest > collateralDUSD) {` branch? This case will not hit that branch, since `subInterest` will be negative.

Bushstar: SubInterest can negative which makes it less than collateralDUSD so the above guard fails.

Bushstar: I do not think this commit is correct though, we are resetting interest at the end, if it is negative we want to reduce the loan amount by it. Should be...

`if (subLoanAmount < 0) subLoanAmount = loanDUSD;`

Jouzo: Here we substract subInterest to collateralDUSD.
We already know that collateralDUSD > subInterest from the if branch above. (this handles positive interest)
If we deduct negative interest from collateralDUSD, it will be > to collateralDUSD which is known to be > 0. subLoanAmount can only be > 0

prasannavl: We can possibly leave the sanity checks on Sub* methods, but believe Jeremy is correct. Should be handled correctly.

PR #1467, opened 2022-09-13
ICX token type fix and enable on testnet

shohamc1:

/kind fix

fix maker bonus transfer token and limit ICX to testnet

Fixes #

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -3740,7 +3740,7 @@ bool IsDisabledTx(uint32_t height, CustomTxType type, const Consensus::Params& c
 
     // Leaving close orders, as withdrawal of existing should be ok?
     // disable ICX orders for all networks other than testnet
-    if (Params().NetworkIDString() != CBaseChainParams::TEST)
+    if (Params().NetworkIDString() != CBaseChainParams::TESTNET)
     {
     	switch (type) {
     	    case CustomTxType::ICXCreateOrder:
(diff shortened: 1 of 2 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -3739,15 +3739,19 @@ bool IsDisabledTx(uint32_t height, CustomTxType type, const Consensus::Params& c
...
+    	switch (type) {
+    	    case CustomTxType::ICXCreateOrder:
+    	    case CustomTxType::ICXMakeOffer:
+    	    case CustomTxType::ICXSubmitDFCHTLC:
+    	    case CustomTxType::ICXSubmitEXTHTLC:
+    	    case CustomTxType::ICXClaimDFCHTLC:
+    	        return true;
+    	    default:
+    	        return false;
+    	}
```
Jouzo, line 3754:
```suggestion
    // disable ICX orders for all networks other than testnet
    if (Params().NetworkIDString() == CBaseChainParams::TESTNET) return false;
    
    switch (type) {
        case CustomTxType::ICXCreateOrder:
        case CustomTxType::ICXMakeOffer:
        case CustomTxType::ICXSubmitDFCHTLC:
        case CustomTxType::ICXSubmitEXTHTLC:
        case CustomTxType::ICXClaimDFCHTLC:
            return true;
        default:
            return false;
```

Fixes `warning: non-void function does not return a value in all control paths`

PR #1473, opened 2022-09-15
Skip collateral factor check on JellyFish

Bushstar:


src/chainparams.cpp
```diff
@@ -760,7 +760,7 @@ class CRegTestParams : public CChainParams {
     explicit CRegTestParams() {
         strNetworkID = "regtest";
         bool isJellyfish = false;
-        isJellyfish = gArgs.GetBoolArg("-jellyfish_regtest", false);
+        isJellyfish = gArgs.GetBoolArg("-regtest-skip-loan-collateral-validation", false);
```
Jouzo, line 763: I'd rather we keep `-jellyfish_regtest` and add an additional flag for skipping `LoanCollateralFactor` validation.
It's confusing now seeing as `isJellyfish` has lots of side effect already. See https://github.com/DeFiCh/ain/blob/4b3c187f590738f44f5574901f1fec233a53c226/src/chainparams.cpp#L884

prasannavl: Agree with Jeremy. Do not touch `jellyfish_regtest`. I'd leave it as it is. The `regtest-skip-loan-collateral-validation` is an additional arg, that is added in addition, not intended to replace jellyfish_regtest.

src/masternodes/govvariables/attributes.cpp
```diff
@@ -1268,6 +1268,11 @@ Res ATTRIBUTES::Apply(CCustomCSView & mnview, const uint32_t height)
                 }
             } else if (attrV0->key == TokenKeys::LoanCollateralFactor) {
                 if (height >= static_cast<uint32_t>(Params().GetConsensus().FortCanningEpilogueHeight)) {
+                    // Skip on JellyFish
+                    if (gArgs.GetBoolArg("-regtest-skip-loan-collateral-validation", false)) {
```
prasannavl, line 1272: 1. Add a check to ensure it's first REGTEST network and chain it with the arg after. 
2. Remove the Skip on jellyfish comment.

prasannavl:
Let the PR be reused for now, but do not force push. Branch off and use new PRs.

PR #1494, opened 2022-09-27
Migrate tokens in futures live economy balances

Bushstar:


prasannavl:
Neat. Looks good. But let verify that we're using old accounting anywhere at the moment.

PR #1497, opened 2022-10-03
Replace ascendant/descendant values with uniform values

Bushstar:


src/masternodes/govvariables/attributes.cpp
```diff
@@ -205,8 +205,7 @@ const std::map<uint8_t, std::map<uint8_t, std::string>>& ATTRIBUTES::displayKeys
                 {TokenKeys::LoanMintingEnabled,        "loan_minting_enabled"},
                 {TokenKeys::LoanMintingInterest,       "loan_minting_interest"},
                 {TokenKeys::DFIP2203Enabled,           "dfip2203"},
-                {TokenKeys::Ascendant,                 "ascendant"},
-                {TokenKeys::Descendant,                "descendant"},
+                {TokenKeys::SplitValues,               "split_values"},
```
prasannavl, line 208: What do you think about calling it as `Migration`?

PR #1504, opened 2022-10-10
Consortium for minting DAT tokens

shohamc1:

/kind feature

#### Additional comments?

Supersedes #1364


src/masternodes/govvariables/attributes.cpp
```diff
@@ -362,6 +378,51 @@ static bool VerifyToken(const CCustomCSView& view, const uint32_t id) {
...
+    for (const auto &key : values.getKeys())
+    {
+        UniValue value(values[key].get_obj());
+        CConsortiumMember member;
+
+        member.status = 0;
+
+        member.name = trim_all_ws(value["name"].getValStr()).substr(0, CConsortiumMember::MAX_CONSORTIUM_MEMBERS_STRING_LENGTH);
+        if (!value["ownerAddress"].isNull())
+            member.ownerAddress = DecodeScript(value["ownerAddress"].getValStr());
```
Bushstar, line 397: `DecodeScript` cannot be called here, it will throw on an invalid address. `DecodeScript` is to be used in RPC calls only.

src/masternodes/govvariables/attributes.cpp
```diff
@@ -872,9 +956,11 @@ UniValue ATTRIBUTES::ExportFiltered(GovVarsFilter filter, const std::string &pre
             if (const auto bool_val = std::get_if<bool>(&attribute.second)) {
                 ret.pushKV(key, *bool_val ? "true" : "false");
             } else if (const auto amount = std::get_if<CAmount>(&attribute.second)) {
-                if (attrV0->type == AttributeTypes::Param &&
-                    (attrV0->typeId == ParamIDs::DFIP2203 || attrV0->typeId == ParamIDs::DFIP2206F) &&
-                    (attrV0->key == DFIPKeys::BlockPeriod || attrV0->key == DFIPKeys::StartBlock)) {
+                if ((attrV0->type == AttributeTypes::Param &&
+                        ((attrV0->typeId == DFIP2203 || attrV0->typeId == DFIP2206F) &&
+                            (attrV0->key == DFIPKeys::BlockPeriod || attrV0->key == DFIPKeys::StartBlock))) ||
+                    (attrV0->type == AttributeTypes::Consortium &&
```
Bushstar, line 962: `type` is checked against `Consortium` twice here, I think this should be...

`(attrV0->type == AttributeTypes::Consortium && attrV0->key == ConsortiumKeys::MintLimit)`

src/masternodes/govvariables/attributes.cpp
```diff
@@ -1049,6 +1174,29 @@ Res ATTRIBUTES::Validate(const CCustomCSView & view) const
                 }
             break;
 
+            case AttributeTypes::Consortium:
+                switch (attrV0->key) {
+                    case ConsortiumKeys::Members:
```
Bushstar, line 1179: ConsortiumKeys::Members and ConsortiumKeys::MintLimit have identical validation, can remove the code and the break under ConsortiumKeys::Members to apply the same code for both.

```
case ConsortiumKeys::Members:
case ConsortiumKeys::MintLimit:
```

src/masternodes/govvariables/attributes.h
```diff
@@ -199,12 +207,62 @@ enum FeeDirValues : uint8_t {
...
+using OracleSplits = std::map<uint32_t, int32_t>;
+using DescendantValue = std::pair<uint32_t, int32_t>;
+using AscendantValue = std::pair<uint32_t, std::string>;
+using CConsortiumMembers = std::map<std::string, CConsortiumMember>;
+using CConsortiumMembersMinted = std::map<DCT_ID, std::map<std::string, CConsortiumMinted>>;
+using CConsortiumGlobalMinted = std::map<DCT_ID, CConsortiumMinted>;
+using CAttributeType = std::variant<CDataStructureV0, CDataStructureV1>;
+using CAttributeValue = std::variant<bool, CAmount, CBalances, CTokenPayback, CTokenCurrencyPair, OracleSplits, DescendantValue, AscendantValue,
+                         CFeeDir, CDexBalances, CConsortiumMembers, CConsortiumMembersMinted, CConsortiumGlobalMinted>;
 using CDexBalances    = std::map<DCT_ID, CDexTokenInfo>;
```
Bushstar, line 261: These are all duplicates and can be removed.

src/masternodes/mn_checks.cpp
```diff
@@ -1227,36 +1236,164 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
 
     Res operator()(const CMintTokensMessage& obj) const {
         // check auth and increase balance of token's owner
-        for (const auto& kv : obj.balances) {
-            const DCT_ID& tokenId = kv.first;
-
-            if (Params().NetworkIDString() == CBaseChainParams::MAIN && height >= static_cast<uint32_t>(consensus.FortCanningCrunchHeight) &&
```
Bushstar, line 1233: The FortCanningCrunchHeight consensus code to prevent loan tokens from being minted is missing and needs to be restored.

src/masternodes/mn_checks.h
```diff
@@ -316,6 +318,28 @@ struct CMintTokensMessage : public CBalances {
...
 
+struct CBurnTokensMessage {
+    enum BurnType : uint8_t
+    {
+        TokenBurn = 0,
+    };
+
+    CBalances amounts;
+    CScript from;
+    uint8_t burnType;
```
Bushstar, line 329: Type should be BurnType for clarity.

src/masternodes/rpc_customtx.cpp
```diff
@@ -105,14 +113,25 @@ class CCustomTxRpcVisitor
...
+        rpcInfo.pushKVs(tokenBalances(obj.amounts));
+        rpcInfo.pushKV("from", ScriptToString(obj.from));
+        std::string type;
+        switch (obj.burnType)
+        {
+            case CBurnTokensMessage::BurnType::TokenBurn:
+                type = "TokenBurn";
+                break;
+            default:
+                type = "TokenBurn";
```
Bushstar, line 129: Maybe this should default to "Unexpected type" as the TokenBurn type should have been picked up from above :thinking:

src/masternodes/rpc_tokens.cpp
```diff
@@ -725,6 +749,113 @@ UniValue minttokens(const JSONRPCRequest& request) {
...
+
+    burnedTokens.burnType = CBurnTokensMessage::BurnType::TokenBurn;
+    if (!metaObj["context"].isNull())
+        burnedTokens.context = DecodeScript(metaObj["context"].getValStr());
+
+    UniValue const & txInputs = request.params[2];
+
+    int targetHeight = chainHeight(*pwallet->chain().lock()) + 1;
+
+    for (auto const & kv : burnedTokens.amounts.balances) {
```
Bushstar, line 816: This check is applied on consensus side via `SubMintedTokens` so this can be removed. When the TX is checked via `execTestTx` at the end of the RPC call it will be rejected if the token is not found.

Bushstar (approved):

src/masternodes/govvariables/attributes.cpp
```diff
@@ -948,6 +1073,29 @@ Res ATTRIBUTES::Validate(const CCustomCSView & view) const
...
+                        if (view.GetLastHeight() < Params().GetConsensus().GreatWorldHeight) {
```
hidiego: `if (view.GetLastHeight() < Params().GetConsensus().GrandCentralHeight) {`

hidiego: GreatWorldHeight -> GrandCentralHeight ?

shohamc1: That's weird, it already is GrandCentral.

https://github.com/DeFiCh/ain/blob/featuer/consortium/src/masternodes/govvariables/attributes.cpp#L1194

hidiego: I was in a different commit when I did this review. Resolving conversation.

Mixa84:
Consortium will be merged directly in master.

PR #1513, opened 2022-10-17
Change block time and difficulty adjust period

Bushstar:


uzyn:
This PR is just a study and is not a confirmed change. 
An actual change would require a DFIP.

Key reasoning for this study:
- 30sec average block time (sometimes as long as 5 minutes) would be bad when we have EVM.
- EVM users are used to faster blocks. Ethereum about 12 secs, consistently.
- This is an exploration for going for lower blocktime for EVM use-case.

PR #1514, opened 2022-10-17
Adds argument `-leveldbchecksum`

hidiego:

/kind feature

Adds new arg `-leveldbchecksum`.
If set, leveldb will stop doing aggressive checking of the data it is processing.

### DIFF
--- a/src/dbwrapper.cpp
+++ b/src/dbwrapper.cpp
@@ -105,7 +105,7 @@ static leveldb::Options GetOptions(size_t nCacheSize)
         v++;
         return v;
     };
-    
+
     leveldb::Options options;
     options.block_cache = leveldb::NewLRUCache(nCacheSize / 2);
     options.write_buffer_size = ceil_power_of_two(std::min(static_cast<size_t>(64)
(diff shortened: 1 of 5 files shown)


src/dbwrapper.cpp
```diff
@@ -126,12 +126,26 @@ CDBWrapper::CDBWrapper(const fs::path& path, size_t nCacheSize, bool fMemory, bo
...
+    auto leveldbchecksum = gArgs.GetArg("-leveldbchecksum", "auto");
+    if(leveldbchecksum == "false"){
+        // set to true by default for all versions above 1.16 of LevelDB in GetOptions()
+        options.paranoid_checks = false;
+    }
+    else if (leveldbchecksum == "true") {
+        readoptions.verify_checksums = true;
+        iteroptions.verify_checksums = true;
+    }
+    else if (leveldbchecksum == "auto"){
```
Bushstar, line 143: Should we remove the if statement here to make this the default else in case someone made a typo when setting leveldbchecksum?

src/init.cpp
```diff
@@ -391,12 +391,22 @@ void SetupServerArgs()
...
+        "-dbcrashratio",
+        "-forcecompactdb",
+        "-leveldbchecksum=<auto|true|false>",
+        "-interrupt-block=<hash|height>",
+        "-stop-block=<hash|height>",
+        "-mocknet",
+        "-mocknet-blocktime=<secs>",
+        "-mocknet-key=<pubkey>",
         "-checkpoints-file",
         // GUI args. These will be overwritten by SetupUIArgs for the GUI
```
Bushstar, line 403: All the options below this line can be removed as they applied to the removed Qt GUI.

src/dbwrapper.cpp
hidiego: you are right we need to handle wrong input, what about a final else throwing an error or at least showing/logging a message and then set to default behaviour?

Bushstar (approved):

Jouzo (approved):

prasannavl:
Rework needed before merging again: https://github.com/DeFiCh/ain/pull/1521#issue-1417396529

PR #1534, opened 2022-10-31
Poolswap to empty to address swaps to from address making to optional

Bushstar:
With a sync from genesis this PR resolves negative interest being taken from the user's balance on payback and when failing to set an address on a pool swap the swapped amount is sent to the from address, not an empty inaccessible address.


src/masternodes/mn_checks.cpp
```diff
@@ -3541,7 +3541,7 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
                 if (paybackTokenId == loanTokenId)
                 {
                     // If interest was negative remove it from sub amount
-                    if (height >= static_cast<uint32_t>(consensus.FortCanningEpilogueHeight) && subInterest < 0)
+                    if (subInterest < 0)
```
kuegi, line 3544: this will risk chain split while everyone updates, right? cause when someone uses the higher DUSD amount in the wallet  after the update, the not-yet-updated nodes will reject that block.
Any plans how to best coordinate that with everyone?

prasannavl: Yes. Indeed. The majority network needs to be on v2.x until the fork block. And then switch over. We're exploring the deployment strategies before getting these merged.

prasannavl: (In addition, we're also exploring fixing https://github.com/DeFiCh/ain/pull/1533 together, use this as a one-shot chance to benefit users and fix past bugs)

kuegi: really love it when this gets in. Major benefit for the community! I believe if cake & mydefichain (main ocean providers IMHO) switch coordinated, LW is fine. Probably need to get the word out to everyone with own node to have a suggested date for the update. 
Looking forward to your plan and how I can help.

src/masternodes/mn_checks.cpp
```diff
@@ -4355,7 +4355,7 @@ Res CPoolSwap::ExecuteSwap(CCustomCSView& view, std::vector<DCT_ID> poolIDs, boo
             intermediateView.Flush();
 
             auto& addView = lastSwap ? view : intermediateView;
-            res = addView.AddBalance(lastSwap ? obj.to : obj.from, swapAmountResult);
+            res = addView.AddBalance(lastSwap ? (obj.to.empty() && height >= static_cast<uint32_t>(Params().GetConsensus().GrandCentralHeight) ? obj.from : obj.to) : obj.from, swapAmountResult);
```
prasannavl, line 4358:
```suggestion
            if (height >= static_cast<uint32_t>(Params().GetConsensus().GrandCentralHeight)) {
                res = addView.AddBalance(lastSwap ? (obj.to.empty() ? obj.from : obj.to) : obj.from, swapAmountResult);
            } else {
                res = addView.AddBalance(lastSwap ? obj.to : obj.from, swapAmountResult);
            }
```

prasannavl (approved):
Just some boring changes to prefer readability.

PR #1539, opened 2022-11-01
Replace ascendant/descendant values with uniform values

Bushstar:
On token split we add a record of that split to Governance variables, the previous record was missing the multiplier amount which is not available elsewhere after the split has completed. This updates the record to include complete information and have the same record for the old and new token.

## New
* `listgovs` / `getgov` entry for token split as follows.
  * key - v0/token/ID/migration
  * value - OLD_ID/MULTIPLIER/NEW_ID/HEIGHT

## Breaking
* Requires reindex.
* Removes `descendant` / `ascendant` keys from `listgovs` / `getgov`


test/functional/feature_token_split.py
```diff
@@ -361,8 +361,8 @@ def check_token_split(self, token_id, token_symbol, token_suffix, multiplier, mi
             assert_equal(result[f'v0/token/{token_id}/loan_minting_enabled'], 'true')
             assert_equal(result[f'v0/token/{token_id}/loan_minting_interest'], '0')
         assert(f'v0/oracles/splits/{self.nodes[0].getblockcount()}' not in result)
-        assert_equal(result[f'v0/token/{token_idv1}/descendant'], f'{token_id}/{self.nodes[0].getblockcount()}')
-        assert_equal(result[f'v0/token/{token_id}/ascendant'], f'{token_idv1}/split')
+        assert_equal(result[f'v0/token/{token_idv1}/migration'], f'{token_idv1}/{multiplier}/{token_id}/{self.nodes[0].getblockcount()}')
+        assert_equal(result[f'v0/token/{token_id}/migration'], f'{token_idv1}/{multiplier}/{token_id}/{self.nodes[0].getblockcount()}')
```
prasannavl, line 365: We might not want to have the same migration attr on both tokens though.. or we could continue to ascendant / descendant, but with this format. 

Token v0: descendant = v0/../v1/.. (linking to next token)
Token v1: ascendant =  v0/../v1/.. (linking to prev token)
Token v1: descendant = v1/..v2/..
Token v2: ascendant = v1/..v2/..
...

prasannavl:
Outdated. Closing.

PR #1541, opened 2022-11-01
Consortium implementation

Mixa84:

/kind feature

## New:

`burntoken` RPC: 

- `values` metadata:
  - `amount`
  - `from`
  - `context`

#### Additional comments?

Supersedes #1364 and #1504


Jouzo (dismissed):

src/masternodes/govvariables/attributes.cpp
```diff
@@ -344,6 +364,60 @@ static bool VerifyToken(const CCustomCSView& view, const uint32_t id) {
...
+        return Res::Err("Not a valid consortium member object!");
+
+    for (const auto &key : values.getKeys())
+    {
+        UniValue value(values[key].get_obj());
+        CConsortiumMember member;
+
+        member.status = 0;
+
+        member.name = trim_all_ws(value["name"].getValStr()).substr(0, CConsortiumMember::MAX_CONSORTIUM_MEMBERS_STRING_LENGTH);
```
Bushstar, line 381: Maybe enforce a minimum length here as well. "name" could either not be set or set blank.

Jouzo (dismissed):

src/masternodes/govvariables/attributes.cpp
Mixa84: I left it optional as maybe there could be a member set without a name. Id is necessary and will be used as unique. This will all be set by foundation so didn't do fool proof validation.

Mixa84:
:+1:

src/masternodes/govvariables/attributes.cpp
```diff
@@ -464,6 +484,64 @@ static bool VerifyToken(const CCustomCSView& view, const uint32_t id) {
...
+        if (member.name.size() < CConsortiumMember::MIN_CONSORTIUM_MEMBERS_STRING_LENGTH) {
+            return Res::Err("Member name too short, must be more than %d chars long", int(CConsortiumMember::MIN_CONSORTIUM_MEMBERS_STRING_LENGTH));
```
helloscoopa, line 503: As not checking for `<=`

```suggestion
            return Res::Err("Member name too short, must be at least %d chars long", int(CConsortiumMember::MIN_CONSORTIUM_MEMBERS_STRING_LENGTH));
```

PR #1552, opened 2022-11-10
Skip token locked check

hidiego:

/kind fix

Adds new parameter skipLockedCheck to `listvaults`. 

If set to "true" information relative to the tokens of the vault is shown even during the split period including; collateral value, loan value, interest value, informative ratio and collateral ratio. 
`skipLockedCheck` is optional and defaults to false. Previous behaviour is maintained showing "-1" on all the above fields as the vault is in frozen state.

#### Additional comments?:

This PR is a contains previous work done in #1326 nwith a clean history, easier to follow changes and merge into master.


shohamc1 (approved):

src/masternodes/rpc_vault.cpp
```diff
@@ -249,9 +247,13 @@ namespace {
             } else {
                 interestsPerBlockBalances = AmountsToJSON(interestsPerBlock);
                 totalInterestsPerBlockValue = ValueFromAmount(totalInterestsPerBlock);
+            }
+            if (!isVaultTokenLocked){
+                // IPB can only be calculated when tokens are not locked.
+                // Value during split period is hard to predict with the current|next price
+                result.pushKV("interestsPerBlock", interestsPerBlockBalances);
```
prasannavl, line 254: Looks good, but still keep the json keys and set it to -1. Don't skip the values.

prasannavl (changes requested):
Looks good beyond some minor changes

PR #1562, opened 2022-11-13
Fix CI full sync test

Bushstar:
Full sync tests fail due to subtract being called on zero value generating an error. Fix is to skip updating Governance variables when setting Add or Sub is in error.


prasannavl:
All the full syncs are successful now. The tests outcome still fails due to the introduction of the new `consortiumtokens` field that's not a part of the diffs yet, but everything else matches up.

PR #1589, opened 2022-11-21
DUSDLock Smartcontract

kuegi:

/kind feature

Implementation for DFIP 2211-D: https://github.com/DeFiCh/dfips/issues/236


src/masternodes/mn_checks.cpp
```diff
@@ -2037,6 +2044,101 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        auto pair = mnview.GetPoolPair(lockToken->first,dusdToken->first);
+        if (!pair) {
+            return Res::Err("lock pool not found");
+        }
+        const auto& lpTokenID = pair->first;
+        auto& pool = pair->second;
+        auto res = pool.AddLiquidity(half, half, [&] /*onMint*/(CAmount liqAmount) {
+
+            CBalances balance{TAmounts{{lpTokenID, liqAmount}}};
+            return AddBalanceSetShares(obj.shareAddress, balance);
```
Krysh90, line 2136: should be `obj.owner`, right?

src/masternodes/rpc_accounts.cpp
```diff
@@ -2085,6 +2085,87 @@ UniValue executesmartcontract(const JSONRPCRequest& request) {
     return NullUniValue;
 }
 
+//TODO: @coredevs: added it here since futureswap is here too, but not sure if this is the right place
```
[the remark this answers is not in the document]
kuegi: will move it. you have an updated PR with tests in 5 hours 👍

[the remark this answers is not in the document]
prasannavl: Please take your time on it. I really wouldn't rush this into this release.

prasannavl:
This is really great work and super appreciate the effort to get this in. But really wouldn't rush this -- would be better to have get some mind space to think about how to actually implement this cleanly. 

Today it's lock 1 and lock 2, what if we need a different variant. It would be better to think through this some more before setting this in stone as a part of an immutable blockchain.

src/masternodes/rpc_accounts.cpp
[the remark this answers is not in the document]
kuegi: just checked, DFIP2201 is also handled in rpc_accounts. actually right above this function. so I leave it here for now.

Mixa84:
@kuegi Nice work. I'll write comments on the questions you asked in the code comments and necessary modifications.

kuegi:
> @kuegi Nice work. I'll write comments on the questions you asked in the code comments and necessary modifications.
@Mixa84 
thx. but its still WIP. since I didn't have a env setup its lots of small things not working out yet. Today I finally set it up and will refactor a bit (since we decided to not squeeze it into v3).
So maybe wait till tomorrow, then the state of the branch will be much better (with tests and all)

src/masternodes/mn_checks.cpp
```diff
@@ -2037,6 +2044,101 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        if(*resultedMint.val > limit) {
+            return Res::Err("Limit reached for this lock token");
+        }
+
+        //do we need CalculateOwnerRewards here?
+        res = mnview.SubBalance(obj.owner, CTokenAmount{dusdToken->first, obj.dusdIn});
+        if (!res)
+            return res;
+        
+        auto half= DivideAmounts(obj.dusdIn,CAmount{2});
```
Mixa84, line 2115: When using division, it will happen that it is not dividable by 2 with exact amount and there will be remainder which would be lost. So you should use `half` and `obj.dUsdIn - half` for the other half.

src/masternodes/mn_checks.cpp
```diff
@@ -2037,6 +2044,101 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        //half DUSD are burned, that amount DUSDLOCK is minted
+        //remaining half + DUSDLOCK added to pool
+        //sub minted or burn?
+        res = mnview.SubMintedTokens(dusdToken->first, half);
```
Mixa84, line 2119: SubMintedToknes are used to maintain the evidence of minted amount, but it doesn't burn anything. So you must also send those DUSD to burn address.

src/masternodes/mn_checks.cpp
```diff
@@ -2037,6 +2044,101 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        if(!lockToken) {
+            return Res::Err("Could not find lock token");
+        }
+        
+        auto dusdToken = mnview.GetToken("DUSD");
+        if(!dusdToken) {
+            return Res::Err("Could not find dusd token");
+        }
+
+        const auto contractAddressValue = GetFutureSwapContractAddress(SMART_CONTRACT_DFIP2211D);
```
Mixa84, line 2096: I don't see this used anywhere?

src/masternodes/rpc_customtx.cpp
```diff
@@ -195,6 +195,19 @@ class CCustomTxRpcVisitor
...
+
+    void operator()(const CLockDUSDMessage& obj) const {
+        CTxDestination dest;
+        if (ExtractDestination(obj.owner, dest)) {
+            rpcInfo.pushKV("owner", EncodeDestination(dest));
+        } else {
+            rpcInfo.pushKV("owner", "Invalid destination");
+        }
+
+        rpcInfo.pushKV("dusdIn", GetDecimaleString(obj.dusdIn));
```
Mixa84, line 208: For amounts we use `ValueFromAmount()`

src/masternodes/rpc_customtx.cpp
```diff
@@ -195,6 +195,19 @@ class CCustomTxRpcVisitor
...
+        rpcInfo.pushKV("lockTime", std::to_string(obj.lockTime));
```
Mixa84, line 209: Don't convert it to string, just cast it to plain `int` as UniValue doesn't support `(u)intX_t` types

src/masternodes/balances.h
```diff
@@ -241,6 +241,22 @@ struct CFutureSwapMessage {
     }
 };
 
+
+struct CLockDUSDMessage {
+    CScript owner;
+    CAmount dusdIn{};
```
Mixa84, line 247: I think `{}` is not necessary.

src/masternodes/mn_checks.cpp
```diff
@@ -2037,6 +2044,101 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor
...
+        if (!contractAddressValue) {
+            return contractAddressValue;
+        }
+
+        auto limit = attributes->GetValue(obj.lockTime == 12 ? limit12Key : limit24Key, CAmount{0});
+        auto resultedMint= SafeAdd(lockToken->second->minted, obj.dusdIn);
```
Mixa84, line 2102: I think only half should be added to minted DUSDLOCK token?

Mixa84:
> > I'll write comments on the questions you asked in the code comments and necessary modifications.
> 
> @Mixa84 thx. but its still WIP. since I didn't have a env setup its lots of small things not working out yet. Today I finally set it up and will refactor a bit (since we decided to not squeeze it into v3). So maybe wait till tomorrow, then the state of the branch will be much better (with tests and all)

I've added some stuff I saw currently and it will maybe help you in the refactor. If you already addressed those just resolve it ;)

src/masternodes/mn_checks.cpp
kuegi: do I? I substract it from the balance and reduce the minted. Thats basically the same thing that happens when they would be used to payback a loan. (its as if they have never been there at all)
Since they are replaced by locked_dusd I thought its better two "swap" minted for minted: reduce minted DUSD and increase minted lock.
if we send them to burn, we increase total amount (not net supply but total minted) in the system which feels "wrong" here.

kuegi:
> > > I'll write comments on the questions you asked in the code comments and necessary modifications.
> > 
> > 
> > @Mixa84 thx. but its still WIP. since I didn't have a env setup its lots of small things not working out yet. Today I finally set it up and will refactor a bit (since we decided to not squeeze it into v3). So maybe wait till tomorrow, then the state of the branch will be much better (with tests and all)
> 
> I've added some stuff I saw currently and it will maybe help you in the refactor. If you already addressed those just resolve it ;)

thx a lot. adapted the code and logic. Now using a more generic approach so the attributes can define what lockperiods are available and which token is used (so no need to hardcode the token key now).

Not sure if the way I did the attribute keys is good or complete crap. 

More tests are likely better, also to split the test function into more smaller tests. But at least they are green now ;)

kuegi:
I am thinking about switching the logic from "lockTime" to "batchId". cause in the end, that number is random.
So we could build it with logic that there are batches in this smartcontract. Each batch would have
* associated lockToken
* limit how much DUSD can be swapped
* blockheight when withdrawal/swap back is possible

Any thoughts on that? @Mixa84 @prasannavl

Krysh90 (dismissed):

src/masternodes/govvariables/attributes.cpp
```diff
@@ -585,6 +593,9 @@ const std::map<uint8_t, std::map<uint8_t, std::function<ResVal<CAttributeValue>(
                  {DFIPKeys::ConsortiumEnabled, VerifyBool},
                  {DFIPKeys::CFPPayout, VerifyBool},
                  {DFIPKeys::EmissionUnusedFund, VerifyBool},
+                 {DFIPKeys::Limit,                   VerifyUInt32},
```
Mixa84, line 596: Make limit as `VerifyPositiveFloat` which will keep the amount in satoshis.

src/masternodes/mn_checks.cpp
```diff
@@ -2171,6 +2180,164 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+        auto res = CheckCustomTx();
+        if (!res)
+            return res;
+
+        if (!HasAuth(obj.owner)) {
+            return Res::Err("Transaction must have at least one input from owner");
+        }
+
+        const auto attributes = mnview.GetAttributes();
+        if (!attributes) {
```
Mixa84, line 2194: this can be only `assert(attributes)` as attributes exist already and if not found then it is a critical error.

src/masternodes/mn_checks.cpp
```diff
@@ -2171,6 +2180,164 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+        }
+
+        auto dusdToken = mnview.GetToken("DUSD");
+        if(!dusdToken) {
+            return Res::Err("Could not find dusd token");
+        }
+
+
+        auto tokenId= attributes->GetValue(tokenKey, (uint32_t)0);
+        auto limit = attributes->GetValue(limitKey,(uint32_t)0)*COIN;
```
Mixa84, line 2219: When you change limit key to `VerifyPositiveFloat` then remove `* COIN`

src/masternodes/mn_checks.cpp
```diff
@@ -2171,6 +2180,164 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+            if (!res)
+                return res;
+
+            auto dusdhalf = obj.source.nValue / 2;
+            auto lockedhalf = obj.source.nValue - dusdhalf;
+
+            if (dusdhalf == 0 || lockedhalf == 0) {
+                return Res::Err("amount too small.");
+            }
+            auto resultedMint = SafeAdd(lockToken->minted, lockedhalf);
```
Mixa84, line 2258: This is redundant, as it is done in `AddMintedTokens` so just remove it.

src/masternodes/mn_checks.cpp
```diff
@@ -2171,6 +2180,164 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+            auto withdrawHeight = attributes->GetValue(withdrawKey, (uint32_t)0);
+            if (withdrawHeight == 0 || withdrawHeight > height) {
+                return Res::Err("This batch is not open for withdrawal yet.");
+            }
+
+            CAmount lockTokenToSwap = obj.source.nValue;
+            if (obj.source.nTokenId == lpTokenID) {
+                // remove LP token first
+                //  subtract liq.balance BEFORE RemoveLiquidity call to check balance correctness
+                {
```
Mixa84, line 2301: You can remove block `{`,`}`

src/masternodes/mn_checks.cpp
kuegi: I need resultedMint to check against the limit beforehand not only to check if the add is fine.

src/masternodes/mn_checks.cpp
kuegi: tried that, doesn't work. question is how we want the limit to be put in. as fi or as whole DUSD. so should "limit = 10k DUSD" lead to a gov-value of 10k or 10k*1e8?
imho it should be "10000".
Then I need the *COIN here to compare limit with resultedMint (which is in fi) later on.

Mixa84: For all amounts that we stored in attributes we used `VerifyPositiveFloat` which will convert 1.23456789 into int64 123456789 - so into COIN precision. You can use `VerifyPositiveFloatOrMinusOne` which was used on Consortium for mintLimit and mintLimitDaily which is probably what we need here as stating -1 is unlimited.

For consistency I would use this format, even if the limit will be huge here ant satoshis won't matter but we should keep it constant.

kuegi:
@Mixa84 Thank you very much for the comments. I tried to implement all of them.

kuegi:
@Mixa84 @prasannavl anything left for me to do here?

Mixa84 (changes requested):

src/masternodes/govvariables/attributes.h
```diff
@@ -85,6 +86,9 @@ enum DFIPKeys : uint8_t {
     GovernanceEnabled    = 'q',
     CFPPayout            = 'r',
     EmissionUnusedFund   = 's',
+    Limit                = 't',
```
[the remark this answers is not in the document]
kuegi: @Mixa84  Is it mandatory that one key is only used for one DFIP? I kept it "generic" so this Limit could be reused if anyother DFIP also needs a limit, same for Token and WithdrawHeight. Verifications are done in a way that it can be used for other DFIPs too. and within the params its anyway in the path of the DFIP.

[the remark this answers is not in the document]
Mixa84: I thought about that and I would like to reuse them for other DFIPs, but I think it is not clear from the text what it does.
`v0/params/dfip2211d/limit/1`, `v0/params/dfip2211d/token/1`
@prasannavl what do you say on this?

prasannavl:
Closing stale PRs. No longer relevant.

PR #1613, opened 2022-11-28
Change ordering of TXs by time

Bushstar:
When looping over TXs to add to the block loop over them by time ordering.


src/init.cpp
```diff
@@ -510,7 +510,7 @@ void SetupServerArgs()
     gArgs.AddArg("-regtest-minttoken-simulate-mainnet", "Simulate mainnet for minttokens on regtest -  default behavior on regtest is to allow anyone to mint mintable tokens for ease of testing", ArgsManager::ALLOW_ANY | ArgsManager::DEBUG_ONLY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-simulatemainnet", "Configure the regtest network to mainnet target timespan and spacing ", ArgsManager::ALLOW_ANY | ArgsManager::DEBUG_ONLY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-dexstats", strprintf("Enable storing live dex data in DB (default: %u)", DEFAULT_DEXSTATS), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
-    gArgs.AddArg("-blockfeeordering", strprintf("Whether to order transactions by fee, otherwise ordered by time (default: %u)", DEFAULT_FEE_ORDERING), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
+    gArgs.AddArg("-blocktimeordering", strprintf("Whether to order transactions by time, otherwise ordered by fee (default: %u)", DEFAULT_FEE_ORDERING), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
```
kuegi, line 513: Why use fee as default? I thought the default should be time to prevent frontrunning. @prasannavl

prasannavl: Sorry, I missed this earlier. It's because fee is quite an intricate part of quite a few things, including fee estimation and many things that rely on this behaviour. 

In order to avoid surprises, and also since this is a new behavior, would be best to let it be used optionally for a while, before considering it as default.

PR #1629, opened 2022-12-05
chore(libmc.mk): get architecture from script

### LINKED ISSUE #1616: find a way to support m1 & intel macs as build target
#### Why is this needed:
We do not have a way to identify which architecture to use when we get a match for `x86_64-apple-darwin18` since both the intel & m1 macs are a match. We need to find a way to identify m1 macs so we can set the target to `aarch64-apple-darwin`.

DieHard073055:

/kind chore

Figure out if the host architecture is apple m1 mac or intel.

Fixes #1616


depends/libmc_target.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/bash
```
wafflespeanut, line 1: If this works as expected when building for multiple targets, then we could rename it to `rust_targets.sh` and use it for `libain.mk` as well.

DieHard073055: updated filename.

wafflespeanut: And using it in `libain.mk`?

DieHard073055: and `libain.mk`

PR #1627, opened 2022-12-01
Add cycle filtering and pagination to listgovproposals

dcorral:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:

- Adds cycle filtering in `listgovproposals` RPC

- Adds pagination to `listgovproposals` RPC :

Pagination arguments can be provided via the pagination CLI argument.
```
defi-cli listgovproposals 'all' 'rejected' 1 '{"limit": 2}'
defi-cli listgovproposals '' ''' 0 '{"limit": 2}'
```

If you do not want to set the RPC's optional arguments, users can pass arguments via an object instead.

```
defi-cli listgovproposals '{"type": "all", "pagination": {"limit": 2}}'
```

PR #1635, opened 2022-12-07
Add pagination support for `listgovproposalvotes`

shohamc1:

/kind feature

Provides pagination support for the `listgovproposalvotes` RPC.

Pagination arguments can be provided via the `pagination` CLI argument.

```bash
$ defi-cli listgovproposalvotes 'propId' 'all' -1 '{"limit": 2}'
``` 
If you do not want to set the RPC's optional arguments, users can pass arguments via an object instead.

```bash
$ defi-cli listgovproposalvotes '{"proposalId": "propId", "pagination": {"limit": 2}}'
``` 

There are three pagination arguments available:
- `start`: Index of the first entry to return. Usually `start + limit` of the previous call. 
- `including_start`: To include first entry of the results. This is set to `false` if `start` has been set.
- `limit`: Maximum number of entries to return in one RPC call.

Successive pagination calls would look like:
```bash
$ defi-cli listgovproposalvotes 'propId' 'all' -1 '{"start": 0, "including_start": true, "limit": 2}'
$ defi-cli listgovproposalvotes 'propId' 'all' -1 '{"start": 2, "including_start": true, "limit": 2}'
$ defi-cli listgovproposalvotes 'propId' 'all' -1 '{"start": 4, "including_start": true, "limit": 2}'
```


dcorral (commented):
This will filter current cycle of the proposal which is internal to the proposal. We will need some mechanism to filter by global cycle meaning "get all proposals in the Nth voting round". Have a look at #1627 for a WIP implementation of this filter. 

A suggestion was made to call this two filters differently:
- global cycle filter -> batch
- internal cycle -> cycle

UPDATE: This will apply for `listgovproposals` filtering, current implementation for `listgovproposalvotes` is actually valid.

Mixa84 (approved):

Mixa84 (approved):

Jouzo (approved):

PR #1638, opened 2022-12-07
Allow MineBlocksOnDemand on mocknet

Jouzo:

/kind feature

- Allows MineBlocksOnDemand


Jouzo:
> Good thought. Would it be better to set `fNoRetargeting` only on mocktime being set?
> 
> Alternatively, we can remove the branch that currently overrides for mocknet, and reuse mocktime to use a default proof of work limit from `params.pos.diffLimit`.

Agreed. The idea behind this PR is to be able to use `setmocktime` in order to `generatetoaddress` faster, as being done in test framework. 
I changed a bit the logic, reverted setting `fNoRetargeting` on mocknet and updated `MineBlocksOnDemand` directly to restrict side effects.

PR #1650, opened 2022-12-13
Minting tokens to any address

shohamc1:

/kind feature

Allow minting tokens to any address.
- Adds new `to` argument to `minttokens` RPC
- Modify`CMintTokensMessage` to add new fields

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -1310,7 +1310,7 @@ public:
         const auto grandCentralHeight       = static_cast<uint32_t>(consensus.GrandCentralHeight);
 
         // check auth and increase balance of token's owner
-        for (const auto &[tokenId, amount] : obj.balances) {
+        for (const auto &[tokenId, amount] : obj.mint.balances) {
             if (Params().NetworkIDString() == CBaseChainParams::MAIN && height >= fortCanningCrunchHeight &&
                 mnview.GetLoanTokenByID(tokenId)) {
                 return Res::Err("Loan tokens cannot be minted");
@@ -1329,7 +1329,14 @@ public:
                     return minted;
 
                 CalculateOwnerRewards(*mintable.val);
-                auto res = mnview.AddBalance(*mintable.val, CTokenAmount{tokenId, amount});
+                CScript mintTo{*mintable.val};
+
+                if (!obj.to.empty()) {
+                    mintTo = obj.to;
+                    CalculateOwnerRewards(mintTo);
+                }
+
+                auto res = mnview.AddBalance(mintTo, CTokenAmount{tokenId, amount});
                 if (!res)
                     return res;
 
--- a/src/masternodes/mn_checks.h
+++ b/src/masternodes/mn_checks.h
@@ -311,13 +311,15 @@ struct CUpdateTokenMessage {
     }
 };
 
-struct CMintTokensMessage : public CBalances {
-    using CBalances::CBalances;
+struct CMintTokensMessage {
+    CBalances mint;
+    CScript to;
 
     ADD_SERIALIZE_METHODS;
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream &s, Operation ser_action) {
-        READWRITEAS(CBalances, *this);
+        READWRITE(mint);
+        READWRITE(to);
     }
 };
 
--- a/src/masternodes/rpc_tokens.cpp
+++ b/src/masternodes/rpc_tokens.cpp
@@ -639,36 +639,43 @@ UniValue getcustomtx(const JSONRPCRequest& request)
 UniValue minttokens(const JSONRPCRequest& request) {
     auto pwallet = GetWallet(request);
 
-    RPCHelpMan{"minttokens",
-               "\nCreates (and submits to local node and network) a transaction minting your token (for accounts and/or UTXOs). \n"
-               "The second optional argument (may be empty array) is an array of specific UTXOs to spend. One of UTXO's must belong to the token's owner (collateral) address" +
-               HelpRequiringPassphrase(pwallet) + "\n",
-               {
-                    {"amounts", RPCArg::Type::STR, RPCArg::Optional::NO,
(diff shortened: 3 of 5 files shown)


src/masternodes/rpc_tokens.cpp
```diff
@@ -639,45 +639,77 @@ UniValue getcustomtx(const JSONRPCRequest& request)
...
+            txInputs = optionsObj["inputs"].get_array();
+
+        if (optionsObj.exists("to"))
+            to = DecodeScript(optionsObj["to"].get_str());
+    }
+    else {
+        minted   = DecodeAmounts(pwallet->chain(), request.params[0], "");
+        txInputs = request.params[1];
+
+        if (request.params.size() > 2)
```
Mixa84, line 705: I wouldn't put `to` as last param as it means you need to supply `inputs` before `to`. I would avoid adding `to` if it is not a JSON object.

Also, we need to prevent having to as empty address supplied so that it doesn't accidentally go to empty address!

Mixa84 (changes requested):

src/masternodes/rpc_tokens.cpp
Bushstar: I think to needs to be after inputs to not break anything, interface needs to remain the same for services using this. Empty inputs can be provided.

src/masternodes/mn_checks.h
```diff
@@ -313,11 +313,13 @@ struct CUpdateTokenMessage {
 
 struct CMintTokensMessage : public CBalances {
     using CBalances::CBalances;
+    CScript to;
 
     ADD_SERIALIZE_METHODS;
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream &s, Operation ser_action) {
         READWRITEAS(CBalances, *this);
+        READWRITE(to);
```
Bushstar, line 322: Cannot just change serialisation, this will break all previous mint token messages as to will always be expected. See this link below for optional serialisation.

https://github.com/DeFiCh/ain/blob/221a8489a632098f1dbb1cc1af84de98aba369c6/src/masternodes/mn_checks.h#L251-L254

src/masternodes/mn_checks.cpp
```diff
@@ -1329,7 +1329,14 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
                     return minted;
 
                 CalculateOwnerRewards(*mintable.val);
-                auto res = mnview.AddBalance(*mintable.val, CTokenAmount{tokenId, amount});
+                CScript mintTo{*mintable.val};
+
+                if (!obj.to.empty()) {
```
Bushstar, line 1334: Usage of to needs to be behind a fork, otherwise we will have a chain split between upgraded and non-upgraded nodes when it is used. Below fork we need to reject TXs that set the to field.

src/masternodes/rpc_tokens.cpp
```diff
@@ -639,45 +639,77 @@ UniValue getcustomtx(const JSONRPCRequest& request)
...
-                        "A json array of json objects. Provide it if you want to spent specific UTXOs",
+    RPCHelpMan{
+        "minttokens",
+        "\nCreates (and submits to local node and network) a transaction minting your token (for accounts and/or "
+        "UTXOs). \n"
+        "The second optional argument (may be empty array) is an array of specific UTXOs to spend. One of UTXO's must "
+        "belong to the token's owner (collateral) address" +
+            HelpRequiringPassphrase(pwallet) + "\n",
+        {
+                          {"amounts",
```
Bushstar, line 650: The formatting here does not look correct to me.

shohamc1: Fixed: [55850b2](https://github.com/DeFiCh/ain/pull/1650/commits/55850b2a298c4d928c2940e752814ed2fa972a9d)

src/masternodes/mn_checks.h
shohamc1: Fixed: [174f993](https://github.com/DeFiCh/ain/pull/1650/commits/174f9932dc7a291626cc132673d64e11700936f6)

src/masternodes/rpc_tokens.cpp
Mixa84: Ahhh, yes. I forgot for the moment ...

src/masternodes/mn_checks.cpp
shohamc1: Fixed: [3dd0a9a](https://github.com/DeFiCh/ain/pull/1650/commits/3dd0a9ac8e61821361475ebc4ed656e67897abf0)

src/masternodes/mn_checks.cpp
```diff
@@ -1111,8 +1112,13 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
                 if (!minted)
                     return minted;
 
-                CalculateOwnerRewards(*mintable.val);
-                auto res = mnview.AddBalance(*mintable.val, CTokenAmount{tokenId, amount});
+                CScript mintTo{*mintable.val};
+                if (height >= grandCentralNextHeight && !obj.to.empty()) {
```
Bushstar, line 1116: Need to reject a populated to address below the fork height. The problem is that new clients will accept the TX with the to address below fork height and old clients will not be able to deserialise the TX as it will not have the to field. For compatibility we want new clients to reject use of the to field below the fork height, not just ignore it. After the fork such a check can be removed when all users are on the new version.

Also, let's check that to address is valid, extract the destination using ExtractDestination(), if that fails then reject the TX, after call IsValidDestination() on the destination and if that fails reject the TX. We have several places were we just trust whatever the script address is, but we really should be checking that addresses are valid and that users are not sending to junk destinations.

shohamc1: Fixed: https://github.com/DeFiCh/ain/pull/1650/commits/7f09763b10e6fa022abfe4f676bcd493eab6bb7f

Bushstar (approved):

Mixa84 (dismissed):
But I would still put `to` field only in json object. Not as regular parameter.

PR #1654, opened 2022-12-14
On-chain governance fixes

Jouzo:

/kind fix

- Remove proposalToJSON overwriting proposal status
- Add proposal MAX_CYCLES to attributes

### DIFF
--- a/src/masternodes/govvariables/attributes.cpp
+++ b/src/masternodes/govvariables/attributes.cpp
@@ -205,6 +205,7 @@ const std::map<uint8_t, std::map<std::string, uint8_t>> &ATTRIBUTES::allowedKeys
              {"voc_approval_threshold", GovernanceKeys::VOCApprovalThreshold},
              {"quorum", GovernanceKeys::Quorum},
              {"voting_period", GovernanceKeys::VotingPeriod},
+             {"cfo_max_cycles", GovernanceKeys::CFPMaxCycles},
          }},
     };
     return keys;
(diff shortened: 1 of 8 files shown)


src/masternodes/rpc_proposals.cpp
```diff
@@ -9,64 +9,57 @@ struct VotingInfo {
...
     if (isVotingInfoAvailable) {
-        votesPresent = votingInfo->votesPresent;
-        votesYes = votingInfo->votesYes;
+        votesPresent  = votingInfo->votesPresent;
+        votesYes      = votingInfo->votesYes;
         votesPossible = votingInfo->votesPossible;
 
         votesPresentPct = lround(votesPresent * 10000.f / votesPossible);
-        auto valid = votesPresentPct > quorum;
+        auto valid      = votesPresentPct > quorum;
```
Mixa84, line 56: I would remove this valid too, and just check if votesPresent > 0 and always show `VotesYesPct` then!

Jouzo: Removed the `valid` check.
`votesPresent` > 0 is already handled here https://github.com/DeFiCh/ain/blob/221a8489a632098f1dbb1cc1af84de98aba369c6/src/masternodes/rpc_proposals.cpp#L629

Mixa84 (dismissed):

src/masternodes/govvariables/attributes.cpp
```diff
@@ -205,6 +205,7 @@ const std::map<uint8_t, std::map<std::string, uint8_t>> &ATTRIBUTES::allowedKeys
              {"voc_approval_threshold", GovernanceKeys::VOCApprovalThreshold},
              {"quorum", GovernanceKeys::Quorum},
              {"voting_period", GovernanceKeys::VotingPeriod},
+             {"cfo_max_cycles", GovernanceKeys::CFPMaxCycles},
```
Jouzo, line 208:
```suggestion
             {"cfp_max_cycles", GovernanceKeys::CFPMaxCycles},
```

PR #1659, opened 2022-12-19
Fix accounting of burntoken tx in getburninfo

Mixa84:

/kind fix

Account tokens that are not burnt for a consortium member in `tokens` category of `getburninfo`. MAke `from` parameter in `burntokens` rpc as optional when consortium member is burning tokens.


src/masternodes/rpc_accounts.cpp
```diff
@@ -1919,6 +1920,18 @@ UniValue getburninfo(const JSONRPCRequest& request) {
 
     burnView->ForEachAccountHistory(calculateBurnAmounts);
 
+    for (auto const & token: nonConsortiumTokens.balances)
+    {
+        CDataStructureV0 liveKey = {AttributeTypes::Live, ParamIDs::Economy, EconomyKeys::ConsortiumMinted};
+        auto balances = attributes->GetValue(liveKey, CConsortiumGlobalMinted{});
```
Bushstar, line 1926: Create liveKey and get balances before loop.

src/masternodes/rpc_tokens.cpp
```diff
@@ -806,18 +806,43 @@ UniValue burntokens(const JSONRPCRequest& request) {
...
+                for (const auto &member : members) {
+                    if (IsMineCached(*pwallet, member.second.ownerAddress)) {
+                        burnedTokens.from = member.second.ownerAddress;
+                        break;
+                    }
+                }
+            }
+        }
+        if (burnedTokens.from.empty())
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameters, argument \"from\" must not be null");
```
shohamc1, line 836: I think this would be much clearer to users.

```suggestion
            throw JSONRPCError(RPC_INVALID_PARAMETER, "No valid addresses could be found, use the \"from\" argument to set address to burn from");
```

Bushstar (approved):

PR #1664, opened 2022-12-22
Allow any reward address after GCN

Bushstar:
OnChain MN reward address is currently P2PKH or P2WPKH, this PR allows script address types.


prasannavl:
Guarded. So adding for next release to get this merged.

PR #1666, opened 2022-12-23
Add multi-account and multi-TX filters to `accounthistorycount`

### LINKED ISSUE #1665: Add multi-address, multi-txtypes filter support for `accounthistorycount` RPC

As mentioned in the title, It would be better to have multiple addresses and multiple transaction types filtering added into the `accounthistorycount` RPC.

#### Why is this needed:

When we need to query tx counts for multiple addresses or txtypes, with the current implementation we need to have multiple RPC calls and aggregate those data to find totals. So performance wise it would be better to have this abstracted into the rpc to avoid unncessory RPC calls, loops and aggregations on higher code levels.

shohamc1:

/kind feature

- Adds a txtypes argument to allow filtering multiple CustomTxTypes
- Allow owner to be an array, returns results for all addresses in the array

Fixes #1665


shohamc1:
Ignore Jellyfish failure - flaky test.

Mixa84:
@shohamc1 Can we have just a short test which verify that multi-tx filters work?

shohamc1:
Waiting for https://github.com/DeFiCh/ain/pull/1643 to get merged so that we can use [feature_listaccounthistory_multiaccountquery.py](https://github.com/DeFiCh/ain/pull/1643/files#diff-0bfd96c7f1de5e125364f6fe8d5618af0bdf30ba9787b9eb0b6e59f42ba08d96) for the test.

Mixa84 (dismissed):

src/masternodes/rpc_accounts.cpp
```diff
@@ -1538,69 +1574,77 @@ UniValue accounthistorycount(const JSONRPCRequest& request) {
...
-    CScript lastOwner;
     uint64_t count = 0;
-    auto lastHeight = uint32_t(::ChainActive().Height());
-    const auto currentHeight = lastHeight;
 
-    auto shouldContinueToNextAccountHistory = [&](AccountHistoryKey const & key, AccountHistoryValue value) -> bool {
-        if (!owner.empty() && owner != key.owner) {
-            return false;
-        }
+    for (const auto &owner : accountSet) {
```
Bushstar, line 1579: Each loop with owner then loops over the account history with ForEachAccountHistory. Ideally we only want to loop over the entire account history with ForEachAccountHistory once. Can we update shouldContinueToNextAccountHistory to check if accountSet is not empty and if so check if the current owner is in that set?

shohamc1:
Related Jellyfish PR: https://github.com/JellyfishSDK/jellyfish/pull/1943

Bushstar (dismissed):

PR #1667, opened 2022-12-23
Add -leveldbchecksum argument

dcorral:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:

Adds -leveldbchecksum argument.

PR #1669, opened 2022-12-23
Adds -leveldebchecksum argument

dcorral:

/kind fix

Adds -leveldbchecksum argument so leveldb does not do checksum verification.
Possible values: 
- auto: if it's a masternode then it will be set to true, false otherwise
- false
- true

#### Additional comments?:

Clean history over #1667


src/init.cpp
```diff
@@ -944,6 +944,20 @@ void InitParameterInteraction()
         if (gArgs.SoftSetBoolArg("-whitelistrelay", true))
             LogPrintf("%s: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1\n", __func__);
     }
+
+    // Parse leveldb checksum
+    auto leveldbchecksum = gArgs.GetArg("-leveldbchecksum", "auto");
```
prasannavl, line 949: @dcorral can you just set a global variable at this point? Instead of using `SoftSetBoolArg` and re-setting these over and over?

Bushstar: Not needed, this is called in InitParameterInteraction() which is only called once during start up of the node.

prasannavl: To clarify: Store the result of https://github.com/DeFiCh/ain/pull/1669/files#diff-44bef8fa93d168379ef331d9e11d333be06ce98a8f37ec65f8a987846b90a151R136 in global. And keep setting `gArgs` manually. Let gArgs be the source of truth of what the user provides.

Bushstar: Understood, not for this function but use of -leveldbchecksum across the code base. Will update.

prasannavl (approved):

PR #1682, opened 2023-01-03
SPV: update testnet seed nodes

Bushstar:
Update the SPV DNS seeds for testnet as all bar one of the existing ones were non-responsive.

### DIFF
--- a/src/spv/bitcoin/BRChainParams.cpp
+++ b/src/spv/bitcoin/BRChainParams.cpp
@@ -33,8 +33,8 @@ static const char *BRMainNetDNSSeeds[] = {
 };
 
 static const char *BRTestNetDNSSeeds[] = {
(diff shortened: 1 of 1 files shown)


prasannavl:
Running all them through dig. 

Both old and new *.petertodd.org seems to the most unreliable, while testnet-seed.breadwallet.com seems to work quite well across many different regions. 

```
$ dig seed.tbtc.petertodd.org

; <<>> DiG 9.18.4-2ubuntu2-Ubuntu <<>> seed.tbtc.petertodd.org
;; global options: +cmd
;; connection timed out; no servers could be reached

```

Perhaps, we should remove petertodd ones for the moment as workaround until we can address the bread wallet connection code to be more graceful and resilient to unreliable domains.

Jouzo (approved):

PR #1697, opened 2023-01-17
Fix cycleEndHeight for emergency VOC

Mixa84:

/kind fix

Fix the calculation of `cycleEndHeight` field for emergency VOC


prasannavl:
Just leaving a note here for future context: 

- `cycleEndHeight` is used only on the RPCs and never on the consensus so far, so it's a safe to update even though it's code that's called consensus.

PR #1705, opened 2023-01-25
Add lambda support to Require

Bushstar:
Allows passing of lambdas to Require which can avoid premature string creation.


Jouzo (approved):

prasannavl:
LGTM

Issue #1704, opened 2023-01-25
Neutral vote is mistakenly treated the same as a no-vote for on-chain gov


uzyn:
## Caution

As on-chain governance is activated soon and likely before this consensus error is able to be fixed, **do not cast neutral votes** before this bug is addressed or it would be counted as a no-vote.

#### What happened:

Neutral votes for CFP and VoC are counted the same as no-votes, it should not be that way.

Reference:
https://github.com/DeFiCh/ain/blob/0989ff6fde4530a83fa2c3649c2642615e36714d/src/masternodes/validation.cpp#L2235-L2238

Example:
- 1000 yesses, 200 nos, 5000 neutrals are currently mistakently marked as a _failed_ CFP/DFIP.

#### What you expected to happen:

Neutral votes for CFP and VoC should not be counted as no-votes.

Example:
- 1000 yesses, 200 nos, 5000 neutrals should be counted as a _passing_ CFP/DFIP.

PR #1709, opened 2023-01-26
Disable neutral votes from RPC

Mixa84:

/kind fix

Disable temporary neutral votes for proposals until next hardfork. Support for issue #1704.


dcorral (dismissed):

src/masternodes/rpc_proposals.cpp
```diff
@@ -457,8 +459,10 @@ UniValue votegov(const JSONRPCRequest &request) {
         vote = CProposalVoteType::VoteNo;
     } else if (voteStr == "yes") {
         vote = CProposalVoteType::VoteYes;
-    } else if (voteStr != "neutral") {
+    } else if (gArgs.GetBoolArg("-rpc-governance-accept-neutral", DEFAULT_RPC_GOV_NEUTRAL) && voteStr != "neutral") {
```
prasannavl, line 462: Minor tweak: Let's just read the gArgs once upto top and make the casing a bit simpler?

PR #1712, opened 2023-01-27
Remove positive amount on close vault

Bushstar:
When closing a vault that the remaining balances is negated by the interest delete that remaining balance.


src/init.cpp
```diff
@@ -1756,6 +1773,8 @@ bool AppInitMain(InitInterfaces& interfaces)
                 // Ensure we are on latest DB version
                 pcustomcsview->SetDbVersion(CCustomCSView::DbVersion);
 
+                EraseLoanAmountsInClosedVaults(*pcustomcsview);
```
prasannavl, line 1776: Instead of startup, what do you think about running it once per block on each block with a close vault tx? More reactive and will keep all nodes in sync.

src/masternodes/mn_checks.cpp
```diff
@@ -2839,6 +2839,11 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
 
                 Require(amount + totalInterest <= 0, "Vault <%s> has loans", obj.vaultId.GetHex());
 
+                // If there is an amount negated by interested remove it from loan tokens.
+                if (amount > 0) {
+                    mnview.SubLoanToken(obj.vaultId, {tokenId, amount});
+                }
+
```
prasannavl, line 2846: Even though, this doesn't directly impact consensus, for now, would just comment it to be removed later and guard it just to avoid other unforeseen side effects.

Jouzo (dismissed):

Mixa84 (dismissed):

prasannavl (approved):

prasannavl:
LGTM

PR #1713, opened 2023-01-27
Test framework: Initial impl of exhaustive state verification with rollback

dcorral:

/kind fix

Adds `capture_rollback_verify` decorator to DefiTestFramework. 

Any member function inside a given DefiTestFramework child class can now use `@DefiTestFramework.capture_rollback_verify` decorator. This decorator will capture the state of the chain, run a rollback and verify the data is properly restored after the function execution.

Usage example
```python
from test_framework.test_framework import DefiTestFramework

class ExampleTest(DefiTestFramework):
    
    def test_scenario(self):
        # test logic   

    @DefiTestFramework.capture_rollback_verify
    def test_scenario_with_rollback(self):
        self.test_scenario()

    def run_test(self):
        self.test_scenario()
        self.test_scenario_with_rollback()

if __name__ == '__main__':
    ExampleTest().main()
```

WARNING: This decorator when used in large tests can become very expensive. The following commands are run to gather the data to be verified on rollback:

```python
[
    self.nodes[0].logaccountbalances(),
    self.nodes[0].logstoredinterests(),
    self.nodes[0].listvaults(),
    self.nodes[0].listtokens(),
    self.nodes[0].listgovs(),
    self.nodes[0].listmasternodes(),
    self.nodes[0].listaccounthistory(),
    self.nodes[0].getburninfo(),
    self.nodes[0].getloaninfo(),
    self.nodes[0].listanchors(),
    self.nodes[0].listgovproposals(),
    self.nodes[0].listburnhistory(),
    self.nodes[0].listcommunitybalances()
]
```


prasannavl:
Could you please update the description to reflect the changes, and also add some in code comments on the function on the final state that you expect before it's usable? Along with a note that it's currently an expensive one to be called and that use caution if used before the final state impl is finished. 

Happy to merge it after that to prevent it from becoming stale.

dcorral:
Waiting on fix to #1788 responsible for consortium test fail

prasannavl (approved):

Issue #1536, opened 2022-11-01
Missing RPC data for PaybackWithCollateral


sambP:
If I'm not missing something, there isn't any way to check what really happened in the vault after a PaybackWithCollateral. This breaks our whole balance calculation. Wouldn't it be possible to add at least a vaultSnapshot, like after an auction?

shohamc1:
I'm not too sure on what data seems to be missing, could you provide an example? How does it break the balance calculation?

sambP:
Sure, here is an (to me unrelated...) example:

This is a paybackAsCollateral tx: https://defiscan.live/transactions/5dee7ca7019137fee6ccb20f2589c53bdddaf7f6d233bea4e27c2faf50a312e3

The customTx looks like this:
`{"valid":true,"type":"PaybackWithCollateral","results":{"vaultId":"591882453d1c682bbc1aaaddaf5b0b3801416e1089386b99581c28f28dbb5bfc"}}`

listvaulthistory 591882453d1c682bbc1aaaddaf5b0b3801416e1089386b99581c28f28dbb5bfc blockHeight: 2258841
no response

listaccounthistory dTGeTGkayryJFi1tc6ERNnsL7eo4G5CbaZ blockHeight: 2258841
no response

There is no information about how much collateral in the vault repays the how much loan. So it is currently absolutely impossible to collect historical changes in a vault if this feature was used.

sandstumpen:
@shohamc1
Is there any new information here yet? This function is very important for me for the creation of tax reports.
Thx

pegasus2003:
Sounds like a great idea

phigo90:
I would even go one step further. The complete solution would be to track the incurred interest in a kind of vault history. Similar to how it is currently done with listaccounthistory for e.g. the LM rewards.

RomanShumkov:
> I would even go one step further. The complete solution would be to track the incurred interest in a kind of vault history. Similar to how it is currently done with listaccounthistory for e.g. the LM rewards.

Feel free to give thumbs up here: https://github.com/DeFiCh/ain/issues/957

prasannavl:
Quick update: We're looking into this. Will add more details soon (hopefully next week).

PR #1715, opened 2023-01-30
Track Add/SubVaultCollateral in vault history

Bushstar:
With this PR whenever AddVaultCollateral or SubVaultCollateral is called and the vault index is enabled entries will now be added to the vault index. When withdrawing or depositing collateral the change to an address is shown as well as the change to the collateral showing up as address 'vaultCollateral'. This then will provide a history of what happens to collateral during the auction process and when the payback with collateral transaction is used. This solves https://github.com/DeFiCh/ain/issues/1536.


sambP:
Thanks a lot!

PR #1717, opened 2023-01-30
rpc: votegov: Add support for owner and operator address

DocteurPing:
### Summary

Enables support for address support on `votegov` to vote with an address in addition to the currently accepted masternode ID.  

### How

- Over the wire, TX still uses masternode ID.
- The RPC API now auto selects the masternode ID based on the address provided. 
- The selection order: 
  - MasternodeID
  - OwnerAdress
  - OperatorAddress

### Limitations

- Throws an error when there is no reasonable choice available (invalid address/masternodeId, address does not own/operate a masternode).


src/masternodes/rpc_proposals.cpp
```diff
@@ -449,7 +449,8 @@ UniValue votegov(const JSONRPCRequest &request) {
     RPCTypeCheck(request.params, {UniValue::VSTR, UniValue::VSTR, UniValue::VSTR, UniValue::VARR}, true);
 
     auto propId = ParseHashV(request.params[0].get_str(), "proposalId");
-    auto mnId   = ParseHashV(request.params[1].get_str(), "masternodeId");
+    std::string id = request.params[1].get_str();
+    uint256 mnId = uint256();
```
Bushstar, line 453: Note that you only need as uint256 will default initialise itself as the class has a default constructor.

`uint256 mnId;`

src/masternodes/rpc_proposals.cpp
```diff
@@ -474,9 +475,30 @@ UniValue votegov(const JSONRPCRequest &request) {
             throw JSONRPCError(RPC_INVALID_PARAMETER,
                                strprintf("Proposal <%s> is not in voting period", propId.GetHex()));
         }
+        if (id.length() == 64) {
+            mnId = ParseHashV(id, "masternodeId");
+        }
         auto node = view.GetMasternode(mnId);
         if (!node) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("The masternode %s does not exist", mnId.ToString()));
+            if (id.length() == 34) {
```
Bushstar, line 483: Address can be 26-35 chars long for P2PKH and 42 for P2WPKH. I'd work on getting the correct mnId before calling GetMasternode.

src/masternodes/rpc_proposals.cpp
```diff
@@ -474,9 +475,27 @@ UniValue votegov(const JSONRPCRequest &request) {
...
+        } else {
+            CTxDestination dest = DecodeDestination(id);
+            if (!IsValidDestination(dest)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER,
+                                    strprintf("The masternode id or address is not valid: %s", id));
+            }
+            const CKeyID ckeyId = dest.index() == PKHashType ? CKeyID(std::get<PKHash>(dest)) : CKeyID(std::get<WitnessV0KeyHash>(dest));
+            auto masterNodeIdByOwner = view.GetMasternodeIdByOwner(ckeyId);
+            if (!masterNodeIdByOwner) {
+                auto masterNodeIdByOperator = view.GetMasternodeIdByOperator(ckeyId);
```
Bushstar, line 489: Check that `masterNodeIdByOperator` is valid or you will get a `std::bad_optional_access` thrown on `masterNodeIdByOperator.value()`. I'd structure it like the following:

```
if (auto masterNodeIdByOwner = view.GetMasternodeIdByOwner(ckeyId)) {
    mnId = masterNodeIdByOwner.value();
} else if (auto masterNodeIdByOperator = view.GetMasternodeIdByOperator(ckeyId)) {
    mnId = masterNodeIdByOperator.value();
}
```

src/masternodes/rpc_proposals.cpp
```diff
@@ -474,9 +475,26 @@ UniValue votegov(const JSONRPCRequest &request) {
...
         }
+        if (id.length() == 64) {
+            mnId = ParseHashV(id, "masternodeId");
+        } else {
+            CTxDestination dest = DecodeDestination(id);
+            if (!IsValidDestination(dest)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER,
+                                    strprintf("The masternode id or address is not valid: %s", id));
+            }
+            const CKeyID ckeyId = dest.index() == PKHashType ? CKeyID(std::get<PKHash>(dest)) : CKeyID(std::get<WitnessV0KeyHash>(dest));
```
Bushstar, line 486: We need to make sure that the provided address is either P2PKH or a P2WPKH, otherwise if someone passes a different variant then `std::get` will throw a `std::bad_variant_access`.

```
            const CKeyID ckeyId  = dest.index() == PKHashType ? CKeyID(std::get<PKHash>(dest))
                                 : dest.index() == WitV0KeyHashType ? CKeyID(std::get<WitnessV0KeyHash>(dest))
                                 : CKeyID();
```

Bushstar (approved):

src/masternodes/rpc_proposals.cpp
```diff
@@ -479,9 +480,27 @@ UniValue votegov(const JSONRPCRequest &request) {
...
         }
+        if (id.length() == 64) {
+            mnId = ParseHashV(id, "masternodeId");
+        } else {
+            CTxDestination dest = DecodeDestination(id);
+            if (!IsValidDestination(dest)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER,
+                                    strprintf("The masternode id or address is not valid: %s", id));
+            }
+            const CKeyID ckeyId = dest.index() == PKHashType ?
```
shohamc1, line 491: A bit too hard to read IMO, better to break it down into `if`s.

src/masternodes/rpc_proposals.cpp
```diff
@@ -417,7 +417,7 @@ UniValue votegov(const JSONRPCRequest &request) {
         "\nVote for community proposal" + HelpRequiringPassphrase(pwallet) + "\n",
         {
                                                     {"proposalId", RPCArg::Type::STR, RPCArg::Optional::NO, "The proposal txid"},
-                                                    {"masternodeId", RPCArg::Type::STR, RPCArg::Optional::NO, "The masternode id which made the vote"},
+                                                    {"masternodeId/ownerAddress/operatorAdress", RPCArg::Type::STR, RPCArg::Optional::NO, "The masternode id / owner address / operator address which made the vote"},
```
shohamc1, line 420: Better to leave the arg name unchanged for users using named args, or if we need to use an object to pass args in the future.
 
```suggestion
                                                    {"masternodeId", RPCArg::Type::STR, RPCArg::Optional::NO, "The masternode id / owner address / operator address which made the vote"},
```

PR #1718, opened 2023-01-31
Do not allow non-DAT consortium attributes to be set

shohamc1:

/kind fix


src/masternodes/mn_rpc.cpp
```diff
@@ -551,8 +551,12 @@ UniValue setgov(const JSONRPCRequest& request) {
                 const auto attrMap = attributes->GetAttributesMap();
                 for (const auto& [key, value] : attrMap) {
                     if (const auto attrV0 = std::get_if<CDataStructureV0>(&key)) {
-                        if (attrV0->type == AttributeTypes::Consortium && (attrV0->typeId == 0 || pcustomcsview->GetLoanTokenByID({attrV0->typeId}))) {
-                            throw JSONRPCError(RPC_INVALID_REQUEST, "Cannot set consortium on DFI or loan tokens");
+                        DCT_ID tokenID{attrV0->typeId};
+                        if (attrV0->type == AttributeTypes::Consortium &&
+                            (attrV0->typeId == 0 ||
+                             !pcustomcsview->GetToken(tokenID)->IsDAT() ||
```
Bushstar, line 557: Do not use `->IsDAT()` unless you know a valid object has been returned or we will segfault.

src/masternodes/mn_rpc.cpp
```diff
@@ -743,6 +747,27 @@ UniValue setgovheight(const JSONRPCRequest& request) {
...
+            }
+
+            LOCK(cs_main);
+            const auto attrMap = attributes->GetAttributesMap();
+            for (const auto& [key, value] : attrMap) {
+                if (const auto attrV0 = std::get_if<CDataStructureV0>(&key)) {
+                    DCT_ID tokenID{attrV0->typeId};
+                    if (attrV0->type == AttributeTypes::Consortium &&
+                        (attrV0->typeId == 0 ||
+                         !pcustomcsview->GetToken(tokenID)->IsDAT() ||
```
Bushstar, line 764: Do not use `->IsDAT()` unless you know a valid object has been returned or we will segfault.

prasannavl:
While consortium likely will have larger changes, merging this so it doesn't get stale and can be addressed as a unit later.

PR #1719, opened 2023-01-31
Fix setgov crashes

shohamc1:

/kind fix

When a wallet tries to use `setgov`, `unsetgov` or `setgovheight`, the code crashes due to a bad access error instead of throwing an error. This is because `auths` remains empty since no addresses in the wallet has permission to send these transactions. We add a guard against `ExtractDestination` and allow the TX to be rejected during `execTestTx` instead.


Bushstar (approved):

prasannavl:
Great. Possible to address this on `ExtractDestination` so it's more holistic? (would ideally have been better at `Solver` too, but doesn't look like that's possible without changing more BTC related code)

PR #1723, opened 2023-02-01
Add rollback diff to full sync CI

Jouzo:

/kind feature

- Output all the currently log command to a debug-pre-rollback.log file at START_BLOCK
- Syncs the 50k range
- Rollbacks to START_BLOCK and re-run all the log command and outputs it to debug-post-rollback.log
- Add a CI step where we diff debug-pre-rollback.log against debug-pot-rollback.log

Would have caught re-org issue fixed in https://github.com/DeFiCh/ain/pull/1641

See run https://github.com/DeFiCh/ain/actions/runs/4058261206/jobs/6985038469#step:10:20 with revert of PR 1641 
and run https://github.com/DeFiCh/ain/actions/runs/4058262361/jobs/6985091099#step:10:20 with PR 1641


.github/workflows/fullsync-tests.yml
```diff
@@ -129,10 +129,17 @@ jobs:
...
-      run: ./sync_then_diff.sh
+    - name: Sync
+      run: ./sync.sh
+
+    - name: Diff log
+      run: diff debug-${{matrix.blocks.stop}}.log debug-tmp-${{matrix.blocks.stop}}.log
+
+    - name: Diff rollback log
+      run: diff debug-pre-rollback.log debug-post-rollback.log
+      if: ${{ failure() ||  success() }}
```
prasannavl, line 142: Need this?

prasannavl:
This is neat. 

A few questions:

- Could we avoid an `invalidateblock` in the beginning, even if it's the same start block?
- Don't recall the cause for the edge case of block being stuck and reinvalidation required. Wonder if it's related to #1700 - if it is, we can clean up in addition to this.

.github/workflows/fullsync-tests.yml
Jouzo: Not strictly needed but let's it run even if we have a failure on previous diff log.
It let's us have both diff results but will still mark the sync as failed from previous diff failure (whereas `continue-on-error` would mark it as passed).

Jouzo:
I will do a run without the `reconsiderblock` and check if it has been fixed by #1700

prasannavl (approved):

PR #1752, opened 2023-02-09
Disconnect nodes in rollback_to

Bushstar:
In rollback_to if multiple nodes are connected disconnect them first before rolling back and the restore the connections again afterwards.

### DIFF
--- a/test/functional/feature_on_chain_government_fee_distribution.py
+++ b/test/functional/feature_on_chain_government_fee_distribution.py
@@ -7,9 +7,7 @@
 
 from test_framework.test_framework import DefiTestFramework
 from test_framework.util import (
-    assert_equal,
-    connect_nodes_bi,
-    disconnect_nodes,
+    assert_equal
 )
 from decimal import ROUND_DOWN, Decimal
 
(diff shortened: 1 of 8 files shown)


Mixa84 (approved):

Jouzo (dismissed):

prasannavl (dismissed):

Jouzo (approved):

test/functional/feature_on_chain_government_fee_distribution.py
```diff
@@ -120,19 +118,7 @@ def test_cfp_fee_distribution(self, amount, expectedFee, burnPct, vote, cycles=2
...
-        assert_equal(self.nodes[3].getconnectioncount(), 0)
-
-        # Rollback nodes in isolation
-        for i in range(self.num_nodes):
-            self.rollback_to(height, nodes=[i])
-
-        # Connect nodes
-        for i in range(self.num_nodes - 1):
-            connect_nodes_bi(self.nodes, i, i + 1)
+        self.rollback_to(height, nodes=[0, 1, 2, 3])
```
Jouzo, line 121: What about changing `rollback_to` node parameter to be actual nodes instead of indexes.
Most use case would then be `self.rollback_to(height)` with default = self.nodes.

This is consistent with `sync_mempools` and `sync_blocks`.

```
    def sync_blocks(self, nodes=None, **kwargs):
        sync_blocks(nodes or self.nodes, **kwargs)

    def sync_mempools(self, nodes=None, **kwargs):
        sync_mempools(nodes or self.nodes, **kwargs)

```

prasannavl: Like this suggestion. @Bushstar thoughts?

prasannavl (approved):

PR #1755, opened 2023-02-10
Replace Require strings with lambdas

Bushstar:
The adding of Require impacted performance as strings were always created regardless of whether the conditional was true or false. This PR updates all Require statements not with the exception of mn_checks.

### DIFF
--- a/src/amount.h
+++ b/src/amount.h
@@ -92,7 +92,7 @@ typedef std::map<DCT_ID, CAmount> TAmounts;
 
 inline ResVal<CAmount> SafeAdd(CAmount _a, CAmount _b) {
     // check limits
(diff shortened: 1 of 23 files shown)


Jouzo (dismissed):

prasannavl:
Leaving these for context:

- https://godbolt.org/z/MzMhrcPoh

image

We shouldn't have any observable difference from lambas on regular checks that isn't run through expensive calculations like decimal conversions, since it's effectively the same (if anything an additional function call, but that would be micro-opt that we don't need to be concerned about here) 

Merging for now, since it's already reviewed and verified, on all tests and PR does cover some impactful areas. However, need to evaluate long term maintainability of this later.

PR #1756, opened 2023-02-10
Update voting period on start of new proposal cycle

Jouzo:

/kind fix

- Updates proposal voting period from attributes when passing to next cycle

Fixes https://github.com/DeFiCh/team/issues/43

#### Constraints

This implementation does not guard the change behind a fork guard as this change is restricted to multi-cycles proposals only. 
As of now (block `2680000`), `cfp_max_cycles` govvar is restricted to 1 and `UpdateProposalCycle` is never called.
As long as the `cfp_max_cycles` govvar value is kept at 1 and multi-cycles are not activated, this PR can be merged without fork guard.
If `cfp_max_cycles` were to be updated before next hard fork, this PR would have to be updated. 

Target is `v3.3.0` or next mandatory upgrade


prasannavl:
Can we guard this under next network upgrade? Then we can get this merged.

PR #1762, opened 2023-02-15
Multi-threaded getburninfo

Bushstar:
Get burn values from the burn data based simultaneously across multiple threads by breaking the current height up into chunks.


prasannavl (approved):
Excellent. Looks great.

Let's merge and then I'll add some worker generalisation soon to recycle threads, and we should be good to go.

prasannavl:
Jellyfish still on failures due to Ubuntu 18 deprecation. Ignoring and merging.

PR #1764, opened 2023-02-16
Limit parallel script processing thread pool to 4 by default

prasannavl:
/kind chore

- Limits parallel execution threads of Bitcoin threads to 4 by default.
- We do not utilise this for DeFi TXs (unfortunately), and usually is just a wasted threads sitting around unless there are many many pure UTXO TXs. 4 threads should be sufficient for most practical purposes. 
- Doing this cleanup as a precursor to add a common worker pool for most other thread related activites

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1473,13 +1473,14 @@ bool AppInitMain(InitInterfaces& interfaces)
 
     int script_threads = gArgs.GetArg("-par", DEFAULT_SCRIPTCHECK_THREADS);
     if (script_threads <= 0) {
-        // -par=0 means autodetect (number of cores - 1 script threads)
-        // -par=-n means "leave n cores free" (number of cores - n - 1 script threads)
-        script_threads += GetNumCores();
         // DeFiChain specific:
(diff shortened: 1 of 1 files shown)


Bushstar (approved):

Bushstar:
Perhaps it is better to set DEFAULT_SCRIPTCHECK_THREADS to 4?

Then users can still override and set to 0 to let the client use GetNumCores(). This change will always cap the number of cores to 4 regardless of user settings.

prasannavl:
Good catch. Agreed. Switching to use `DEFAULT_SCRIPTCHECK_THREADS`

prasannavl:
Overlooked earlier - users can already override. This only changes the behaviour of the `auto (0)` config. Rest follow the same. Cannot change script default, since that's the one that will have a behaviour change across different hardware and also requires turning it into a non-const.

prasannavl:
Ignores the additional reverted commits - initial is the most optimal version, I believe.

Bushstar (approved):

Issue #1763, opened 2023-02-15
OCG Voting rewards are not spendable


bernd-mack:
#### What happened:
voted with onchain governance and got reward on my address but wallet does not show tokenbalance and therefore the DFI are not spendable

#### What you expected to happen:
rewards from onchainvoting are listed in gettokenbalances and are spendable

#### How to reproduce it (as minimally and precisely as possible):
vote for a DFIP/CFP and get a reward. Reward is shown on defiscan but not on local wallet

#### What are your environment parameters:
defichain desktop wallet with latest noderelease

#### Anything else we need to know?:
restore with snapshot did not help and another telegram with multiple masternodes told that only one reward is shown in balance.

shohamc1:
Hey, thanks for opening this. It seems that the issue here is the token not showing up on `gettokenbalances`, which should be fixed by the linked PR. 

By my testing, rewards are spendable and can be viewed using `getaccount` in the meanwhile.

bernd-mack:
Thanks and sorry, the wording was not clear (German->English gets lost or does not arrive exactly with me from time to time a little bit of information) :(
Spend if you build the transactions themselves or use account2utxos were not meant, only the gettokenbalances and thereby also the desktop wallet which presumably falls back on it and there with a 0 not automatically converts the tokens if you want to send something.
=> They are spendable, just not automatically when using desktop wallet

PR #1766, opened 2023-02-16
Add missing CalculateOwnerRewards calls

Bushstar:
Before SubBalance is called CalculateOwnerRewards should also be called to make sure the expected pool rewards are present. When AddBalance is called on a new address that might not have been seen by the network before CalculateOwnerRewards should be called to add that address to the account index.

This PR adds one missing call to CalculateOwnerRewards before SubBalance in the currently unused DFIP2201 smart contract. There are also a missing CalculateOwnerRewards calls for AddBalance on CWithdrawFromVaultMessage.

Now that we know that all other SubBalance calls that are currently in use have a corresponding CalculateOwnerRewards call we can remove the fork guards around other calls to this function that were restored in a more cautious manner.

Resolves:
https://github.com/DeFiCh/ain/issues/1763


prasannavl:
Should test the performance impact of this on the block, just to get an estimate before merging.

Jouzo (approved):

PR #1767, opened 2023-02-17
Make alpine dockerfile build statically

DocteurPing:

/kind fix

Alpine dockerfile was not compiling statically. the configuration of the docker file was changed in order to make it compile the binary statically  so that the binary can be run on any linux machine. 

Fixes #


prasannavl:
This uses the depends build. Ideally, depends shouldn't be required on alpine, if everything is built statically.

prasannavl:
Closing stale PRs

PR #1770, opened 2023-02-20
Add checkpoint for maintest and testnest

DocteurPing:

/kind chore

Adding a new checkpoint in the code for mainet and testnet


prasannavl:
cc/ @Bushstar for validation of the correct testnet chain (post recent rollback due to consortium)

PR #1775, opened 2023-02-23
Fix for multi-threaded getburninfo

prasannavl:
/kind fix

- Fixes the last segment issue. 
- Additionally, switches to a worker number agnostic queuing.


src/masternodes/threadpool.h
```diff
@@ -19,7 +19,7 @@ class TaskPool {
     public:
         explicit TaskPool(size_t size);
         void Shutdown();
-        [[nodiscard]] size_t GetAvailableThreads() const { return size; }
+        size_t GetAvailableThreads() { return size; }
```
DocteurPing, line 22: Why were the attributes removed ?

prasannavl: `const` is removed because this is a more abstract API. Will likely not be const later as it will return dynamic value of threads. Can probalby add nodiscard.

DocteurPing (dismissed):

Bushstar (approved):

DocteurPing (approved):

PR #1794, opened 2023-03-03
Fix getburninfo by removing consortium tokens

Mixa84:

/kind fix

Remove consortium tokens from getburninfo as the delta calc was incorrect. As consortium won't be activated for a while, we'll return on this before consortium will get activated!


Jouzo (approved):

prasannavl:
Closing in favor of https://github.com/DeFiCh/ain/pull/1796

PR #1795, opened 2023-03-03
Replace CLockFreeGuard with AtomicMutex 

prasannavl:
/kind fix

- `CLockFreeGuard` uses barrier that's unsuitable for general use case, and results in multiple threads entering critical sections due to premature barrier optimizations
- Additionally, remove the 1ms thread sleep, which on platforms like Linux which default to higher resolution thread quantums might be palatable, but will have heavy regressive behaviour on platforms like Windows which default to 15ms quantum slices.
- Since CLockFree is primarily used for extremely cheap ops, spin waiting is just simpler // though likely have to reassess it's use cases later on if it's even useful here.

### DIFF
--- a/src/masternodes/threadpool.cpp
+++ b/src/masternodes/threadpool.cpp
@@ -31,12 +31,12 @@ void ShutdownDfTxGlobalTaskPool() {
 
 
 void TaskGroup::AddTask() { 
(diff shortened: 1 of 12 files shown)


prasannavl:
Closed in favor of https://github.com/DeFiCh/ain/pull/1797

PR #1797, opened 2023-03-06
Fix CLockFreeGuard, introduce AtomicMutex

prasannavl:
/kind fix

- `CLockFreeGuard` uses barrier that's unsuitable for general use case, and results in multiple threads entering critical sections due to premature barrier optimizations. This fixes it. 
- Additionally, the 1ms thread sleep isn't helpful, which on platforms like Linux which default to higher resolution thread quantums might be palatable, but will have heavy regressive behaviour on platforms like Windows which default to 15ms quantum slices.
- Introduce `AtomicMutex` that progressively spins, yields and sleeps to remove this, while also working with regular `unique_lock` patterns.
- The goal is to remove `CLockFreeGuard` entirely in subsequent PRs and replace it with standard patterns of `std::unique_lock` coupled with `AtomicMutex`

### DIFF
--- a/src/sync.h
+++ b/src/sync.h
@@ -380,6 +380,11 @@ public:
     }
 
     bool try_lock() noexcept {
+        // We locked it if and only if it was a false -> true transition.
+        // Otherwise, we just re-wrote an already existing value as true which is harmless 
+        // We could theoritically use CAS here to prevent the additional write, but
+        // but requires loop on weak, or using strong. Simpler to just use an exchange for
+        // for now, since all ops are seq_cst anyway.
         return !flag.exchange(true, std::memory_order_seq_cst);
     }
 };
--- a/src/sync.h
+++ b/src/sync.h
@@ -338,7 +338,7 @@ private:
     int64_t yields;
 
 public:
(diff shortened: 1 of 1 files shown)


src/sync.h
```diff
@@ -331,17 +331,81 @@ struct SCOPED_LOCKABLE LockAssertion
...
+    void unlock() {
+        flag.store(false, std::memory_order_seq_cst);
+    }
+
+    bool try_lock() noexcept {
+        // We locked it if and only if it was a false -> true transition.
+        // Otherwise, we just re-wrote an already existing value as true which is harmless 
+        // We could theoritically use CAS here to prevent the additional write, but
+        // but requires loop on weak, or using strong. Simpler to just use an exchange for
+        // for now, since all ops are seq_cst anyway.
```
Jouzo, line 387:
```suggestion
        // We could theoritically use CAS here to prevent the additional write
        // but requires loop on weak, or using strong. Simpler to just use an exchange
        // for now, since all ops are seq_cst anyway.
```

Bushstar (approved):

PR #1810, opened 2023-03-08
Add sleep in batchvote rpc

prasannavl:
/kind fix

- Node can become too busy to broadcast during batch voting. This workarounds it by for now by adding an idle period to process messages in between.

- Wait time is just added as an escape hatch for now since it's meant for limited advanced uses cases and not for widespread adoption.

- Note: The node also needs to keep catching up on other things like validating incoming blocks, possibly other minor RPCs, etc in between in this time the sleep is added. 

Rationale: Ideal targets are institutions that vote on behalf of users, so likely 500 to 3000 votes at a time using the current single threaded wallet logic that's held by global locks.

- 1s: up to 500s (8m) to 3000s (50m) of additional wait time.
- 0.5s: up at 250s (4m) to 1500s (25m) of additional wait time.

Considering these nodes might have many addresses per wallet, using a large time factor of 30s - 5m to sign the TX.

- 30s:
  - 500: ~250m (50m = 20% overhead)
  - 3000: ~1500m (50m = 3% overhead)
- 5m:
  - 500: ~2500m (2% overhead)
  - 3000: ~15000m (0.3% overhead)

Overhead is high only on low counts, where the consumers are better off using single vote. 

Edit: Now using 0.5s as default, since `-walletfastselect` has much better observed perf. characteristics. This should allows us to turn it as a default and just use a 0.5s for safety for now.


src/masternodes/rpc_proposals.cpp
```diff
@@ -739,6 +739,11 @@ UniValue votegovbatch(const JSONRPCRequest &request) {
         execTestTx(CTransaction(rawTx), targetHeight, optAuthTx);
 
         ret.push_back(signsend(rawTx, pwallet, optAuthTx)->GetHash().GetHex());
+        // Sleep the RPC worker thread a bit, so that the node can 
+        // relay the TXs as it works through. Otherwise, the main 
+        // chain can be locked for too long that prevent broadcasting of
+        // TXs
+        std::this_thread::sleep_for(std::chrono::seconds(1));
```
Bushstar, line 746: 1 second seems like a long time. Have we tested smaller time periods?

prasannavl: - TL;DR: 20% overhead if your overall time is ~4h / (~<1h overhead, where count is still under a few hours) 
- But overhead becomes in-significant and exponentially lesser as the number or the time spent on signing grows. 
- Doing a corollary: if signing becomes much faster (with `-walletfastselect`, etc) `votegov` on a loop is still a way to get the job done faster. One or the other could cater to these.

prasannavl: Switched to 0.5s in light of the perf. observations data with `-walletfastselect`

Bushstar (approved):

PR #1815, opened 2023-03-13
Add EVM Engine and Rust libs

Mixa84:
## Summary

- Adding SputnikVM code
- Adapting build process
- Adding Rust tools as prerequisite for building  

## Implications

- Index changes
  - [ ] Reindex required
  - [ ] Reindex optional
  - [x] Reindex not required

- Consensus


src/key_io.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include <assert.h>
 #include <string.h>
+#include <logging.h>
```
dcorral, line 12:
```suggestion
```

src/rust/crates/ain-evm-state/src/handler.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn call_evm(
+        &self,
+        caller: H160,
+        address: H160,
+        value: U256,
+        data: Vec<u8>,
+        _gas_limit: u64,
+        access_list: Vec<(H160, Vec<H256>)>,
+    ) -> (ExitReason, Vec<u8>) {
+        // TODO Add actual gas, chain_id, block_number
```
canonbrother, line 38: FYI `chain_id` (if needed)
https://chainlist.org/?search=defichain&testnets=true

PR #1818, opened 2023-03-14
net: rate limit the processing of incoming addr messages

Bushstar:
Adds the following Bitcoin back port from Dogecoin to our codebase to apply a rate limiter to the `addr` network P2P message.

https://github.com/dogecoin/dogecoin/pull/3008

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -83,14 +83,8 @@ static_assert(INBOUND_PEER_TX_DELAY >= MAX_GETDATA_RANDOM_DELAY,
 /** Limit to avoid sending big packets. Not used in processing incoming GETDATA for compatibility */
 static const unsigned int MAX_GETDATA_SZ = 1000;
 
-/** The maximum rate of address records we're willing to process on average. Can be bypassed using
- *  the NetPermissionFlags::Addr permission. */
-static constexpr double MAX_ADDR_RATE_PER_SECOND{0.1};
-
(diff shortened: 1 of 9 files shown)


Mixa84 (dismissed):

prasannavl:
To finishing the backport integration: 
- Tests: Remove rate limit for regtests
- SPV: Reimplement token bucket for SPV to addr to add rate limit there as well.

src/net_processing.cpp
```diff
@@ -2274,7 +2274,7 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
         const uint64_t nCurrentTime = GetTimeMicros();
         if (pfrom->nAddrTokenBucket < MAX_ADDR_PROCESSING_TOKEN_BUCKET) {
             const uint64_t nTimeElapsed = std::max(nCurrentTime - pfrom->nAddrTokenTimestamp, uint64_t(0));
-            const double nIncrement = nTimeElapsed * MAX_ADDR_RATE_PER_SECOND / 1e6;
+            const double nIncrement = nTimeElapsed * (Params().NetworkIDString() == CBaseChainParams::REGTEST ? 10000 : MAX_ADDR_RATE_PER_SECOND) / 1e6;
```
prasannavl, line 2277: Instead, can we pull these as a flag and set it on init? So, regtest, just starts with a different default, and it's configurable as well. So, we can tweak it if there is ever a misuse without doing another binary upgrade.

PR #1819, opened 2023-03-14
Remove `Require` from governance attributes

Bushstar:
To improve performance `Require` logic was removed from some of the hotpaths earlier and also converted to lambas for lazy evaluation of the error code path.

However this has defeated the rationale for introduction in the first place making it verbose and noisy. So, to keep with one  convention through the code base, this PR continues the removal of Require. 

Area: Gov Var Attributes

### DIFF
--- a/src/masternodes/errors.h
+++ b/src/masternodes/errors.h
@@ -298,11 +298,11 @@ public:
     }
 
     static Res GovVarApplyInvalidFactor(const CAmount ratio) {
-        return Res::Err("Factor cannot be more than or equal to the lowest scheme rate of %d", GetDecimalString(ratio * CENT));
+        return Res::Err("Factor cannot be more than or equal to the lowest scheme rate of %s", GetDecimalString(ratio * CENT));
     }
 
     static Res GovVarApplyDFIPActive(const std::string &str) {
-        return Res::Err("Cannot set block period while %d is active", str);
+        return Res::Err("Cannot set block period while %s is active", str);
     }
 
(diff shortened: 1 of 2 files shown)


Jouzo (dismissed):

src/masternodes/errors.h
```diff
@@ -128,6 +128,206 @@ class DeFiErrors {
...
+    static Res GovVarApplyUnexpectedType() {
+        return Res::Err("Unexpected type");
+    }
+
+    static Res GovVarApplyInvalidPool(const uint32_t pool) {
+        return Res::Err("No such pool (%d)", pool);
+    }
+
+    static Res GovVarApplyInvalidFactor(const CAmount ratio) {
+        return Res::Err("Factor cannot be more than or equal to the lowest scheme rate of %d", GetDecimalString(ratio * CENT));
```
DocteurPing, line 301: What the variable is consider at a number with "%d" instead of "%s"

src/masternodes/errors.h
```diff
@@ -128,6 +128,206 @@ class DeFiErrors {
...
+    }
+
+    static Res GovVarApplyDFIPActive(const std::string &str) {
+        return Res::Err("Cannot set block period while %d is active", str);
```
DocteurPing, line 305: What the variable is consider at a number with "%d" instead of "%s"

prasannavl:
Logic LGTM. Let's fix @DocteurPing's catch and go ahead.

PR #1820, opened 2023-03-14
Adds coin select flags to defid with option to override from CLI

Jouzo:
## Summary

- Adds coin select flags to defid
- Adds `GetOptionalBoolArg` for when a flag should not have any default value. This is here used to override `defid` flag iff flag is also set via `defi-cli`

## Examples

```
defid -walletfastselect=1

defi-cli sendtoaddress <address> 1
```
- Will use value from defid

---

```
defid -walletfastselect=1

defi-cli -walletfastselect=0 sendtoaddress <address> 1
```
- Value will be overridden and set to 0 from defi-cli

## Flags

```
-walletfastselect - Faster coin select - Enables walletcoinoptskipsolvable and walletcoinopteagerselect. This ends up in faster selection but has the disadvantage of not being able to pick complex input scripts. Default: false

-walletcoinoptskipsolvable - Coin select option: Skips IsSolvable signable UTXO check. Default: false

-walletcoinopteagerselect - Coin select option: Take fast path and eagerly exit on match even without having through the entire set. Default: false
```

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -181,7 +181,7 @@ DEFI_CORE_H = \
   masternodes/proposals.h \
   masternodes/tokens.h \
   masternodes/threadpool.h \
-  masternodes/coinselect.h \ 
+  masternodes/coinselect.h \
   masternodes/undo.h \
   masternodes/undos.h \
   masternodes/validation.h \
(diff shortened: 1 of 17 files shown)


prasannavl:
`CreateDefault` will still always initialize everything. We'll need to tweak this a bit.

prasannavl:
Added changes here: https://github.com/DeFiCh/ain/compare/pvl/coinselect-cli?expand=1. Believe that should provide a better base and mostly solve it. 

To-do on that branch:
- Find the best way to detect defi-cli or defid (best to use it statically if we have an optional `define` from somewhere)
- Port the request metadata pass on to the rest of methods in this branch

DocteurPing (dismissed):

prasannavl:
LGTM

Bushstar (approved):

PR #1822, opened 2023-03-15
Replace more require usages

Bushstar:
Continue the replacement of Require which performs worse than conventional conditional statements.


prasannavl (approved):
Logic LGTM.

PR #1826, opened 2023-03-17
Move sleep-time from node argument to rpc argument

DocteurPing:
Previously added a sleep_time node parameter for votegovbatch. Moved it as an RPC parameter instead of node parameter.

### DIFF
--- a/src/masternodes/rpc_proposals.cpp
+++ b/src/masternodes/rpc_proposals.cpp
@@ -596,7 +596,8 @@ UniValue votegovbatch(const JSONRPCRequest &request) {
                              {"masternodeId", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED, "The masternode ID, operator or owner address"},
                              {"decision", RPCArg::Type::STR, RPCArg::Optional::OMITTED, "The vote decision (yes/no/neutral)"},
                      }},
-                 },
+                    {"sleepTime", RPCArg::Type::NUM, RPCArg::Optional::OMITTED, "Sets sleeping time for voteGovBatch, by default the value is set to 500ms"}
+            },
             RPCResult{"\"hash\"                  (string) The hex-encoded hash of broadcasted transaction\n"},
             RPCExamples{HelpExampleCli("votegovbatch", "{{proposalId, masternodeId, yes}...}") +
                         HelpExampleRpc("votegovbatch", "{{proposalId, masternodeId, yes}...}")},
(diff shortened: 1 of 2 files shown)


Bushstar (approved):

src/masternodes/rpc_proposals.cpp
```diff
@@ -596,7 +596,8 @@ UniValue votegovbatch(const JSONRPCRequest &request) {
                              {"masternodeId", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED, "The masternode ID, operator or owner address"},
                              {"decision", RPCArg::Type::STR, RPCArg::Optional::OMITTED, "The vote decision (yes/no/neutral)"},
                      }},
-                 },
+                    {"sleepTime", RPCArg::Type::NUM, RPCArg::Optional::OMITTED, "Sets sleeping time for voteGovBatch, by default the value is set to 500ms"}
```
shohamc1, line 599: Formatting looks off

src/masternodes/rpc_proposals.cpp
```diff
@@ -611,7 +612,7 @@ UniValue votegovbatch(const JSONRPCRequest &request) {
     RPCTypeCheck(request.params, {UniValue::VARR}, false);
 
     const auto &keys = request.params[0].get_array();
-    auto sleepTime = gArgs.GetBoolArg("-sleep-time", SLEEP_TIME_MILLIS);
+    auto sleepTime = request.params[1].isNull() ? SLEEP_TIME_MILLIS : request.params[1].get_int();
```
shohamc1, line 615: I think you need to check if the param is set first, otherwise the node may crash.

src/masternodes/rpc_proposals.cpp
Bushstar: Thought the same at a glance but look okay in the file imo.

https://github.com/DeFiCh/ain/blob/9fcb98c12c504cd3500ff8265a81ae5a22ebde1c/src/masternodes/rpc_proposals.cpp#L599

src/masternodes/rpc_proposals.cpp
Bushstar: .isNull() is the check.

src/masternodes/rpc_proposals.cpp
```diff
@@ -612,7 +613,7 @@ UniValue votegovbatch(const JSONRPCRequest &request) {
     RPCTypeCheck(request.params, {UniValue::VARR}, false);
 
     const auto &keys = request.params[0].get_array();
-    auto sleepTime = request.params[1].isNull() ? SLEEP_TIME_MILLIS : request.params[1].get_int();
+    auto sleepTime = (request.params.size() > 1 && request.params[1].isNull()) ? SLEEP_TIME_MILLIS : request.params[1].get_int();
```
Bushstar, line 616: You can leave this as it but isNull() is safe to call without a size() check as UniValue defines [] usage and makes it safe to use on non-existent entries. Check other isNull() usage in the code on optional RPC arguments.

Bushstar (approved):

PR #1835, opened 2023-03-21
Switch Tx ordering fallback from hash to entry time

kuegi:
## Summary

In fee based tx ordering, the fallback for ordering within the same fee level is based on the txId (hash). This means that positioning ones tx within the same fee level results in PoW calculations to get the "right" tx-hash.

This puts normal users who just send a tx at an disadvantage ("frontrunning" still possible within fee levels) and since we are a PoS chain and not a PoW chain, it would be better to change that.

## Implications

IMHO no implications on consensus etc.

- Index changes
  - [ ] Reindex required
  - [ ] Reindex optional
  - [x] Reindex not required

- Consensus


Bushstar (approved):

prasannavl:
LGTM on the logic. Thanks @kuegi. Tests still need fixing.

kuegi:
yes sorry, missed the tests. will fix.

kuegi:
🤔 jellyfish tests seem to go into a timeout, but no idea how this can be connected to the change.

prasannavl:
No worries, thanks for wrapping it up quickly! 

LGTM

PS. Tests - all green locally, Jellyfish worker VMs are running out memory, will need to upsize them for the time being.

prasannavl (approved):

kuegi:
@prasannavl just saw that you require signed commits. sorry didn't set it up before (did now). should I squash and recommit with sign?

kuegi:
@prasannavl with the testnet release approaching, it would be great if we can finally merge this in. So we not miss it for the next release.

kuegi:
@prasannavl @Bushstar since this is already approved and marked as "next release": can we merge it in?

prasannavl:
Thanks Kuegi. Apologies for the delay - this got missed among other things.

PR #1837, opened 2023-03-22
Add transferbalance and evmtx custom transactions.

Mixa84:
## Summary

- Adding two custom txs on Defichain side for evm
- Transfer Balance tx: it will be used for various balance transfers, among which evm-in and evm-out. Those will be used to transfer funds to and from ETH addresses.
- Evm tx: will hold raw evm tx which will be passed to evm to process

## RPCs

- transferbalance: <type> {"from_address":"amount@token"} {"to_address":"amount@token"}
- evmtx: <raw_evm_tx_bytes>

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -141,6 +141,10 @@ std::string ToString(CustomTxType type) {
             return "Vote";
         case CustomTxType::UnsetGovVariable:
             return "UnsetGovVariable";
+        case CustomTxType::TransferBalance:
+            return "TransferBalance";
+        case CustomTxType::EvmTx:
+            return "EvmTx";
         case CustomTxType::None:
             return "None";
     }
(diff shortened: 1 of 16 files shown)


src/masternodes/mn_checks.cpp
```diff
@@ -3788,6 +3800,64 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+    Res operator()(const CTransferBalanceMessage &obj) const {
+        auto res = Res::Ok();
+
+        // owner auth
+        if (obj.type != CTransferBalanceType::EvmOut)
+            for (const auto &kv : obj.from) {
+                res = HasAuth(kv.first);
+                if (!res)
+                    return res;
+            }
```
kuegi, line 3812: does EvmOut mean "sending from defichain to EVM" or "from EVM to defichain"?
cause this check implies that EvmOut means "from defichain to EVM"

but at line 3834  it looks like "EvmIn" is considered as "defichain to EVM". 
Maybe rename it DFCtoEVM and EVMtoDFC to prevent such confusion?

src/masternodes/mn_checks.cpp
```diff
@@ -3788,6 +3800,64 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
         return mnview.AddProposalVote(obj.propId, obj.masternodeId, vote);
     }
 
+    Res operator()(const CTransferBalanceMessage &obj) const {
+        auto res = Res::Ok();
+
+        // owner auth
+        if (obj.type != CTransferBalanceType::EvmOut)
```
kuegi, line 3807: why don't we need to check auth for EvmOut too? Is that check happening on EVM side? Without the check, a transferbalance EvmOut might currently just create DFI out of thin air, right?

PR #1850, opened 2023-03-31
Add auth for EvmOut transfer balance

Mixa84:
## Summary

- Add auth of DFI equivalent address of ETH from address in the tx
- Check in validation that DFI auth corresponds to ETH address in from


Bushstar (approved):

src/masternodes/mn_checks.cpp
```diff
@@ -3844,6 +3844,30 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+                        auto it = input.scriptSig.begin();
+                        CPubKey pubkey(input.scriptSig.begin() + *it + 2, input.scriptSig.end());
+                        auto script = GetScriptForDestination(WitnessV16EthHash(pubkey));
+                        if (script == addr)
+                            foundAuth = true;
+                    }
+                }
+                if (!foundAuth)
+                    return Res::Err("authorization not found for %s in the tx", ScriptToString(addr));
+            }
```
kuegi, line 3869: Why not do this at line 3813 where the Auth for the other cases is checked?

kuegi: Why would you loop two times? you check Auth and Eth-type in the "Auth section" so you do not have to check here.

Would be clearer code-wise:
* check valid auth and address type
* check values and sums
* if all good: do actual transfer.

Mixa84: It doesn't matter where it is done. I wanted to do it in the part for EVMOut if section, but because other auths are the same I didn't do them in their own section but in the beginning.

Jouzo (approved):

PR #1857, opened 2023-04-04
Update toolchains, deps and further cleanup

prasannavl:
## Summary

- This continues https://github.com/DeFiCh/ain/pull/1855
- Removes flags on CI that aren't used when DOCKERFILE is specified. Only `TARGET` is passed into the containers.
- Switch ubuntu to `latest` in all cross compilations, gcc builds and CI.
- Switch clang build to debian 10 to support existing older OSes all the way to Ubuntu 18.04 LTS
  - Use Ubuntu to stay on tip. Cross compilation of Windows and Mac is perfect as it relies on it's own toolchain
  - Use Debian for long term support
- Use multiplatform docker images as base for armv7 and arm64 docker images. 
- Propagate the clang toolchain to depends builds as well
- Enable static linking of libgcc
- Merge `share` and `contrib`. Ideally would have made more semantic sense to just merge everything into share and and let it just be a part of the distribution. However, the ones in contrib has a larger base, and is used all over the place and requires significant changes. So, merging it under contrib. In the future, could potentially rename it to share or keep it as it is, and cherry pick what's needed into the distribution with make config, as it currently does.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


Jouzo (approved):
Built and tested on macOS 12.5.1

PR #1859, opened 2023-04-04
Add EVM block header to coinbase

Jouzo:
## Summary

- Add TransactionQueue and TransactionQueueMap to be used with context
- Add finalize_block and related FFI sending back the rlp encoded header 
-  Add EVM header to coinbase


src/defid.cpp
```diff
@@ -68,7 +68,7 @@ static bool AppInit(int argc, char* argv[])
     util::ThreadRename("init");
     init_runtime();
     // Test linking to dummy address
-    evm_add_balance("0xf3088943Fa15Ff33D5C3Af9845dc65073Ae2bAc8", 100);
+    evm_add_balance("0xf3088943Fa15Ff33D5C3Af9845dc65073Ae2bAc8", ArithToUint256(100).ToArrayReversed());
```
Jouzo, line 71:
```suggestion
```

src/defid.cpp
```diff
@@ -68,7 +68,6 @@ static bool AppInit(int argc, char* argv[])
     util::ThreadRename("init");
     init_runtime();
     // Test linking to dummy address
```
Jouzo, line 70:
```suggestion
```

PR #1866, opened 2023-04-06
chore(workflow): evm ci

canonbrother:
## Summary

For keeping the evm side in a release-steady state


Jouzo (commented):
Let's add clippy as well

.github/workflows/tests-evm.yml
```diff
@@ -0,0 +1,74 @@
+name: Tests - EVM
+
+on:
+  push:
+    paths: [ src/rust/** ]
```
canonbrother, line 5: triggered by file changes

PR #1879, opened 2023-04-09
Revamp of build, CI tests and packaging - 2

prasannavl:
## Summary

**Status: Closing in favor of https://github.com/DeFiCh/ain/pull/1880**

- Replacement for https://github.com/DeFiCh/ain/pull/1877

## Implications

- Storage
  - [ ] Database reindex required
  - [ ] Database reindex optional
  - [ ] Database reindex not required
  - [x] None

- Consensus
  - [ ] Network upgrade required
  - [ ] Includes backward compatible changes
  - [ ] Includes consensus workarounds
  - [ ] Includes consensus refactors
  - [x] None

PR #1880, opened 2023-04-10
Overhaul of build, CI tests and lints, packaging and deps

prasannavl:
## Summary

*This is a cleaned up split version of https://github.com/DeFiCh/ain/pull/1877 and https://github.com/DeFiCh/ain/pull/1879*

This revamps the following:

- Build process
- Tests
- Lints
- Packaging
- Others

---

### Build

> `./make.sh build`

- New out-of-tree build:

```
...
- build
  - aarch64-linux-gnu
     - bin
     - src
     - config.guess
     - config.sub
     - Makefile
     - ...
  - arm-linux-gnueabihf
  - depends
  - x86_64-apple-darwin
  - x86_64-pc-linux-gnu
  - x86_64-w64-mingw32
- src
- make.sh
...
```

- All of the builds are contained inside the `build` dir. 
- In addition to the full build, `bin` and `share` dirs are deployed into the root of `build/<host>` dir for convenience.
- Docker releases continue to package it in `./build/defichain-<version>` dir. 
- Removing the build is as simple as removing the `./build` dir or `./make.sh purge`.
- `TARGET=<host> /make.sh clean` can individually clean targets 
- The autoconf files (the only files that are generated in-tree) can be removed by `./make.sh clean-conf`
- Numerous fixes and overhaul in `./make.sh`. See `doc/build-quick.md` or read file for more info. 
  - Adds built-in support for `greadlink` on osx

#### Context

- Default Bitcoin build recommendations are in-tree and are highly sub-optimal. They pollute the source tree significantly and also expect many hard coded in-tree behaviours and modifications. In addition, also throws of many systems due to the constantly changing files, despite gitignore is a hint to ignore them. Some dev tools get easily confused and results in ping pong of rebuilds causing a sluggish dev experience.
- This single in-tree build also makes it extremely difficult to comprehend the changes as well tests and do cross compilations without a full rebuild.
- This revamps the build process to ensure that builds are out of source tree. 
- Depends system is hardcoded to be in place by bitcoin. This fixes that by adding a `DESTDIR` to depends and update depends scripts and packages accordingly to be agnostic.
- Adapts and simplifies dockerfiles to remove unused multi-stages for easier maintenance

---
### Tests

- Test logs now use the build folder and doesn't pollute the source tree. 
- Removes unstable functions from base framework
- Tests by default expect binaries to be built in-tree. Add logic to detect defaults for out of source tree config 
- `./make.sh` now has support target aware tests, and tests are contained in `<build-dir>/test-runner` instead of system `/tmp`
- Adds CI grouping markers across several places so that CI logs are comprehensible for humans

### Lints

- Re-enables multiple disabled lints with adaptations
- Update linter deps

#### CI

- Adapts the CI tests to not build out of tree and makes it runnable. Add logic to detect default for out of source tree config
- Simplifies CI tests by removing orthogonal methods that are hard to maintain and simply reuses `./make.sh`
- Simplifies lints in the same manner.

---
### Packaging and Others

- Adds LICENSE and README.md into `share/defi` dir for packaging
  - README.md has been moved from the root location to here to stay in line with Unix FHS
- Removes `test_defi` and `bench_defi` from the distribution (They are still built as usual)
- `./make.sh` key functions now emit GitHub CI group to categorise logs
- Fixes ARM64 builds due to missing CRC intrinsics 
- Resolve compiler warnings

---
## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


prasannavl:
Irrelevant linux build failure due to docker login. Believe CI workflows need to be in master to retrieve secrets.

PR #1882, opened 2023-04-11
evm: Connect and store blocks

shohamc1:
## Summary

- Store block data in `finalize_block`
- Add `eth_blockNumber` and `eth_getBlockByCount` RPCs

### DIFF
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -83,6 +83,7 @@ dependencies = [
  "cxx-gen",
  "env_logger",
  "ethereum",
+ "fake",
  "heck",
  "jsonrpsee-core",
  "jsonrpsee-http-client",
@@ -96,6 +97,7 @@ dependencies = [
  "prost",
  "prost-build",
  "quote",
+ "rand",
  "regex",
  "serde",
  "serde_json",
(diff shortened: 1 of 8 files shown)


src/rust/crates/ain-evm/src/handler.rs
```diff
@@ -13,4 +20,85 @@ impl Handlers {
...
+
+        let mut parent_hash: H256 = Default::default();
+        let mut number: usize = 0;
+
+        let blocks = self.block.blocks.read().unwrap();
+        if !blocks.is_empty() {
+            parent_hash = blocks.first().unwrap().header.hash();
+            number = blocks.len() + 1;
+        }
+        drop(blocks);
```
Jouzo, line 60:
```suggestion
        let (parent_hash, number) = {
            let blocks = self.block.blocks.read().unwrap();
            blocks
                .first()
                .and_then(|first_block| Some((first_block.header.hash(), blocks.len() + 1)))
                .unwrap_or((H256::default(), 0))
        };
```
What about something like this so we remove the need to drop explicititly and the mut variables

canonbrother:
heyy @shohamc1 one more request yea.. can you add unite test for the rpc as well?
[src/rust/crates/ain-grpc/src/tests.rs](https://github.com/DeFiCh/ain/blob/a17212215f6b645e37c07a3567b367fb0688c523/src/rust/crates/ain-grpc/src/tests.rs)
THANKS!!

src/rust/crates/ain-evm/src/handler.rs
shohamc1: This looks excellent, will do this instead. Was unaware of this syntax in Rust.

shohamc1: Fixed: https://github.com/DeFiCh/ain/pull/1882/commits/922dca0ccda9a4d6a384bda1f0ccc07fea88d777

PR #1884, opened 2023-04-11
[WIP] EVM C++ FFI

Mixa84:
## Summary

- Adding C++ functions into Rust necessary for some ETH rpcs to work


src/rust/crates/ain-evm-cpp-ffi/build.rs
```diff
@@ -0,0 +1,20 @@
+fn main() {
+    cxx_build::bridge("src/lib.rs")
+        .define("CLIENT_VERSION_MAJOR", Some("3"))
```
prasannavl, line 3: Just include the generated config.h file instead?

shohamc1 (approved):
FFI calls work 🙂

PR #1889, opened 2023-04-12
Feature/evm proto change

DocteurPing:
This PR add the possibility for the message EthGetBlockByHashResult to be an object or a string


shohamc1 (changes requested):
`eth_getBlockByHash` always returns a `EthBlockInfo` struct. 

The change should be in the transactions field in `EthBlockInfo`, which can be an array of TX hashes or an array of transaction objects depending on the second boolean param.

prasannavl:
Conflict resolution needed.

PR #1900, opened 2023-04-13
evm: Save failed transactions in block

shohamc1:
## Summary

- Save failed transactions in EVM block as per Ethereum specification


Jouzo (approved):

prasannavl:
Conflict resolution needed. May be add the gen to `.git/info/exclude` for now

PR #1912, opened 2023-04-14
evm: Enable C++ FFI

shohamc1:


lib/ain-cpp-imports/build.rs
```diff
@@ -0,0 +1,27 @@
...
+        .unwrap()
+        .parent()
+        .unwrap()
+        .join("src");
+
+    cxx_build::bridge("src/lib.rs")
+        .include(lib_path)
+        .flag_if_supported("-std=c++17")
+        .flag_if_supported("-Wno-unused-parameter")
+        .compile("ain-evm-cpp-ffi");
```
prasannavl, line 17: This one

prasannavl (commented):
Then Makefile doesn't have the correct renames. So, assume this is only building from previous cache or so.

PR #1916, opened 2023-04-17
Add Eth_Accounts RPC

DocteurPing:
This pull request add the RPC Eth_accounts for the EVM implementation.


prasannavl:
We might want to filter out and send only the ETH addresses here though and not include the other. 

cc/ do get a review by @Bushstar once the logic is sorted

Bushstar (approved):

PR #1926, opened 2023-04-20
Implement evm RPC into test framework

sieniven:
## Summary

- Modification of test framework to implement evm RPC
- Handling of EVM RPCs to the correct jsonrpc port on defid node

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


sieniven:
Commit #2ad8aca redirects the output of rust lib's env_logger to stdout instead of the default stderr.

sieniven:
@Jouzo for sure. How should i showcase this? Currently the implementation is such that the TestNode class creates an instance of the HTTPConnection the the evm jrpc port, and stores a map of eth specific RPCs. When making RPCs to the node, we map to see if the call is eth-specific, and then pass it to the correct HTTPConnection instance accordingly.

lib/ain-grpc/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub fn add_grpc_server(_runtime: &Runtime, _addr: &str) -> Result<(), Box<dyn Er
 
 pub fn init_runtime() {
     log::info!("Starting gRPC and JSON RPC servers");
-    LogBuilder::from_env(Env::default().default_filter_or(Level::Info.as_str())).init();
+    LogBuilder::from_env(Env::default().default_filter_or(Level::Info.as_str())).target(Target::Stdout).init();
```
prasannavl, line 55:
```suggestion
    LogBuilder::from_env(Env::default().default_filter_or(Level::Info.as_str()))
        .target(Target::Stdout)
        .init();
```

make.sh
```diff
@@ -348,8 +348,10 @@ test_py() {
 
     _ensure_enter_dir "${release_target_dir}"
 
+    _fold_start "functional-tests"
```
prasannavl, line 351:
```suggestion
```

make.sh
```diff
@@ -348,8 +348,10 @@ test_py() {
...
     # shellcheck disable=SC2086
     ./test/functional/test_runner.py --tmpdirprefix "./test_runner/" --ansi "$@"
+    _fold_end
```
prasannavl, line 354:
```suggestion
```

Issue #1678, opened 2022-12-28
`createmasternode` with custom reward address


mrgrauel:
#### What would you like to be added:
* add an option to `createmasternode` to define a custom reward address

#### Why is this needed:
* the same function is possible with `updatemasternode`
* a custom address should be definable on creation

charlestonxorg:
Lets do it

prasannavl:
Resolved since v3

PR #1941, opened 2023-04-27
Add custom reward address to `createmasternode` RPC 

DocteurPing:
This PR was created to fix this [issue](https://github.com/DeFiCh/ain/issues/1678). It will allow the user to create a masternode while specifying a custom reward address.

## RPCs

- Createmasternode RPC was changed. A new parameter was added to the end of the list in order to specify the reward address. This parameter is optional. and if not specified it will be the address of the owner.


src/masternodes/mn_checks.h
```diff
@@ -249,6 +251,8 @@ struct CCreateMasterNodeMessage {
         if (!s.eof()) {
             READWRITE(timelock);
         }
+        READWRITE(rewardType);
```
shohamc1, line 254: Would need a fork check. Not sure if we can extend this message since it has already been extended once before.

shohamc1 (changes requested):

src/masternodes/mn_checks.h
Bushstar: Could do like last time I extended this custom TX message.

```
        if (!s.eof()) {
            READWRITE(rewardType);
        }
```

src/masternodes/mn_checks.cpp
```diff
@@ -820,6 +820,8 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
         node.creationHeight      = height;
         node.operatorType        = obj.operatorType;
         node.operatorAuthAddress = obj.operatorAuthAddress;
+        node.rewardAddressType   = obj.rewardType;
```
Bushstar, line 823: Needs fork guard.

DocteurPing: i've set the height to the NextNetworkUpdate since we don't have it set yet.

src/masternodes/mn_checks.cpp
```diff
@@ -820,6 +820,10 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
         node.creationHeight      = height;
         node.operatorType        = obj.operatorType;
         node.operatorAuthAddress = obj.operatorAuthAddress;
+        if (height >= static_cast<uint32_t>(consensus.NextNetworkUpgradeHeight)) {
+            node.rewardAddressType   = obj.rewardType;
```
Bushstar, line 824: In CMasternodesView::CreateMasternode add checks that the type is either NoDestType, PKHashType, ScriptHashType or WitV0KeyHashType, so 0, 1, 2 or 4.

src/masternodes/rpc_masternodes.cpp
```diff
@@ -165,7 +174,12 @@ UniValue createmasternode(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Address (%s) is not owned by the wallet", EncodeDestination(ownerDest)));
     }
 
+    if (rewardDest.index() != 1 && rewardDest.index() != 4) {
```
Bushstar, line 177: Remove this check and let it happen on the consensus side. Reward will include type 2 as well on the next hard fork due to PR 1664.

src/masternodes/rpc_masternodes.cpp
```diff
@@ -133,15 +134,23 @@ UniValue createmasternode(const JSONRPCRequest& request)
 
     std::string ownerAddress = request.params[0].getValStr();
     std::string operatorAddress = request.params.size() > 1 && !request.params[1].getValStr().empty() ? request.params[1].getValStr() : ownerAddress;
+    std::string rewardAddress = !request.params[4].getValStr().empty() ? request.params[4].getValStr() : ownerAddress;
```
Bushstar, line 137: Leave empty if no reward is set, so just get an empty string here, dest will be 0 which is the toggle as to whether to use the reward address or not.

blocs0 (approved):

prasannavl:
Closing stale PRs.

Issue #427, opened 2021-05-25
I suggest caching of GetMasternodeLastBlockTime


muirglacier:
#### What would you like to be added:

GetMasternodeLastBlockTime() in the worst case loops through all master nodes, with each causing a DB lookup.
I suggest caching the lastBlockTime of each nodeId (and updating the cache in the "increment/decrement minted blocks" function)

#### Why is this needed:

When you have many master nodes running on one client, things get really slow. Each second, for each local masternode, big chunks of the DB are queried in the for-loop in GetMasternodeLastBlockTime() in order to compute the stakerBlockTime.
This not only causes a high CPU load but the last local master node in your list has a time disadvantage due to the delay in all stakerBlockTimes that had to be previously computed.

Also, the listmasternodes RPC command becomes a lot quicker. I have added the cache in my local defid and listmasternodes went from around 5 seconds down to 0.48 seconds.

uzyn:
Thanks for this @muirglacier 

> I have added the cache in my local defid and listmasternodes went from around 5 seconds down to 0.48 seconds.

Would you like to submit a PR for this improvement?

PR #1946, opened 2023-04-28
CI build workflow cleanup

sieniven:
## Summary

This PR cleans up the current CI workflow. Note that the CI workflow was initially tested with build-dev. Efforts to push multi-arch images is discontinued in order prioritize consolidation to 1 dockerfile. Refer to Build-Dev run #427 for full build results for all supported platform architecture types, as the eventual build-dev will only contain the 3 architecture: linux-x64, win-x64, and osx-x64.

- Ports all target architecture specific dependencies and build steps into make.sh.
- Standardization of build into a single, target platform architecture agnostic dockerfile (defi.dockerfile). The host target is specified as a Docker build argument instead when compiling the binaries.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


sieniven:
Removing docker buildx pipeline as manifest list of defi image is created manually. Deployment of binaries will be shifted into the same build dockerfile (defi.dockerfile), and the target platform of the image is passed as build argument.

sieniven:
Discontinuing efforts to push multi-arch images using docker buildx into a single docker image tag on Dockerhub registry, as that will require segregating deployment image from build image. Our current pipeline implements cross-compilation of defi application due to the caveat of compiling for osx and windows platform, thus porting over to buildx will require additional complexity when it comes to support for having 2 separate build pipelines which makes it less than ideal. Thus only linux-x64 image will be pushed to the registry.

PR #1949, opened 2023-05-02
Fix functional test config paths

sieniven:
## Summary

- This PR fixes the fucntional test configuration file's relative path when running the various functional tests using test_runner.py.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -88,6 +88,9 @@ def get_default_config_path():
         # aarch64 / arm builds
         current_file_path + "/../../../build/arm-linux-gnueabihf/test/config.ini",
         current_file_path + "/../../../build/aarch64-linux-gnu/test/config.ini",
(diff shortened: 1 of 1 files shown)


sieniven:
Summary on this PR rationale:
1. the default functional test configuration file will be using ./build (if there is a configuration in there), since it is likely that the developer will choose this to be prioritize over host specific build (in the scenario where the developer may have multiple builds). This is also the usual standard way to build C++ projects inside a ./build directory.
2. The next default config path will be using in-tree path, since this likely meant that the developer explicitly built the project in tree.
3. ./build/<host> configurations will follow, with the default priority host architectures selected are Linux-x64, OSX-x64, OSX-arm64 and win-x64 for standard developement environments. This build is the more specific default when using make.sh to build, so it is easier for developers to build all platforms.

Bushstar (approved):

PR #1955, opened 2023-05-02
Convert uint256 to array or reversed array

Bushstar:
Adds the ability to convert uint256 to an array or a reversed array.


src/uint256.h
```diff
@@ -52,6 +53,20 @@ class base_blob
     void SetHex(const std::string& str);
     std::string ToString() const;
 
+    [[nodiscard]] std::array<uint8_t, WIDTH> ToArray() const
+    {
+        std::array<uint8_t, WIDTH> reversedArray;
+        std::copy(data, data + sizeof(data), reversedArray.begin());
+        return reversedArray;
```
Mixa84, line 60:
```suggestion
        std::array<uint8_t, WIDTH> array;
        std::copy(data, data + sizeof(data), array.begin());
        return array;
```

Mixa84 (dismissed):

prasannavl (approved):

PR #1957, opened 2023-05-02
Pass miner address to EVM side

Bushstar:
Passes the miner address to the EVM side to allow fees on the EVM side to be paid to the block miner.


src/validation.cpp
```diff
@@ -2828,6 +2828,50 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
     }
     mnview.SetLastHeight(pindex->nHeight);
 
+    const CDataStructureV0 enabledKey{AttributeTypes::Param, ParamIDs::Feature, DFIPKeys::EVMEnabled};
+
+    if (pindex->nHeight >= chainparams.GetConsensus().NextNetworkUpgradeHeight && attributes->GetValue(enabledKey, false)) {
```
prasannavl, line 2833: How about abstracting this into a separate function along with `IsEVMEnabled`

prasannavl:
Ignore comments for now to keep diff down. Leaving them to be resolved post-merge.

PR #1958, opened 2023-05-02
EVM Custom TX types

Bushstar:
Adds custom transaction types to allow interacting with EVM.


prasannavl:
Post-merge comments (Ignore for now to keep diff down): 

- `TransferBalance` -> `TransferDomain`
- `EvmTx` -> Abstract as `xVMTx` with EVM subtype.

PR #1960, opened 2023-05-02
Refine rust build to use local rust toolchain

sieniven:
## Summary

- This PR refines the rust build pipeline to enable the use of local rust toolchain during build stage.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


configure.ac
```diff
@@ -1316,6 +1316,14 @@ if test x$build_defi_wallet$build_defi_cli$build_defi_tx$build_defid$use_bench$u
   AC_MSG_ERROR([No targets! Please specify at least one of: --with-utils --with-libs --with-daemon --enable-bench or --enable-tests])
 fi
 
+AC_CHECK_PROG([local_rust_toolchain],cargo,yes)
```
prasannavl, line 1319: I'm not sure if we need this config, rather than use CARGO_HOME and RUSTUP_HOME to implicitly indicate it and work intuitively than an extra config

lib/Makefile.am
```diff
@@ -1,8 +1,8 @@
-# TODO: Refine these vars with graceful swichover to local and
-# ensure it works without them set as well as with them set
-CARGO_HOME ?= 
-RUSTUP_HOME ?= 
-CARGO ?= cargo
+if LOCAL_RUST_TOOLCHAIN
```
prasannavl, line 1: Don't believe this is valid syntax

configure.ac
sieniven: Removed config and check.

lib/Makefile.am
sieniven: Resolved with commit b38b4be.

sieniven:
Closing this PR as build changes to source local rust toolchain is already inside feature/evm.

PR #1967, opened 2023-05-04
Fix eth json RPC issue when running DFI in containerized environment

sieniven:
## Summary

This PR fixes the IP address binded to the eth json rpc server and grpc server from the hardcoded loopback interface (127.0.0.1) to read args "-ethrpcbind" and "-grpcbind".

1. Refactored init_runtime into init and init_evm_runtime.
- init function call will pass argc and argv from the C++ side, and convert the arg variables into Rust's &str.
- init_evm_runtime will initialize the Rust logger and runtime.
- fixed logging in init_evm_runtime.
2. Include args to bind eth_rpc server and grpc server to addresses. (-ethrpcbind, grpcbind)
- by default if address to bind is not specified or -rpcallowip arg is not set, we will use the standard lo interfafce (127.0.0.1)

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-grpc/src/lib.rs
+++ b/lib/ain-grpc/src/lib.rs
@@ -18,9 +18,7 @@ use log::Level;
 use crate::rpc::{MetachainRPCModule, MetachainRPCServer};
 
 use std::error::Error;
-use std::ffi::CStr;
 use std::net::SocketAddr;
-use std::os::raw::c_char;
 use std::sync::Arc;
 
 use ain_evm::runtime::{Runtime, RUNTIME};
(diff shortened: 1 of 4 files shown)


sieniven:
1. Refactored init_runtime into init and init_evm_runtime:
- init function call will pass argc and argv from the C++ side, and convert the arg variables into Rust's &str.
- init_evm_runtime will initialize the Rust logger and runtime.

2. Include args to bind eth_rpc server and grpc server to addresses. (-ethrpcbind, grpcbind)
- by default if address to bind is not specified or -rpcallowip arg is not set, we will use the standard lo interfafce (127.0.0.1)

PR #1971, opened 2023-05-05
DVM: Enable DUSD loops in vaults 

kuegi:
## Summary

This pull request applies the proposed code change from the approved [special DFIP](https://defiscan.live/governance/97f7dc8f631d65deda6867f2b2bde5f20860559323f0918f11604c9cf2f05bfe/). 

The goal is to allow 100% DUSD looped vaults again. Important: only 100% DUSD collateral gets a special case to prevent exploits.

### Effective change
If feature flag is activated, all loans are possible when the collateral is 100% DUSD. 
without the flag/current behaviour: only dToken-loans are possible in this case.

It is guarded via a feature flag to allow for activation/deactivation without needed consensus-update in the future.

## Implications

- Storage
  - [X] None

- Consensus
  - [ ] None


kuegi:
@prasannavl I would appreciate if someone from the team finds the time to quickly review this (its really not much code). Since the according DFIP was approved with over 90%, we should add that to the next release IMHO.

src/masternodes/mn_checks.cpp
```diff
@@ -3003,6 +3010,18 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
         auto isPostFCR = static_cast<int>(height) >= consensus.FortCanningRoadHeight;
         auto isPostGC  = static_cast<int>(height) >= consensus.GrandCentralHeight;
+        auto isPostNext =  static_cast<int>(height) >= consensus.NextNetworkUpgradeHeight;
+
+        if(isPostNext) {
+            const CDataStructureV0 enabledKey{AttributeTypes::Param, ParamIDs::Feature, DFIPKeys::AllowDUSDLoops};
+            auto attributes = mnview.GetAttributes();
+            assert(attributes);
+            auto DUSDLoopsAllowed= attributes->GetValue(enabledKey, false);
+            if(DUSDLoopsAllowed && hasDUSDColl && !hasOtherColl) {
```
Bushstar, line 3020: This should check that the loans are 100% DUSD. There is a hasDUSDLoans bool passed to CollateralPctCheck which can check if there is a DUSD loan but a hashOtherLoans or something similar would also need to be added.

Bushstar (approved):

PR #1973, opened 2023-05-07
Add EVM state trie

Jouzo:
## Summary

- Add EVM state trie
- Persist state to RocksDB via VSDB dependency
- Introduce BridgeTX EvmIn/EvmOut on EVM side. 
- Add native `evmrawtx` RPC call
- Test creation, deployment, call and storage of smart contract and its state at different block heights
- Test account state at different block heights
- Code cleanup

## Missing

- Return failed EvmIn/EvmOut as TX
- Define Vicinity.
- Implement Vicinity related Backend trait methods
- Keep executor's backend private and introduce new (Commitable/Persistent)Backend trait

## Implications

- Storage
  - [ ] None

- Consensus
  - [x] Devnet rollback required


prasannavl:
Note: Before we merge into master, let's make sure to fork it into defich and switch to that. Merging it for now to prevent conflicting work

Issue #1852, opened 2023-04-03
ETH-JSON RPC Spec Implementation


canonbrother:
## Summary

- Issue dedicated to track and document ETH-JSON RPC in ain.
  - Reference:
    - https://ethereum.org/en/developers/docs/apis/json-rpc/
    - https://ethereum.github.io/execution-apis/api-documentation/

#### Status 
- The major RPCs for MetaMask usage and contract deployments are in place.
- Priority based on eco-system usage. 
- All RPCs that are a part of ETH-JSON spec are in scope to be implemented. 
- What's not in scope: Additional geth, and debug RPCs that aren't a part of the spec.

### `block`

- eth_getBlockByHash 
    - [x] impl @shohamc1 #1864
    - [x] unit test
- eth_getBlockByNumber
    - basic
        - [x] impl @shohamc1 #1882
        - [ ] unit test
    - full
        - [ ] impl
        - [ ] unit test
- eth_getBlockTransactionCountByHash
    - [x] impl
    - [x] unit test
- eth_getBlockTransactionCountByNumber
    - [x] impl @dcorral #1905
    - [ ] unit test
- ~~eth_getUncleCountByBlockHash~~
- ~~eth_getUncleCountByBlockNumber~~

### `client`
- eth_chainId
    - [x] impl @shohamc1 #1891
    - [x] unit test
- ~~eth_syncing~~
- ~~eth_coinbase~~
- eth_accounts
    - [x] impl @DocteurPing #1916
    - [x] unit test 
- eth_blockNumber 
    - [x] impl @shohamc1 #1882
    - [ ] unit test 
- net_version 
    - [x] impl @shohamc1 #1891
    - [x] unit test    

### `execute`

- eth_call
    - [x] impl
    - [ ] unit test
- eth_estimateGas
    - basic
        - [x] impl @Mixa84 #1940
        - [x] unit test
    - full
        - [ ] impl
        - [ ] unit test
- ~~[ ] eth_createAccessList~~

### `fee market`
- eth_gasPrice
    - basic
        - [x] impl @Mixa84 #1940
        - [x] unit test
    - full
        - [ ] impl
        - [ ] unit test
- eth_maxPriorityFeePerGas
    - [ ] impl
    - [ ] unit test
- eth_feeHistory
    - [ ] impl
    - [ ] unit test

### `filter`

- [ ] eth_newFilter
- [ ] eth_newBlockFilter
- [ ] eth_newPendingTransactionFilter
[...]

PR #1984, opened 2023-05-15
feat(lib): add rpc `eth_sendTransaction`

canonbrother:
## Summary

- as per title

## RPCs

-  eth_sendTransaction

fixes part of #1852


lib/ain-grpc/src/transaction_request.rs
```diff
@@ -0,0 +1,99 @@
...
+    Legacy(LegacyTransactionMessage),
+    EIP2930(EIP2930TransactionMessage),
+    EIP1559(EIP1559TransactionMessage),
+}
+
+/// Transaction request coming from RPC
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
+pub struct TransactionRequest {
```
Jouzo, line 18: Can we re-use `CallRequest` or keep only this one instead ?

lib/ain-grpc/src/rpc.rs
```diff
@@ -461,6 +472,86 @@ impl MetachainRPCServer for MetachainRPCModule {
...
+        };
+
+        let chain_id = ain_cpp_imports::get_chain_id()
+            .map_err(|e| Error::Custom(format!("ain_cpp_imports::get_chain_id error : {e:?}")))?;
+
+        let block_number = BlockNumber::Num(self.block_number().unwrap().as_u64());
+
+        let nonce = match request.nonce {
+            Some(nonce) => nonce,
+            None => match self.get_transaction_count(from, Some(block_number)) {
```
Jouzo, line 500: Let's not call other RPCs, we can use 
```
self
                .handler
                .evm
                .get_nonce(address, block_number)
                .map_err(|e| {
                    Error::Custom(format!("Error getting address transaction count : {e:?}"))
                })?
```               

directly here

lib/ain-grpc/src/rpc.rs
```diff
@@ -461,6 +472,86 @@ impl MetachainRPCServer for MetachainRPCModule {
...
+    fn send_transaction(&self, request: TransactionRequest) -> RpcResult<String> {
+        debug!(target:"rpc","[send_transaction] Sending transaction: {:?}", request);
+
+        let from = match request.from {
+            Some(from) => from,
+            None => {
+                let accounts = match self.accounts() {
+                    Ok(accounts) => accounts,
+                    Err(e) => return Err(e),
+                };
```
Jouzo, line 484:
```suggestion
                let accounts = match self.accounts()?;
```

lib/ain-grpc/src/transaction_request.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub nonce: Option<U256>,
+    /// Pre-pay to warm storage access.
+    #[serde(default)]
+    pub access_list: Option<Vec<AccessListItem>>,
+    /// EIP-2718 type
+    #[serde(rename = "type")]
+    pub transaction_type: Option<U256>,
+}
+
+impl From<TransactionRequest> for Option<TransactionMessage> {
```
Jouzo, line 48: Would be better to implement TryFrom here and have it return an error instead of None on last match arm

lib/ain-grpc/src/transaction_request.rs
canonbrother: hmm.. yea i have doubt on this earlier.. question.. why splitting 1 struct to 2 structs since they contain same field (slightly diff)..
btw i think current design is make more sense.. obviously splitting into 2 structs is to serve for own specific purpose.. readability and easier to maintain

lib/ain-grpc/src/transaction_request.rs
canonbrother: noted that.. its a great practice.. but we ady have error handling in another way round
https://github.com/DeFiCh/ain/blob/fde4c9d71b41502533d4569e35affb2de1534cad/lib/ain-grpc/src/rpc.rs#L543-L547

Jouzo (approved):

PR #1995, opened 2023-05-17
Fix failing EVM tests

Bushstar:


lib/ain-grpc/src/transaction.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
 impl From<SignedTx> for EthTransactionInfo {
     fn from(signed_tx: SignedTx) -> Self {
         let input = if signed_tx.data().is_empty() {
-            String::from("0x")
+            String::from("0x0")
```
shohamc1, line 13: This was an intentional change (https://github.com/DeFiCh/ain/commit/e49b64cc41b6eefe78c18c90eb3be8f1c6ec4bb8), and should be `0x` not `0x0`.

[Example tx](https://etherscan.io/tx/0x72e88cea69baff2759e5e05ff8ab58afd2a79736fe968a1691a3ea98939198cb)

shohamc1 (changes requested):

shohamc1 (approved):

PR #2023, opened 2023-06-05
evm: Support contracts in functional tests

shohamc1:
Todo

- Remove checksum when generating address on node
- Get solc binary from PATH
- Install solc in `make.sh`


test/functional/test_framework/evm_provider.py
```diff
@@ -1,18 +1,16 @@
...
-from .test_node import TestNode
 
 
 class EVMProvider:
-    def __init__(self, node: TestNode):
-        self.generator = node.generate
-        self.w3 = Web3(Web3.HTTPProvider(node.get_evm_rpc()))
+    def __init__(self, rpc_url, generate):
+        self.generator = generate
+        self.w3 = Web3(Web3.HTTPProvider(rpc_url))
```
prasannavl, line 14: __init__ taking things explicitly is perfect. 
Feel free to add helpers like `from_node(node)` that constructs init from there. It's a good standard convention.

test/functional/test_framework/evm_key_pair.py
```diff
@@ -0,0 +1,31 @@
...
+            Private key does not correspond to provided address.
+            Address provided: {pkey_address}
+            Address computed: {account.address}
+            """)
+    else:
+        return pkey, pkey_address
+
+
+class KeyPair:
+    def __init__(self, node=None, pkey: str = None, pkey_address: str = None):
```
prasannavl, line 20: Let keypair not take `Node`.  Keep key pair free of node dependency. 
Instead, use static `from/with` helpers for cleaner construction. 

Eg: `generate_from_node` that uses the node to construct with the right init

test/functional/test_framework/evm_contracts.py
```diff
@@ -0,0 +1,42 @@
...
+
+from solcx import compile_standard
+
+
+class EVMContract:
+    # Solidity compiler version
+    _solc_version = "0.8.10"
+    _path_prefix = "../contracts"
+
+    def __init__(self, file_name: str, contract_name: str):
```
prasannavl, line 12: Could you apply the same construction conventions as the above for this as well and not do a error prone file read during construction? 

Let's init take content_str. Use a static helper `from_file` to construct it free to fallible methods in the ctor path.

PR #2026, opened 2023-06-05
Refactor `transferdomain` tx

Mixa84:
## Summary

- Refactor TransferDomainMessage to be in format of array of pairs of source and destination parameters each specifying address, amount and domain in which is the target and having optional data for future use if necessary.

## RPCs
transferdomain:
```
1. array                        (json array, required) A json array of src and dst json objects
     [
       {                        (json object)
         "src": {               (json object) Source arguments
           "address": "str",    (string, required) Source address
           "amount": "str",     (string, required) Amount transfered, the value is amount in amount@token format
           "domain": n,         (numeric, required) Domain of source: 1 - DVM, 2 - EVM
           "data": "str",       (string) Optional data
         },
         "dst": {               (json object) Destination arguments
           "address": "str",    (string, required) Destination address
           "amount": "str",     (string, required) Amount transfered, the value is amount in amount@token format
           "domain": n,         (numeric, required) Domain of source: 1 - DVM, 2 - EVM
           "data": "str",       (string) Optional data
         },
       },
       ...
     ]
```

## Implications

- Storage
  - [X] Database reindex required
  - [ ] None

- Consensus
  - [X] Includes consensus refactors
  - [ ] None


src/validation.cpp
```diff
@@ -2013,14 +2013,13 @@ static unsigned int GetBlockScriptFlags(const CBlockIndex* pindex, const Consens
...
 static void RevertTransferDomain(const CTransferDomainMessage &obj, CCustomCSView &mnview) {
-    if (obj.type == CTransferDomainType::DVMTokenToEVM) {
-        for (const auto& [owner, balance] : obj.from) {
-            mnview.AddBalances(owner, balance);
-        }
-    } else {
-        for (const auto& [owner, balance] : obj.to) {
-            mnview.SubBalances(owner, balance);
-        }
+    for (const auto &idx : obj.transfers) {
```
prasannavl, line 2016:
```suggestion
   // NOTE: Each domain's revert is handle by it's own domain module. This function reverts only the DVM aspect. EVM will handle it's own revert.  
    for (const auto &idx : obj.transfers) {
```

dcorral:
Usage description seems outdated 

https://github.com/defich/ain/blob/8213f3d7e7120a22ae7ba4e5ec18c1734603afcd/src/masternodes/rpc_accounts.cpp#L1990-L1993

dcorral:
The utility of having amount in the dst object is dependent on the nature of the transactions you want to support. For one-to-one transactions, it's redundant, but for one-to-many transactions, it's necessary. As it is now it is a one-to-one transfer so it makes no sense to have it on the dst object.

Mixa84:
> The utility of having amount in the dst object is dependent on the nature of the transactions you want to support. For one-to-one transactions, it's redundant, but for one-to-many transactions, it's necessary. Currently it is a one-to-one transfer and amount can be extracted from src directly.

Yes I agree, but this is done like this because the tx structure needs to be flexible to support all future uses. For now RPC is made strict as tx data with an JSON object, but easier usage like specifying amount and destination address and maybe one domain will be additionally added to support most default usage cases.

PR #2033, opened 2023-06-07
evm: Install solc from make.sh

shohamc1:


make.sh
```diff
@@ -519,6 +519,13 @@ pkg_install_rust() {
     _fold_end
 }
 
+pkg_install_solc() {
+    _fold_start "pkg-install-solc"
+    python3 -m pip install py-solc-x
```
prasannavl, line 524: Isn't pipx the recommended way to do this?

shohamc1: We would have to configure the Python tests to run in the pipx virtual environment in that case.

prasannavl: IIRC, pipx install will just install globally (user) in a cleaner env  rather than how pip was polluting `$HOME/.local/`. 
Believe calling it without a specific env is just a better behaved (finally) drop in replacement for pip install. 

Ubuntu 23.04+ will make pip install fail directly

prasannavl: Note though: If Ubuntu 20.04 has py-solc-x, would rather just use that. Could you check that? That removes the need for the python management.

shohamc1: Just tried `pipx`, it does't work with the following error:

```
No apps associated with package py-solc-x. Try again with '--include-deps' to include apps of dependent packages, which are listed above. If you are attempting to install a library,
pipx should not be used. Consider using pip or a similar tool instead.
```

Which makes sense since `py-solc-x` is a library. It is also not available on the Ubuntu repositories.

prasannavl:
LGTM once conflicts resolved.

prasannavl (approved):

PR #2034, opened 2023-06-08
evm: Load state from genesis.json

shohamc1:
#### Summary

Added `-ethstartstate=path/genesis.json` to support eth genesis configuration


Jouzo:
This JSON-based initialization is exclusively reserved for regtest environments.

### Warning

Loading state from JSON will overwrite previous stored state

### Errors

The use of `-ethstartstate` flag will return an error if an attempt is made to load a genesis state from a JSON file outside of a regtest environment.

canonbrother (approved):

PR #2038, opened 2023-06-09
Set max gas limit per block

Bushstar:
Limits the amount of gas used per block to 30 million.


src/validation.cpp
```diff
@@ -2637,7 +2642,16 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
+            uint64_t gasUsed{};
+            const auto res = ApplyCustomTx(accountsView, view, tx, chainparams.GetConsensus(), pindex->nHeight, gasUsed, pindex->GetBlockTime(), nullptr, i, evmContext);
+
+            totalGas += gasUsed;
+            if (totalGas > MAX_BLOCK_GAS_LIMIT) {
+                return state.Invalid(ValidationInvalidReason::CONSENSUS,
+                                     error("%s: ApplyCustomTx failed. Gas limit: %d Gas used: %d", __func__, MAX_BLOCK_GAS_LIMIT, totalGas),
+                                     REJECT_CUSTOMTX, "over-gas-limit");
+            }
+
```
Jouzo, line 2654: If the TX goes above the MAX_BLOCK_GAS_LIMIT, we need to reject it and try the following ones.
The if block needs to subtract this TX used_gas so the following ones have a chance to get in the block

prasannavl: How about this?

- See if there's atleast one EVM TX that's included. If it is, keep the same logic, skip the rest and fail the block. 
- If it's not, reject it 

This prevents head of line blocking by a single TX, but safely rejects others, without us having to worry about reordering for now.

lib/ain-rs-exports/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub mod ffi {
     pub struct ValidateTxResult {
         nonce: u64,
         sender: [u8; 20],
+        used_gas: u64,
```
prasannavl, line 43: How about `Option<u64>`?

prasannavl: No support on CXX at the moment

PR #2051, opened 2023-06-12
evm: Add `input` to eth_call

### LINKED ISSUE #2048: EVM: standard (?) field `input` not allowed in eth transaction
even a simple eth_call fails with 
`unknown field input, expected one of from, to, gasPrice, maxFeePerGas, maxPriorityFeePerGas, gas, value, data, nonce, accessList, type`
if the input field is provided. But this seems to be standard in eth. The generated tx by web3.js contains `data` and `input` with the same content.

shohamc1:
Try to use `data` in call request when `input` is not available. When `input` is available, `data` is ignored.

References:
- https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/internal/ethapi/transaction_args.go#LL47C1-L49C71
- https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/internal/ethapi/transaction_args.go#L67

Fixes #2048

### DIFF
--- a/lib/ain-grpc/src/rpc/eth.rs
+++ b/lib/ain-grpc/src/rpc/eth.rs
@@ -269,7 +269,7 @@ impl MetachainRPCServer for MetachainRPCModule {
                 to,
                 value.unwrap_or_default(),
                 // https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/internal/ethapi/transaction_args.go#L67
-                /// We accept "data" and "input" for backwards-compatibility reasons.
+                // We accept "data" and "input" for backwards-compatibility reasons.
                 //  "input" is the newer name and should be preferred by clients.
                 // 	Issue detail: https://github.com/ethereum/go-ethereum/issues/15628
                 &input


Jouzo (approved):

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -267,7 +268,13 @@ impl MetachainRPCServer for MetachainRPCModule {
                 from,
                 to,
                 value.unwrap_or_default(),
-                &data.map(|d| d.0).unwrap_or_default(),
+                // https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/internal/ethapi/transaction_args.go#L67
+                /// We accept "data" and "input" for backwards-compatibility reasons.
```
kuegi, line 272: produces a rust warning about doc comment. should be // right?

kuegi:
IMHO input is also needed for a normal SC transaction. web3.js sends it on every txs alongside `data`

lib/ain-grpc/src/rpc/eth.rs
prasannavl, line 272:
```suggestion
                // We accept "data" and "input" for backwards-compatibility reasons.
```

prasannavl:
Thanks for this catch @kuegi. Yes, `input` seems to be the new data type. `data` is just maintained for compatibility. 

This is will amended throughout the usages of Txs in the spec. Just going ahead and merging call part at the moment.

prasannavl (approved):

Issue #2046, opened 2023-06-11
defid 4.0.0.-beta1 not running with GLIBC 2.31


Stonygan:
## Summary

defid 4.0.0-beta1 not running with Debian GLIBC 2.31-13+deb11u6) 2.31

## Steps to Reproduce

~/.defi/defid

`.defi/defid: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.32 not found (required by .defi/defid)`
`.defi/defid: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.34 not found (required by .defi/defid)`
`.defi/defid: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.33 not found (required by .defi/defid)`

same with defi-cli.

Is it planned to compile defid the same as with previous versions or is this beta the current build status as it will be later?

## Environment

[Please fill all of the following or NA if not applicable]

- Node Version: defid 4.0.0-beta1
- Block height on bug if applicable: NA
- TX or TX type on bug if applicable:  NA
- OS with version: debian 11.7
- Any other relevant environment info:  
- Linux 5.10.0-23-amd64 x86_64

prasannavl:
Quick glance: https://tracker.debian.org/pkg/glibc - old stable seems to be on: [2.28-10+deb10u1](https://packages.debian.org/source/oldstable/glibc). 

Ubuntu 20.04: https://launchpad.net/ubuntu/+source/glibc is on 2.31.

However, will still have to look at why your logs went as high up as GLIBC_2.33. We'll check it out - and lower our bases if possible. Ideally would like to support Debian 11. Although this has now moved to oldstable as of 2 days ago.

PR #2054, opened 2023-06-12
Further build refinements, use ain-builder, rust 1.70

prasannavl:
### Builder base
- Switch builder base images to defi/ain-builder: https://github.com/DeFiCh/containers/blob/main/ain-builder/Dockerfile
- This is used to avoid re-doing the work of updating containers on each run. 
  - This also often results in random errors due to Ubuntu repo timeouts. This avoids this by using a pre-warmed up base.
  - In addition, helps speed up the CI
- Note: There is no difference in how the post-base items are handled. They still go through the same dep installs - and follow the exact same procedure as they would on a fresh container.
  - This just provides it pre-warmed up, so they can be expected to take a faster path. 
  - This ensure any changes that's made here are still executed, if the base deps are out of date, etc.
  
 ### CI 
 
- Fix tests failing due to incorrect init
- Enable EVM tests on master
- Cleanup workflows

### make.sh

- Add PATH to `ci-export-vars`
- Switch to `apt-get` on automation.  

### Misc

- Switch rust version to latest stable: 1.70 
- In addition, this also fixes https://github.com/DeFiCh/ain/issues/2046
- Remove stale comment from docker images


contrib/dockerfiles/aarch64-linux-gnu.dockerfile
```diff
@@ -17,7 +17,7 @@ RUN ./make.sh ci-setup-deps-target
 
 COPY . .
 RUN ./make.sh build-deps
-RUN MAKE_DEBUG=${DEBUG} ./make.sh build-conf
+RUN MAKE_DEBUG=${MAKE_DEBUG} ./make.sh build-conf
```
prasannavl, line 20: We can remove this now?

PR #2059, opened 2023-06-13
Add ICX feature flag. Fix bug on mainnet.

Mixa84:
## Summary

- Add ICX feature flag in attributes
- Fix makerBonus bug on mainnet (it was fixed for testnet and regtest already)

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None

### DIFF
--- a/src/masternodes/mn_checks.cpp
+++ b/src/masternodes/mn_checks.cpp
@@ -2057,6 +2057,9 @@ public:
     }
 
     Res operator()(const CICXCreateOrderMessage &obj) const {
+        if (!IsICXEnabled(height, mnview))
+            return Res::Err("Cannot create tx, ICX is not enabled");
+
         Require(CheckCustomTx());
 
         CICXOrderImplemetation order;
@@ -2082,6 +2085,9 @@ public:
     }
 
     Res operator()(const CICXMakeOfferMessage &obj) const {
+        if (!IsICXEnabled(height, mnview))
+            return Res::Err("Cannot create tx, ICX is not enabled");
+
         Require(CheckCustomTx());
 
         CICXMakeOfferImplemetation makeoffer;
@@ -2122,6 +2128,9 @@ public:
     }
 
     Res operator()(const CICXSubmitDFCHTLCMessage &obj) const {
+        if (!IsICXEnabled(height, mnview))
+            return Res::Err("Cannot create tx, ICX is not enabled");
+
         Require(CheckCustomTx());
 
         CICXSubmitDFCHTLCImplemetation submitdfchtlc;
@@ -2235,6 +2244,9 @@ public:
     }
 
     Res operator()(const CICXSubmitEXTHTLCMessage &obj) const {
+        if (!IsICXEnabled(height, mnview))
+            return Res::Err("Cannot create tx, ICX is not enabled");
+
         Require(CheckCustomTx());
 
         CICXSubmitEXTHTLCImplemetation submitexthtlc;
@@ -2333,6 +2345,9 @@ public:
     }
 
     Res operator()(const CICXClaimDFCHTLCMessage &obj) const {
+        if (!IsICXEnabled(height, mnview))
+            return Res::Err("Cannot create tx, ICX is not enabled");
+
         Require(CheckCustomTx());
 
         CICXClaimDFCHTLCImplemetation claimdfchtlc;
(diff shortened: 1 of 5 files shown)


Bushstar (approved):

src/masternodes/mn_checks.cpp
```diff
@@ -2385,7 +2400,8 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
         DCT_ID BTC = FindTokenByPartialSymbolName(CICXOrder::TOKEN_BTC);
         if (order->idToken == BTC && order->orderPrice == COIN) {
```
prasannavl, line 2401: Let's abstract a bit more so it's clearer, self documenting on what it does and take the bugged code out of the main path. 

``
```
// This is a function for bugged blockchain region check, and maintained purely for maintaining compatibility with older network data.
ICXLegacyBugPath(params..) {
 lambda IsICXBuggedRegion(network, height) {
   
  }
  
  lamda BuggedICXTransfer() { .. }

  IsICXBuggedRegion(network, height) {
   ..BuggedICXTransfer();
  }
}



... 

if (!ICXLegacyBugPath(params)) {

}
```

src/masternodes/mn_checks.cpp
```diff
@@ -4019,31 +4035,35 @@ bool IsDisabledTx(uint32_t height, CustomTxType type, const Consensus::Params &c
...
+        // ICXCreateOrder      = '1',
+        // ICXMakeOffer        = '2',
+        // ICXSubmitDFCHTLC    = '3',
+        // ICXSubmitEXTHTLC    = '4',
+        // ICXClaimDFCHTLC     = '5',
+        // ICXCloseOrder       = '6',
+        // ICXCloseOffer       = '7',
+
+        // disable ICX orders for all networks other than testnet
+        if (Params().NetworkIDString() == CBaseChainParams::REGTEST ||
```
prasannavl, line 4048: Let's turn it such that this checks the `IsICXBuggedRegion()` region. And it's a single source of check.

src/masternodes/mn_checks.cpp
```diff
@@ -4019,31 +4035,35 @@ bool IsDisabledTx(uint32_t height, CustomTxType type, const Consensus::Params &c
...
-    // Leaving close orders, as withdrawal of existing should be ok
-    switch (type) {
-        case CustomTxType::ICXCreateOrder:
-        case CustomTxType::ICXMakeOffer:
-        case CustomTxType::ICXSubmitDFCHTLC:
-        case CustomTxType::ICXSubmitEXTHTLC:
-        case CustomTxType::ICXClaimDFCHTLC:
-            return true;
-        default:
+    if (height < Params().GetConsensus().NextNetworkUpgradeHeight) {
```
prasannavl, line 4038: Consensus is a pass down as args. Don't pull from global state. Use local context, and also pull the height above, if we still need this after ICXEnabled. Would be preferred if we abstract it into ICXEnabled and let that take clear and full ownership of ICX and remove this entirely.

src/masternodes/mn_checks.cpp
```diff
@@ -5004,19 +4992,23 @@ bool IsMainNetwork() {
...
-
-    const CDataStructureV0 enabledKey{AttributeTypes::Param, ParamIDs::Feature, DFIPKeys::ICXEnabled};
-    auto attributes = view.GetAttributes();
-    assert(attributes);
-    return attributes->GetValue(enabledKey, false);
+    // ICX transactions allowed before NextNetwrokUpgrade and some of these conditions
+    else if (height < consensus.FortCanningParkHeight || IsRegtestNetwork() || (IsTestNetwork() && static_cast<int>(height) >= 1250000))
+        return true;
+    // ICX transactions disabled
+    else if (height >= consensus.FortCanningParkHeight && height < consensus.NextNetworkUpgradeHeight)
```
Bushstar, line 5006: Looks like you can just skip this else if and return false. Thinking this function might generate a warning about no return at the end.

Bushstar: It does indeed generate this warning.

```
masternodes/mn_checks.cpp: In function ‘bool IsICXEnabled(int, const CCustomCSView&, const Consensus::Params&)’:
masternodes/mn_checks.cpp:5014:1: warning: control reaches end of non-void function [-Wreturn-type]
```
In addition to that there are  these warnings as well.

```
masternodes/mn_checks.cpp: In lambda function:
masternodes/mn_checks.cpp:2406:48: warning: comparison of integer expressions of different signedness: ‘uint32_t’ {aka ‘unsigned int’} and ‘const int’ [-Wsign-compare]
 2406 |                     (IsMainNetwork() && height >= consensus.NextNetworkUpgradeHeight))
      |                                         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
masternodes/mn_checks.cpp: In function ‘bool IsDisabledTx(uint32_t, CustomTxType, const Consensus::Params&)’:
masternodes/mn_checks.cpp:4037:10: warning: unused variable ‘nextNetworkUpgradeHeight’ [-Wunused-variable]
 4037 |     auto nextNetworkUpgradeHeight = static_cast<uint32_t>(consensus.NextNetworkUpgradeHeight);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
```

Mixa84: Thnx for the review, resolved!

Bushstar (approved):

PR #2060, opened 2023-06-13
EVM: Implement missing block_number_to_u256 arms

Jouzo:
## Summary

- Implement missing block_number_to_u256 arms

### TBD:

What is considered `safe` in our case. Ref: https://docs.alchemy.com/reference/ethereum-developer-guide-to-the-merge#what-are-safe-and-finalized

> The most recent crypto-economically secure block, cannot be re-orged outside of manual intervention driven by community coordination. Intuitively, this block is “unlikely” to be re-orged.


lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -217,36 +217,39 @@ impl MetachainRPCModule {
...
+                .handler
+                .storage
+                .get_block_by_number(&U256::from(n))
+                .map(|block| block.header.number),
+            BlockNumber::Earliest => self
+                .handler
+                .storage
+                .get_block_by_number(&U256::zero())
+                .map(|block| block.header.number),
+            BlockNumber::Safe | BlockNumber::Finalized => self
```
prasannavl, line 236: Let's also try to combine the same workflow that the core block uses to detect if there are > 120 and pull up a warning. 

- If confirmation > N + no chain > 120 detected = Safe would be the only reasonable way to add this to maintain a semblance of parity with ethereum ecosystem

PR #2069, opened 2023-06-14
EVM: add `input` to eth_sendTransaction

kuegi:
same as #2051 did it for eth_call, nor also for transactions. as noticed in #2048 

Try to use data in transaction request when input is not available. When input is available, data is ignored.

References:

* https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/internal/ethapi/transaction_args.go#LL47C1-L49C71
* https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/internal/ethapi/transaction_args.go#L67


lib/ain-grpc/src/transaction_request.rs
```diff
@@ -54,7 +56,10 @@ impl From<TransactionRequest> for Option<TransactionMessage> {
                 gas_price: req.gas_price.unwrap_or_default(),
                 gas_limit: req.gas.unwrap_or_default(),
                 value: req.value.unwrap_or_default(),
-                input: req.data.map(Bytes::into_vec).unwrap_or_default(),
+                input: req
+                    .input
+                    .map(Bytes::into_vec)
+                    .unwrap_or(req.data.map(Bytes::into_vec).unwrap_or_default()),
```
shohamc1, line 62: This should be a cleaner way to achieve this.

```suggestion
                input: req
                    .input
                    .or_else(|| req.data)
                    .map(Bytes::into_vec)
                    .unwrap_or_default(),
```

shohamc1 (approved):

Jouzo (approved):

prasannavl:
Neat. Thanks @kuegi :)

PR #2077, opened 2023-06-16
Fix PoolPair value overflow undefined behavior bug

sieniven:
## Summary

- Fix bug when converting PoolPair structs from COIN to satoshis without value checks, causing value overflow and undefined behaviour.

Context:
- Legacy transactions with maxPrice attribute unset will set maxPrice to POOLPRICE_MAX.
- PoolSwap transactions with maxPrice set above maximum satoshis will be rejected.
- Conditional statement in mn_checks.cpp::ExecuteSwap previously skips entering the check to reject if price paid post-swap is above max price provided if maxPrice is set to legacy constant POOLPRICE_MAX, since any swap price is accepted.

Fixes:
- checkMaxPoolPrice function ensures no value overflow and value of PoolPair is below the maximum satoshis (1.2B).
- setMaxPoolPrice helper function to safeguard against setting numeric limits in both integer and fraction values.
- New PoolSwap transactions, when maxPrice is not defined, will default to setting numeric limits of satoshis which will be converted to COIN and stored in integer and fraction fields. This further safeguards against going into undefined behaviour.
- Fix default poolPrice in mn_checks.cpp::ExecuteSwap and mn_checks.cpp::SwapToDFIorDUSD to use setMaxPoolPrice.
- Conditional statement in mn_checks will check if maxPrice is maximum, accounting for both legacy POOLPRICE_MAX and the new default maximum PoolPrice.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


sieniven:
Tested consensus changes in the PR with the old rpc_poolpair.cpp on poolpair and poolswap functional tests to ensure backwards compatibility.
Screenshot 2023-06-16 at 4 27 14 PM

Bushstar:
Let's run a full sync test when they are working again to double check the consensus changes.

Bushstar (approved):

PR #2085, opened 2023-06-19
evm: Implement filter RPCs

shohamc1:
Adds the following RPCs
- `eth_newFilter`
- `eth_newBlockFilter`
- `eth_getFilterChanges`
- `eth_uninstallFilter`
- `eth_getFilterLogs`
- `eth_newPendingTransactionFilter`


lib/ain-evm/src/filters.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    pub fn delete_filter(&self, filter_id: usize) -> bool {
+        let mut filters = self.filters.write().unwrap();
+
+        match filters.remove(&filter_id) {
+            Some(_) => true,
+            None => false,
+        }
+    }
```
Jouzo, line 205:
```suggestion
pub fn delete_filter(&self, filter_id: usize) -> bool {
        self.filters.write().unwrap().remove(&filter_id).is_some()
    }
```

lib/ain-evm/src/filters.rs
```diff
@@ -0,0 +1,206 @@
...
+                }
+                _ => {}
+            }
+        }
+    }
+
+    pub fn get_entries_from_filter(&self, filter_id: usize) -> Result<Vec<H256>, &str> {
+        let mut filters = self.filters.write().unwrap();
+
+        let filter = filters.get_mut(&filter_id).unwrap();
```
Jouzo, line 189:
```suggestion
        let Some(filter) = filters.get_mut(&filter_id) else {
            return Err("Filter not found");
        };
```

lib/ain-evm/src/filters.rs
```diff
@@ -0,0 +1,206 @@
...
+        match filters.get(&filter_id) {
+            Some(filter) => Ok(filter.clone()),
+            None => Err("Unable to find filter"),
+        }
+    }
+
+    pub fn update_last_block(&self, filter_id: usize, block_height: U256) {
+        let mut filters = self.filters.write().unwrap();
+
+        let filter = filters.get_mut(&filter_id).unwrap();
```
Jouzo, line 146: Need to handle None here

lib/ain-evm/src/filters.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub fn create_logs_filter(
+        &self,
+        address: Option<Vec<H160>>,
+        topics: Option<Vec<Option<H256>>>,
+        from_block: U256,
+        to_block: U256,
+        current_block_height: U256,
+    ) -> usize {
+        let mut filters = self.filters.write().unwrap();
+        let mut filter_id = self.filter_id.write().unwrap();
```
Jouzo, line 86: Could probably hold this lock less

lib/ain-evm/src/log.rs
```diff
@@ -99,4 +100,29 @@ impl LogHandler {
...
+        while block_number <= filter.to_block {
+            debug!("Will query block {block_number}");
+            block_numbers.push(block_number);
+            block_number += U256::one();
+        }
+
+        block_numbers
+            .into_iter()
+            .flat_map(|block_number| {
+                self.get_logs(filter.clone().address, filter.clone().topics, block_number)
```
Jouzo, line 123: Can update get_logs to take the options by ref and remove all get_logs clones

Jouzo (commented):
Few nitpicks but logic looks good to me

lib/ain-evm/src/filters.rs
shohamc1: These should be safe to use since we get the filter to check if it exists at the beginning of every RPC call. I have added in the changes in case we use these methods elsewhere.

PR #2084, opened 2023-06-19
fix(evm): store genesis block into db

canonbrother:
## Summary

some fixes: 
- [x] extra_data type vec -> bytes to allow deserialised
- [x] set base fee for genesis
- [x] store genesis into block in db

## Hive

hive-result

PR #2086, opened 2023-06-19
Set initial base fee when loading state from json

Jouzo:
## Summary

- Set initial base fee when loading state from json


Jouzo:
Closed for #2084

PR #2092, opened 2023-06-20
evm: Check nonce in `finalize_block`

shohamc1:


src/miner.cpp
```diff
@@ -282,12 +282,12 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
 
         xvm = XVM{0,{0, uint256(blockHash), blockResult.miner_fee / CAMOUNT_TO_GWEI}};
 
-        std::vector<std::string> failedTransactions;
-        for (const auto& rust_string : blockResult.failed_transactions) {
-            failedTransactions.emplace_back(rust_string.data(), rust_string.length());
+        std::vector<std::string> evictedTransactions;
```
Bushstar, line 285: failedTransactions should also be removed from the block, the only failed transactions that are actually removed are TransferDomain TXs. All failed EVM TXs are not removed.

shohamc1: Failed bridge transactions are added to `evicted_transactions` and will be removed.

PR #2099, opened 2023-06-21
EVM precompile support

Jouzo:
## Summary

- Introduces MetachainPrecompiles, which includes all Ethereum precompiles available as of the Shanghai fork.


test/functional/feature_evm.py
```diff
@@ -349,6 +349,7 @@ def run_test(self):
         # Test rollback of EVM related TXs
         self.nodes[0].invalidateblock(self.nodes[0].getblockhash(101))
         assert_equal(self.nodes[0].getblockcount(), 100)
+        assert_equal(True, False)
```
Bushstar, line 352: I personally just use...

`assert(False)`

PR #2100, opened 2023-06-21
Guard DUSD loops with Changi Intermediate 2

Bushstar:
Adds the ChangiIntermediate3 fork and puts the DUSD loop changes behind it.


kuegi:
@Bushstar not needed since its anyway behind the feature flag, and we just don't activate the flag right away. right? so no need for extra fork height.

Bushstar:
Since we will be activating it I prefer to put it behind a fork. We tell everyone to update and only activate after the fork has passed.

kuegi:
> Since we will be activating it I prefer to put it behind a fork. We tell everyone to update and only activate after the fork has passed.

agreed, but still: wouldn't need a fork. we can also say "will be activated after block X, everybody needs to update". no difference for users if its an official fork guard or not. just more code

PR #2108, opened 2023-06-22
Add -regtest-dev

Bushstar:
Allows running of regtest with -regtest-dev to enable mining of blocks with `-gen=1` and `generatetoaddress` and enables all forks from block 1.


Mixa84 (dismissed):
Working good, tested it with some test.

PR #2123, opened 2023-06-27
make.sh: Add Fedora support

shohamc1:
- Adds fedora dependencies to make.sh for local development.
- Configure clang compiler to not use version suffix on certain distros.


prasannavl:
Closing the loop: 

Reasons:
- makefiles itself are distro agnostic and can be compiled from most standard distros. 
- `./make.sh` flow is focused workflow for Debian/Ubuntu. 
- Goal: Do it for one platform and do it well. 
- Adding support for others is not just a non-goal for make.sh, but also discouraged to keep make.sh complexity under check and leaner.   
- That said, adding it to docs or script that wraps or replaces make.sh are absolutely welcome and encouraged to be contributed under: `./contrib/` dir -- just not suited for make.sh flow.

PR #2124, opened 2023-06-27
Simplify and refactor transfer domain 

prasannavl:
## Summary

- Refactors TransferDomain into smaller units with clearer concerns and simplified code branching 
- Introduce `ChangiBuggyIntermediates` to keep all intermediate maintained code branches
  - TODO: Move all intermediate code into this class, so the main code path is bug-free and clear. 
  - Easier to remove to code branches with confidence without convoluted branches. 
- Refactor `VMDomainMapView` into `VMDomainGraphView`
  - Reuses `VMDomainEdge` type along with the result of consenus. 

### Consensus changes
- Adds a size check for TransferDomain objects.
- If a null object TransferDomain is added to changi, that'll make it backward incompatible. 
  - If made incompatible, we'll have to add it into changi intermediates again. 

## Implications

- Storage
  - [x] None

- Consensus
  - [x] Includes backward compatible changes
  - [ ] None


src/masternodes/changiintermediates.h
```diff
@@ -0,0 +1,23 @@
+#ifndef DEFI_MASTERNODES_CHANGIINTERMEDIATES_H
+#define DEFI_MASTERNODES_CHANGIINTERMEDIATES_H
+
+#include <consensus/params.h>
```
Bushstar, line 4: Include should be in the cpp file with forward declarations here instead. Includes in headers should be as minimal as possible to avoid include loops.

src/masternodes/changiintermediates.h
```diff
@@ -0,0 +1,23 @@
...
+#include <consensus/tx_check.h>
+#include <masternodes/evm.h>
+#include <masternodes/masternodes.h>
+#include <masternodes/mn_checks.h>
+
+/// The file and class that has all the old things that will be removed.
+/// The goal of this entire class and file is to be removed. It's a collection of the 
+/// accumulated bugs in early releases. 
+class ChangiBuggyIntermediates {
```
Bushstar, line 13: We only need a function, this class is superfluous.

src/masternodes/mn_checks.cpp
```diff
@@ -3975,130 +3976,102 @@ Res HasAuth(const CTransaction &tx, const CCoinsViewCache &coins, const CScript
...
-enum class DomainTransferType : uint8_t {
-    DVM,
-    EVM,
-};
+static Res ValidateTransferDomainScripts(const CScript &srcScript, const CScript &destScript, VMDomainEdge aspect) {
+    CTxDestination src, dest;
+    auto res = ExtractDestination(srcScript, src);
+    if (!res) return DeFiErrors::ScriptUnexpected(srcScript);
 
-static bool DomainTransferAllowedAddress(const CScript &address, DomainTransferType type) {
```
Bushstar, line 3983: There are many places in mn_checks are further across the codebase use this exact check for PKHashType and WitV0KeyHashType. This function would have been used for them at a later date. I did not update them when adding this function to avoid further conflicts with the mn_check split PR.

src/masternodes/changiintermediates.h
prasannavl: There are other items that need lifting for branching?

src/masternodes/changiintermediates.h
prasannavl: Oh right, my bad.. missed this

src/masternodes/changiintermediates.h
Bushstar: Any of those branches need to be inside a class?

prasannavl: The idea is to refactor them to them so all the buggy code can be brought into one place for removal. And forces us to cleanup the code paths prior to removing the forks.

src/masternodes/mn_checks.cpp
prasannavl: Yes, please. Let's further abstract from this point and lift up as needed to reuse if that ok to you?

Bushstar (dismissed):

Mixa84 (dismissed):

PR #2125, opened 2023-06-27
Move EVM init till after chainstate init

Bushstar:


prasannavl:
TODO(later): Abstract.and cleanup cc/ @sieniven

PR #2135, opened 2023-06-30
Limit Eth TXs from the same sender in the mempool to 64

Bushstar:


src/validation.cpp
```diff
@@ -894,11 +894,37 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
...
+        if (isEvmTx) {
+            auto txMessage = customTypeToMessage(txType);
+            auto res = CustomMetadataParse(height, chainparams.GetConsensus(), metadata, txMessage);
+            if (!res) {
+                return state.Invalid(ValidationInvalidReason::TX_NOT_STANDARD, error("Failed to parse EVM tx metadata"), REJECT_INVALID, "failed-to-parse-evm-tx-metadata");
+            }
+            
+            const auto obj = std::get<CEvmTxMessage>(txMessage);
+            CrossBoundaryResult result;
+            const auto txResult = evm_try_prevalidate_raw_tx(result, HexStr(obj.evmTx), false);
```
prasannavl, line 920: Check `result`

src/validation.cpp
```diff
@@ -894,11 +894,37 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
...
+            if (pool.ethTxsBySender[txResult.sender].size() >= MEMPOOL_MAX_ETH_TXS) {
+                return state.Invalid(ValidationInvalidReason::TX_MEMPOOL_POLICY, error("Too many Eth trransaction from the same sender in mempool. Limit %d.", MEMPOOL_MAX_ETH_TXS), REJECT_INVALID, "too-many-eth-txs-by-sender");
```
prasannavl, line 922: TODO: Here `ethTxsBySender` can still leak when invalid.

src/validation.cpp
Bushstar: Checked in mn_checks

prasannavl (approved):

PR #2140, opened 2023-07-03
Fixes and tests for vmmap RPC

DocteurPing:
## Summary

- Fixes for vmmap RPC in-line with bech32 support
- More meaningful tests for the vmmap/logvmmaps RPCs. 

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


test/functional/feature_evm_rpc.py
```diff
@@ -156,6 +166,15 @@ def test_vmmap(self):
         evm_block = '0x0000000000000000000000000000000000000000000000000000000000000000'
         assert_raises_rpc_error(-32600, "DB r/w failure: 0000000000000000000000000000000000000000000000000000000000000000", self.nodes[0].vmmap, evm_block, 6)
 
+        # Check if invalidate block is working for mapping. After invalidating block, the transaction and block shouldn't be mapped anymore.
+        self.nodes[0].invalidateblock(dvm_block)
+        assert_raises_rpc_error(-32600, "DB r/w failure: " + dvm_block, self.nodes[0].vmmap, dvm_block, 5)
```
prasannavl, line 171: This exposes a different bug. There should never be r/w failures unless it's a disk failure. The `ReadBy` calls shouldn't be returning r/w, just a different not found error.

DocteurPing: We don't have the possibility to check the kind of error with readBY. the function is just returning a bool. is there a way to retrieve the error type afterwards ?

prasannavl: Don't think so for Read. Currently, Read false is more of, just key doesn't exist. So just switch it to a `KeyNotFound` error. For Write it's D/B error.

Bushstar (approved):

PR #2161, opened 2023-07-06
Refine rs workflow and refactors

prasannavl:
## Summary

This starts to clean up the rust integration and rust libraries.

- Cleanup integration workflow with correct startup and shutdown hooks. 
  - Currently without it, in edge cases can result in inconsistent data. This should fix that.
  - TODO: Refactor more to maintain a coherent lock flow across C++ and Rust sides
- This also starts to establish a proper terminology on the Rust libs with all features as it's own microservice.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


src/init.cpp
```diff
@@ -1968,18 +1980,6 @@ bool AppInitMain(InitInterfaces& interfaces)
...
             }
 
-            std::vector<std::pair<std::string, uint16_t>> eth_endpoints;
-            std::vector<std::pair<std::string, uint16_t>> g_endpoints;
-            SetupRPCPorts(eth_endpoints, g_endpoints);
-
-            // Start the GRPC and ETH RPC servers
-            // Default to using the first address passed to bind
-            CrossBoundaryResult result;
-            start_servers(result, eth_endpoints[0].first + ":" + std::to_string(eth_endpoints[0].second), g_endpoints[0].first + "." + std::to_string(g_endpoints[0].second));
```
Bushstar, line 1978: This needs to be called before VerifyDB, moving it to after will cause VerifyDB to crash if it calls ConnectBlock on blocks with Eth TXs.

prasannavl: Ah, great catch. Thanks! Will resolve.

PR #2162, opened 2023-07-07
vmmap: DVM - EVM block number support

DocteurPing:
## Summary

- This PR is adding mapping of block number for the `vmmap` RPC. 

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-rs-exports/src/lib.rs
```diff
@@ -426,3 +430,26 @@ fn start_servers(result: &mut CrossBoundaryResult, json_addr: &str, grpc_addr: &
...
+    let block = RUNTIME
+        .handlers
+        .storage
+        .get_block_by_number(&U256::from(height));
+    if block.is_some() {
+        return block.unwrap().header.hash().to_fixed_bytes()
+    }
+    [0; 32]
+
+}
```
shohamc1, line 444:
```suggestion
fn evm_get_block_hash_by_number(height: u64) -> [u8; 32] {
    match RUNTIME
        .handlers
        .storage
        .get_block_by_number(&U256::from(height))
    {
        Some(block) => block.header.hash().to_fixed_bytes(),
        None => [0; 32],
    }
}
```

shohamc1 (commented):
Would be better to have block height passed around as `U256` since `u64` will be susceptible to overflows much earlier.

lib/ain-rs-exports/src/lib.rs
```diff
@@ -98,6 +99,9 @@ pub mod ffi {
         ) -> Vec<u8>;
 
         fn evm_disconnect_latest_block();
+
+        fn evm_get_block_hash_by_number(height: u64) -> [u8; 32];
```
prasannavl, line 103: Please use `CrossBoundaryResult` so any actual errors from inner APIs aren't swallowed and instead propagated to layers up, and differentiate between 0 and a real error.

DocteurPing: Fixed in https://github.com/DeFiCh/ain/pull/2162/commits/edc4ea7ef3f3206e67a8c0c7c04a535cc42df68f

src/masternodes/rpc_evm.cpp
```diff
@@ -144,30 +147,83 @@ UniValue evmtx(const JSONRPCRequest& request) {
...
+        return pindex->GetBlockHeader().deprecatedHeight;
+    }
+    return dvm_block.msg;
+}
+
+UniValue handleMapBlockNumberDVMToEVMRequest(const std::string &input) {
+    uint64_t height;
+    bool success = ParseUInt64(input, &height);
+    if (!success) {
+        return DeFiErrors::InvalidBlockNumber(input).msg;
```
prasannavl, line 178: Note that these won't throw. Just return a success string.

src/masternodes/rpc_evm.cpp
```diff
@@ -11,42 +12,39 @@ enum class VMDomainRPCMapType {
...
-enum class VMDomainIndexType {
-    BlockHash,
-    TxHash
-};
+static int VMDomainRPCMapTypeCount = 9;
 
+enum class VMDomainIndexType { BlockHash, TxHash };
 
-UniValue evmtx(const JSONRPCRequest& request) {
+UniValue evmtx(const JSONRPCRequest &request) {
```
shohamc1, line 24: Incorrect formatting

src/masternodes/rpc_evm.cpp
```diff
@@ -144,42 +149,81 @@ UniValue evmtx(const JSONRPCRequest& request) {
     return send(MakeTransactionRef(std::move(rawTx)), optAuthTx)->GetHash().ToString();
 }
 
+UniValue handleMapBlockNumberEVMToDVMRequest(const std::string &input) {
```
shohamc1, line 152: Maybe we can change this function to `mapBlockNumberEVMToDVMRequest`? Don't think we have any other functions called `handle*`.

prasannavl: `handle` is for things that return actually validates and returns JSON / univalue errors. If it's map, it should only have returned been a pure map fn, which here it's not. Should be there in another discussion thread.

sieniven (approved):

sieniven (approved):

PR #2164, opened 2023-07-07
Refactor ERC55 address support

Bushstar:


prasannavl (dismissed):

prasannavl:
Tests still need fixing. LGTM once sorted.

src/compressor.cpp
```diff
@@ -22,7 +22,7 @@ static bool IsToKeyID(const CScript& script, CKeyID &hash)
     if (script.size() == 25 && script[0] == OP_DUP && script[1] == OP_HASH160
                             && script[2] == 20 && script[23] == OP_EQUALVERIFY
                             && script[24] == OP_CHECKSIG) {
-        memcpy(&hash, &script[3], 20);
+        memcpy(hash.begin(), &script[3], 20);
```
prasannavl, line 25: Can we instead do `CKeyID.GetRaw` that points to the raw hash and just use the address of that? 

Since memcpy shouldn't be relying on addresses returned by iterators. Iterator address can be different from the raw hashes.

prasannavl: Sure, I hear that - however `begin`, `end` is usually associated with iter interfaces and hence the nitpick to avoid a misrepresented convention.

src/outputtype.cpp
```diff
@@ -63,21 +63,34 @@ CTxDestination GetDestinationForKey(const CPubKey& key, OutputType type)
...
+        CTxDestination uncompressedLegacy = PKHash(pubkeyCopy);
+        return std::vector<CTxDestination>{std::move(keyid), std::move(p2sh), std::move(segwit), std::move(eth), std::move(uncompressedLegacy)};
     } else {
+        pubkeyCopy.Compress();
         CTxDestination eth = WitnessV16EthHash(key);
-        return std::vector<CTxDestination>{std::move(keyid), std::move(eth)};
+        CTxDestination segwit = WitnessV0KeyHash(pubkeyCopy);
+        CTxDestination p2sh = ScriptHash(GetScriptForDestination(segwit));
+        return std::vector<CTxDestination>{std::move(keyid), std::move(p2sh), std::move(segwit), std::move(eth)};
     }
```
prasannavl, line 94: Can't this be:

```
if (key.IsCompressed()) {
      pubkeyCopy.Decompress();
}
```

And just not un-equivocally use uncompressed for legacy and compressed for eth and bech32 without the branching?

src/script/signingprovider.cpp
```diff
@@ -106,28 +111,33 @@ bool FillableSigningProvider::GetPubKey(const CKeyID &address, CPubKey &vchPubKe
...
 {
     LOCK(cs_KeyStore);
+
     mapKeys[pubkey.GetID()] = key;
-    if (compressedPubKey.IsValid()) {
-        mapKeys[pubkey.GetEthID()] = key;
-        mapEthKeys[pubkey.GetEthID()] = key;
-        mapKeys[compressedPubKey.GetID()] = key;
+
+    auto pubkeyCopy = pubkey;
```
prasannavl, line 129: Any specific reason to not just unconditionally store all 3 for easier maintainability?

Bushstar: We do store all three. Check the preceding line...

`mapKeys[pubkey.GetID()] = key;`

Bushstar: We then store whatever two was not already added below.

src/outputtype.cpp
Bushstar: Sure. Will update and use everywhere.

prasannavl (dismissed):

prasannavl:
Have to resolve test failures, still before merging. Shaping up great!

prasannavl (dismissed):

prasannavl (approved):

PR #2170, opened 2023-07-11
make.sh: Install Python dependencies in venv

shohamc1:


make.sh
```diff
@@ -987,6 +1010,22 @@ _ensure_enter_dir() {
     mkdir -p "${dir}" && pushd "${dir}" > /dev/null
 }
 
+_ensure_venv_active() {
```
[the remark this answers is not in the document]
prasannavl: Run the actions: activate, deactive
Ensure: do if needed, but ensures the result is a predictable env post run.

make.sh
```diff
@@ -347,10 +347,20 @@ test() {
     _fold_end
 
     _fold_start "functional-tests"
+
+    # enter python virtualenv
+    local build_dir="${BUILD_TARGET_DIR}"
```
prasannavl, line 352: Hmm. Does the python libs have to be host specific (build_target_dir), or host agnostic (build_dir)? (though both point to the same dir, that's the internal context)? 

Also, this currently couples pyenv into build_dir. How about instead, use a var in setup_vars `PYENV=${BUILD_DIR}/.pyenv`

This way it's configurable. And use that var for all py activations without depending on build_dir, so can point to any env without having to change build_dir.

shohamc1: Addressed in [161ea46](https://github.com/DeFiCh/ain/pull/2170/commits/161ea46490e7369736f4c93d0c81ce4fb438bfae)

make.sh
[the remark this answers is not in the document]
shohamc1: Addressed in [161ea46](https://github.com/DeFiCh/ain/pull/2170/commits/161ea46490e7369736f4c93d0c81ce4fb438bfae)

make.sh
```diff
@@ -542,9 +550,21 @@ pkg_install_rust() {
 }
 
 pkg_install_web3_deps() {
-    _fold_start "pkg-install-solc"
+    _fold_start "pkg-install-web3-deps"
+    # create and enter virtual environment
+    local python_venv_dir="${PYTHON_VENV_DIR}"
+    _ensure_enter_dir "${python_venv_dir}"
+    python3 -m venv .
```
prasannavl, line 557: This shouldn't be needed. Can be refined further, and just do the venv creation inside activate.

make.sh
```diff
@@ -542,9 +550,21 @@ pkg_install_rust() {
...
+
+    _activate_venv
```
prasannavl, line 559: Prefix with py to to ensure global function space is clearly namespaced

prasannavl (dismissed):

make.sh
```diff
@@ -33,6 +33,7 @@ setup_vars() {
     BUILD_TARGET_DIR="${BUILD_DIR}"
     BUILD_DEPENDS_DIR=${BUILD_DEPENDS_DIR:-"${BUILD_DIR}/depends"}
     BUILD_DEPENDS_DIR="$(_canonicalize "$BUILD_DEPENDS_DIR")"
+    PYTHON_VENV_DIR=${PYTHON_VENV_DIR:-"${BUILD_DIR}/penv"}
```
prasannavl, line 36:
```suggestion
    PYTHON_VENV_DIR=${PYTHON_VENV_DIR:-"${BUILD_DIR}/pyenv"}
```

prasannavl (approved):

PR #2175, opened 2023-07-13
Feature gate transferdomain.

Bushstar:


src/masternodes/mn_checks.cpp
```diff
@@ -4069,14 +4089,18 @@ Res ValidateTransferDomain(const CTransaction &tx,
...
+
     if (height >= static_cast<uint32_t>(consensus.ChangiIntermediateHeight4)) {
         if (obj.transfers.size() < 1) {
             return DeFiErrors::TransferDomainInvalid();
         }
     }
 
     for (const auto &[src, dst] : obj.transfers) {
-        auto res = ValidateTransferDomainEdge(tx, height, coins, consensus, src, dst);
+        auto res = ValidateTransferDomainEdge(tx, mnview, height, coins, consensus, src, dst);
```
prasannavl, line 4103: Don't need to pass mnview. Just gather all the data in one go, say `TransferDomainLiveConfig` and pass that around instead. Keeps it cleaner, better testable.

prasannavl (approved):

PR #2191, opened 2023-07-18
Fixes to EVM functional tests

sieniven:
## Summary

- Add additional test pipeline to test for burnt fees from block base gas fees to DFI burn address on accounts layer.
- Fixes to all broken EVM functional test pipelines from fee changes in validation pipeline from beta5 fixes to gas prices
- Gas limits in eth txs require at least MIN_GAS_PER_TX value (21,000)
- Gas prices in eth txs require at least initial block base fees (10gwei)

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


test/functional/feature_evm.py
```diff
@@ -135,7 +135,16 @@ def run_test(self):
         assert_equal(int(self.nodes[0].eth_getBalance(eth_address)[2:], 16), 10000000000000000000)
         assert_equal(int(self.nodes[0].eth_getBalance(to_address)[2:], 16), 0)
 
-        # Get miner DFI balance before transaction
+        # Send tokens to burn address
+        burn_address = "mfburnZSAM7Gs1hpDeNaMotJXSGA7edosG"
+        self.nodes[0].importprivkey("93ViFmLeJVgKSPxWGQHmSdT5RbeGDtGW4bsiwQM2qnQyucChMqQ")
```
shohamc1, line 140: What is this key for?

Bushstar:
feature_setgov.py is failing as well.

`AssertionError: Expected substring (Unrecognised type argument provided, valid types are: consortium, gov, locks, oracles, params, poolpairs, token,) not found:Unrecognised type argument provided, valid types are: consortium, evm, gov, locks, oracles, params, poolpairs, token, transferdomain, vaults,`

test/functional/feature_evm.py
Bushstar: It is the burn address private key. I think this is needed to allow accounttoaccount as it requires the destination to be owned by the wallet. Really need to add a bool to skip that requirement.

sieniven: Resolving.

Bushstar (approved):

PR #2205, opened 2023-07-21
Refactor to use typed amounts and checked numerics

sieniven:
## Summary

- Refactor of amounts (wei, gwei and satoshis) with rust newtype pattern.
- Refactor fallible fee calculation methods to guard against value overflows with new test case for this in feature_evm_fees.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-evm/src/evm.rs
```diff
@@ -153,6 +153,7 @@ impl EVMServices {
                         }
                     }
 
+                    let prepay_gas = calculate_prepay_gas(&signed_tx).unwrap();
```
shohamc1, line 156: Need to propagate the error instead of unwrapping.

shohamc1 (changes requested):

prasannavl (approved):

PR #2206, opened 2023-07-21
Remove Changi intermediate forks

Bushstar:


src/masternodes/mn_checks.cpp
```diff
@@ -4498,7 +4476,12 @@ ResVal<uint256> ApplyAnchorRewardTxPlus(CCustomCSView &mnview,
             cbValues.begin()->second,
             anchorReward);
 
-    CTxDestination destination = FromOrDefaultKeyIDToDestination(finMsg.rewardKeyType, finMsg.rewardKeyID, KeyType::MNRewardKeyType);
+    CTxDestination destination;
+    if (height < consensusParams.NextNetworkUpgradeHeight) {
+        destination = FromOrDefaultKeyIDToDestination(finMsg.rewardKeyType, finMsg.rewardKeyID, KeyType::MNRewardKeyType);
+    } else {
+        destination = FromOrDefaultKeyIDToDestination(finMsg.rewardKeyType, finMsg.rewardKeyID, KeyType::MNRewardKeyType);
+    }
```
kuegi, line 4484: there is no difference between the branches. typo?

Bushstar: Good spot. Just added to avoid an unintentional fork and just restored the same behaviour.

PR #2209, opened 2023-07-21
Reject EVM TXs with zero fees. Check total EVM fees paid.

Bushstar:


src/masternodes/validation.cpp
```diff
@@ -2444,6 +2450,7 @@ static void ProcessEVMQueue(const CBlock &block, const CBlockIndex *pindex, CCus
     if (pindex->nHeight >= chainparams.GetConsensus().ChangiIntermediateHeight4) {
         cache.AddBalance(Params().GetConsensus().burnAddress, {DCT_ID{}, static_cast<CAmount>(blockResult.total_burnt_fees)});
         cache.AddBalance(minerAddress, {DCT_ID{}, static_cast<CAmount>(blockResult.total_priority_fees)});
+        assert(blockResult.total_burnt_fees + blockResult.total_priority_fees <= totalEvmFees);
```
sieniven, line 2453: We should do a stricter check, since the total burnt fees and total priority fees should tally to the total evm fees of that evm block.

src/masternodes/mn_checks.cpp
```diff
@@ -3939,15 +3942,21 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
+                return Res::Err("evm tx does not pay a fee");
+            }
 
             evm_try_queue_tx(result, evmContext, HexStr(obj.evmTx), tx.GetHash().GetByteArray(), prevalidateResults.gas_used);
             if (!result.ok) {
                 LogPrintf("[evm_try_queue_tx] failed, reason : %s\n", result.reason);
                 return Res::Err("evm tx failed to queue %s\n", result.reason);
             }
+
+            totalEvmFees += prevalidateResults.tx_fees;
```
sieniven, line 3956: This is incorrect. The fees returned here isnt actually the gas fees that the sender will be paying, but the prepay gas fees. Will make a quick PR to make the FFI results clearer.

sieniven (changes requested):

src/masternodes/mn_checks.cpp
sieniven: The check will still pass if we are intending to check that the total prepay gas fees of the block is less than or equal to the total block fees actually paid (burnt + priority fees).

Bushstar: The miner code also uses tx_fees from the evm_try_validate_raw_tx result to check if one TX pays more than another. Sounds like that will also be the incorrect behavior. Can you also refactor to include the total fees for us here and in the miner code?

sieniven: `const auto& gasFees = evmTXFees.at(evmKey);
                        if (txResult.tx_fees > gasFees) {`

If you are referring to tx fees being added to the ordered map, then this pipeline is correct. On the ethereum side, txs with the same nonce will be replaced in the mempool with a new tx with higher prepay gas fees.

sieniven: the tx_fees used in this PR are all incorrect though. evm_try_validate_tx and evm_try_prevalidate_tx do not return the actual gas fees that will be used in the tx.

Bushstar: Thanks for confirming that things are correct on the miner side.

src/masternodes/validation.cpp
sieniven: Resolved in 74e07a2

src/masternodes/mn_checks.cpp
sieniven: Resolved in 74e07a2

sieniven:
The implementation to discard evm txs that have 0 priority fee is discontinued - evm tx with no priority fee is valid in ethereum, but miner in the future should prioritize mining evm transactions, issue opened in: https://github.com/DeFiCh/ain/issues/2217

sieniven:
EVM fee check is also implemented to be stricter - total_burnt_fees + total_priority_fees == total_fees

src/miner.cpp
```diff
@@ -1146,7 +1146,7 @@ namespace pos {
         //
         auto pblocktemplate = BlockAssembler(chainparams).CreateNewBlock(scriptPubKey, blockTime);
         if (!pblocktemplate) {
-            throw std::runtime_error("Error in WalletStaker: Keypool ran out, please call keypoolrefill before restarting the staking thread");
+            return Status::stakeWaiting;
```
prasannavl, line 1149:
```suggestion
            LogPrintf("Error in WalletStaker: Keypool ran out, please call keypoolrefill before restarting the staking thread\n");
            return Status::stakeWaiting;
```

prasannavl (approved):

sieniven (approved):

PR #2210, opened 2023-07-21
evm: Intrinsic gas check

sieniven:
## Summary

- Implementation of intrinsic gas check to validate evm transaction gas limits, aligning to ethereum's intrinsic gas check pipeline using SputnikVM's Gasometer library.
- Intrinsic gas calculation for the amount of gas that the transaction uses before tx is called.
- Calculations include a constant tx fee for call / creation txs, inclusive of a fee for every byte of data in the tx.
- Data parameters are aligned to ethereum Shanghai fork parameters: 4 gas for zero byte, 16 gas for non-zero byte data
- Access list gas calculation also includes the cost of access list address size and the total access list storage key size.
- Access list parameters are aligned to the ethereum Shanghai fork parameters at 2400 gas for every access list addresses and 1900 gas for every storage key.
- For the full parameters configuration, refer to: https://github.com/rust-blockchain/evm/blob/master/runtime/src/lib.rs
- Note that the changes are non-backwards compatible.
- The full intrinsic gas check algorithm from Ethereum yellow paper:
Screenshot 2023-07-21 at 12 32 12 PM

## References
1. IntrinsicGas calculation from ethereum (go-ethereum::core::state_transition.go::IntrinsicGas) - https://github.com/ethereum/go-ethereum/blob/7f756dc1185d7f1eeeacb1d12341606b7135f9ea/core/state_transition.go#L68C1-L116C2
2. Word size calculation (go-ethereum::core::state_transition.go::toWordSize) - https://github.com/ethereum/go-ethereum/blob/7f756dc1185d7f1eeeacb1d12341606b7135f9ea/core/state_transition.go#L119-L125
3. Parameters (go-ethereum::params/protocol_params.go) - https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go

## Housekeeping
1. Add comments for validate_raw_tx validation pipeline
2. Fixes to maxFeePerGas function call in signed txs - maximum fee for legacy txs should be gasPrice specified.

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


lib/ain-evm/src/gas.rs
```diff
@@ -0,0 +1,38 @@
+use crate::transaction::SignedTx;
+use ethereum::TransactionAction;
+use evm::Config;
+use evm_gasometer::{call_transaction_cost, create_transaction_cost, Gasometer, TransactionCost};
```
shohamc1, line 4: Do not need to separately include the gasometer dependency, evm already reexports gasometer.

```suggestion
use evm::gasometer::{call_transaction_cost, create_transaction_cost, Gasometer, TransactionCost};
```

lib/ain-evm/src/core.rs
```diff
@@ -224,11 +225,9 @@ impl EVMCoreService {
...
             }
 
-            if gas_limit < MIN_GAS_PER_TX {
-                debug!("[validate_raw_tx] gas limit is below the minimum gas per tx");
-                return Err(anyhow!("gas limit is below the minimum gas per tx").into());
-            }
-        } else if balance < MIN_GAS_PER_TX || balance < prepay_fee {
+            // Validate tx gas limit with intrinsic gas
+            check_tx_intrinsic_gas(&signed_tx)?;
+        } else if balance < MIN_GAS_PER_TX.into() || balance < prepay_fee {
```
shohamc1, line 230: I think this check is not required since it is a subset of the intrinsic gas check.

The call cost is added [here](https://docs.rs/evm-gasometer/0.39.0/src/evm_gasometer/lib.rs.html#230) and the gas check is done [here](https://docs.rs/evm-gasometer/0.39.0/src/evm_gasometer/lib.rs.html#260).

sieniven: Thats incorrect. I think you are confusing gas and fees. This validation checks that the account sender has enough dfi in the account to pay for the prepay gas fees.

lib/ain-evm/src/gas.rs
sieniven: Thanks for this, resolved in commit d427fb922

lib/ain-evm/src/core.rs
shohamc1: The unit for `MIN_GAS_PER_TX` is gas and the unit for `balance` is wei from what I see. I don't think this comparison makes sense, IMO only the second check should be in there.

sieniven: that is erroneous pipeline forguarded with changiintermediate, which will be removed in beta6.

sieniven: `
// Changi intermediate constant

pub const MIN_GAS_PER_TX: u64 = 21_000;`

Reference gas.rs::line 12-13

shohamc1: Ah makes sense in that case :disappointed:

shohamc1 (approved):

lib/ain-evm/src/core.rs
sieniven: Resolving.

PR #2213, opened 2023-07-24
evm: Update state root on every block

shohamc1:


test/functional/feature_evm_state_root_change.py
```diff
@@ -0,0 +1,37 @@
...
+        w3 = Web3(Web3.HTTPProvider(self.nodes[0].get_evm_rpc()))
+
+        NUM_BLOCKS = 5
+        state_roots = set()
+        for i in range(NUM_BLOCKS):
+            node.generate(1)
+            block = w3.eth.get_block('latest')
+            state_roots.add(Web3.to_hex(block["stateRoot"]))
+
+        assert_equal(len(state_roots), NUM_BLOCKS)
```
canonbrother, line 33: hmm.. length can't assert the state_root is different in every block...
need `assert_not_equal` to support this test.. eg: 
```
b0 = w3.eth.get_block('latest')
node.generate(1)
b1 = w3.eth.get_block('latest')
assert_not_equal(b0['state_root'], b1['state_root'])
```

shohamc1: `state_roots` is a set so it removes duplicate entries during `add`. So if any two blocks have the same hash, the set would be one element shorter than expected and the test would fail.

Running this test on a master build fails
```
2023-07-24T07:19:01.369000Z TestFramework (ERROR): Assertion failed
Traceback (most recent call last):
  File "/home/shohamc1/projects/ain/test/functional/test_framework/test_framework.py", line 217, in main
    self.run_test()
  File "/home/shohamc1/projects/ain/test/functional/feature_evm_state_root_change.py", line 33, in run_test
    assert_equal(len(state_roots), NUM_BLOCKS)
  File "/home/shohamc1/projects/ain/test/functional/test_framework/util.py", line 41, in assert_equal
    raise AssertionError("not(%s)" % " == ".join(str(arg) for arg in (thing1, thing2) + args))
AssertionError: not(1 == 5)
```

lib/ain-evm/src/evm.rs
```diff
@@ -58,32 +67,34 @@ impl EVMServices {
...
                 receipt: ReceiptService::new(Arc::clone(&storage)),
                 logs: LogService::new(Arc::clone(&storage)),
                 filters: FilterService::new(),
                 storage,
-            })
-        }
+            }
+        };
+
+        Ok(services)
```
sieniven, line 97: Is there a rationale for changing the instantiation function to this?

lib/ain-evm/src/evm.rs
```diff
@@ -143,6 +154,22 @@ impl EVMServices {
 
         let mut executor = AinExecutor::new(&mut backend);
 
+        if current_block_number == U256::zero() {
+            let CounterContractInfo {
+                address,
+                storage,
+                bytecode,
+            } = EVMServices::counter_contract().unwrap();
```
sieniven, line 162: We should use proper error handling in ain-contracts lib functions and propagate them properly here.

lib/ain-evm/src/evm.rs
```diff
@@ -143,6 +154,22 @@ impl EVMServices {
...
+            executor
+                .deploy_contract(address, bytecode, storage)
+                .unwrap();
+        } else {
+            let CounterContractInfo {
+                address, storage, ..
+            } = EVMServices::counter_contract().unwrap();
```
sieniven, line 169: Likewise over here. If contract deployment was unsuccessful with deploy_contract, error should be propagated and node state (if shutting down) should transit more elegantly.

lib/ain-evm/src/evm.rs
```diff
@@ -143,6 +154,22 @@ impl EVMServices {
...
+            executor.update_storage(address, storage).unwrap();
```
sieniven, line 170: Error should be propagated.

lib/ain-evm/src/evm.rs
```diff
@@ -326,4 +353,33 @@ impl EVMServices {
...
+            .evm
+            .block
+            .get_latest_block_hash_and_number()
+            .unwrap_or_default();
+        let count = U256::from(
+            SERVICES
+                .evm
+                .core
+                .get_storage_at(address, U256::one(), latest_block_number)?
+                .unwrap_or_default()
```
sieniven, line 370: Should we be propagating this error?

lib/ain-contracts/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    let length = input.len();
+
+    let mut storage_value = H256::default();
+    storage_value.0[31] = (length * 2) as u8;
+    storage_value.0[..length].copy_from_slice(input.as_bytes());
+
+    storage_value
+}
+
+pub fn get_bytecode(input: &str) -> Result<Vec<u8>, Box<dyn Error>> {
```
sieniven, line 25: Might want to add in rust debug logs and errors to propagate them downstream.

lib/ain-contracts/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    let bytecode_json: serde_json::Value = serde_json::from_str(input)?;
+    let bytecode_raw = bytecode_json["object"]
+        .as_str()
+        .ok_or_else(|| anyhow!("Bytecode object not available".to_string()))?;
+
+    Ok(hex::decode(&bytecode_raw[2..]).map_err(|e| anyhow!(e.to_string()))?)
+}
+
+pub fn get_counter_bytecode() -> Result<Vec<u8>, Box<dyn Error>> {
```
sieniven, line 34: Might want to add in rust debug logs and errors to propagate them downstream.

sieniven (changes requested):

lib/ain-evm/src/evm.rs
shohamc1: Old change I forgot to undo, will revert this.

lib/ain-contracts/src/lib.rs
shohamc1: The errors are already being propagated here it seems. Not sure what you mean about the logs :/

make.sh
```diff
@@ -527,6 +527,13 @@ pkg_install_rust() {
     _fold_end
 }
 
+pkg_install_solc_linux() {
```
sieniven, line 530: rename to pkg_install_solc() instead since our default pkg manager is linux-based os.

lib/ain-evm/src/evm.rs
shohamc1: This is not an error. In the initial state where the storage is unset for the address, we want to get the default value instead of panicking.

make.sh
```diff
@@ -949,6 +956,7 @@ ci_setup_deps() {
     DEBIAN_FRONTEND=noninteractive pkg_setup_locale
     DEBIAN_FRONTEND=noninteractive pkg_install_llvm
     DEBIAN_FRONTEND=noninteractive pkg_install_rust
+    DEBIAN_FRONTEND=noninteractive pkg_install_solc_linux
```
sieniven, line 959: CI pipeline for installing packages should be for setting up CI runner with necessary deps to run the build workflow. Target architectures differ for the multi-arch build pipeline for our supported architectures, so target specific deps should not be here. Target specific dependencies should be shifted inside build-depends pipeline, which will install the correct target dependencies based on the specified target architecture.

sieniven (changes requested):

lib/ain-evm/src/evm.rs
```diff
@@ -143,6 +152,21 @@ impl EVMServices {
 
         let mut executor = AinExecutor::new(&mut backend);
 
+        if current_block_number == U256::zero() {
```
canonbrother, line 155: some note here.. it won't work until we remove the genesis restriction on regtest only and dependence at `ethStartState` flag

shohamc1: Would it work if we do this on block 1?

lib/ain-contracts/src/lib.rs
shohamc1: The code already does this.

lib/ain-evm/src/evm.rs
canonbrother: yea block 1 will work.. can add a note above the line? 
```
TODO(): change back to `U256::zero()` after https://github.com/DeFiCh/ain/issues/2216
```

canonbrother: Ignore the thread.. some memory conflicts..
the block 0 exists before feature/evm enable
It's deployed when compiling the defid

lib/ain-evm/src/backend.rs
```diff
@@ -187,6 +187,68 @@ impl EVMBackend {
...
+            .map_err(|e| EVMBackendError::TrieRestoreFailed(e.to_string()))?;
+
+        Ok(U256::from(
+            state
+                .get(storage_index.clone().as_bytes())
+                .unwrap_or_default()
+                .unwrap_or_default()
+                .as_slice(),
+        ))
+    }
```
Jouzo, line 216: Unused in this PR.

lib/ain-evm/src/backend.rs
```diff
@@ -187,6 +187,68 @@ impl EVMBackend {
...
             .unwrap_or_default()
     }
+
+    pub fn get_account_storage(&self, contract: H160, storage_index: H256) -> Result<U256> {
+        let account = self.get_account(&contract).unwrap_or(Account {
+            nonce: U256::zero(),
+            balance: U256::zero(),
+            storage_root: H256::zero(),
+            code_hash: H256::zero(),
+        });
```
Jouzo, line 197: Use let else { return U256::default() } to return early

lib/ain-evm/src/backend.rs
```diff
@@ -187,6 +187,68 @@ impl EVMBackend {
...
+                balance: U256::zero(),
+                nonce: U256::zero(),
+            },
+            None,
+            storage,
+            false,
+        )?;
+
+        Ok(())
+    }
```
Jouzo, line 251: Wouldn't it make sense for these two to accept basic as extra param ? Maybe as Option

lib/ain-evm/src/evm.rs
```diff
@@ -326,4 +350,33 @@ impl EVMServices {
 
         Ok(())
     }
+
+    pub fn counter_contract() -> Result<CounterContractInfo, Box<dyn Error>> {
```
Jouzo, line 354: Needs comment to explicit that it returns the contract with the count already incremented

test/functional/feature_evm_state_root_change.py
```diff
@@ -0,0 +1,37 @@
+from test_framework.test_framework import DefiTestFramework
```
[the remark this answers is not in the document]
shohamc1: Fixed: 034ff1696b2676ffec3eec04eabbaae524bd6f69

lib/ain-evm/src/evm.rs
shohamc1: Fixed: 49ec7c6bd99f38f299817819353c1a652bdc55a4

lib/ain-evm/src/backend.rs
shohamc1: Fixed: 49ec7c6bd99f38f299817819353c1a652bdc55a4

lib/ain-evm/src/backend.rs
shohamc1: Fixed: 49ec7c6bd99f38f299817819353c1a652bdc55a4

lib/ain-evm/src/backend.rs
shohamc1: Hmm the nonce isn't really used in contracts (it only increases if the contract deploys another contract). It's also unlikely that any of the state injection contracts would need any EVM balance for operations. I think we can add it later if a need arises.

make.sh
shohamc1: Fixed: 45f8c29ce87c91e714ab0edb9f654a265531868b

Jouzo (approved):

sieniven (approved):

PR #2219, opened 2023-07-24
Add base fee to block header

Jouzo:


sieniven (dismissed):

shohamc1 (commented):
We can probably remove all usages of `base_fee_map`

https://github.com/DeFiCh/ain/blob/d34e18981d76417dddc7d5075ba2147ec5e88516/lib/ain-evm/src/storage/data_handler.rs#L53

prasannavl (approved):

PR #2220, opened 2023-07-25
Refactor miner: Isolate EVM processing

prasannavl:
## Summary

- Cleanup and isolate the EVM process handling in the miner.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


src/miner.cpp
```diff
@@ -649,6 +709,93 @@ void BlockAssembler::SortForBlock(const CTxMemPool::setEntries& package, std::ve
...
+            checkedDfTxHashSet.erase(addrTxs[addrKey.nonce]->GetTx().GetHash());
+            addrTxs[addrKey.nonce] = txIter;
+            auto count{addrKey.nonce};
+            for (const auto& [nonce, entry] : addrTxs) {
+                checkedDfTxHashSet.erase(entry->GetTx().GetHash());
+                replaceByFee.emplace(count, entry);
+                ++count;
+            }
+            evmAddressTxsMap.erase(addrKey.address);
+            evm_remove_txs_by_sender(evmQueueId, addrKey.address);
```
sieniven, line 778: Same thing as caught in PR #2220. When we call evm_remove_txs_by_sender, the current pipeline is such that we are actually removing all transactions from that specific sender in the txqueue. This actually means that if for example we send 5 evm txs from 0 to 5 nonces, and then an evm tx of nonce 3 with higher prepay gas is submitted which replaces the evm tx nonce 3 - we are wiping all txs from 0 to 5 here. This test isnt actually covered in feature_evm, since the nonce replacement is nicely replacing nonce 0 in the test case which misses this specific edge case.

src/miner.cpp
```diff
@@ -649,6 +709,93 @@ void BlockAssembler::SortForBlock(const CTxMemPool::setEntries& package, std::ve
...
+                    ++it;
+                }
+            }
+            // Buggy code to fix below:
+            checkedDfTxHashSet.erase(addrTxs[addrKey.nonce]->GetTx().GetHash());
+            addrTxs[addrKey.nonce] = txIter;
+            auto count{addrKey.nonce};
+            for (const auto& [nonce, entry] : addrTxs) {
+                checkedDfTxHashSet.erase(entry->GetTx().GetHash());
+                replaceByFee.emplace(count, entry);
```
sieniven, line 774: Similar to PR #2220 - is there a reason to be using count instead of directly using the nonce matching the evm tx when storing descendant evm txs inside this index?

src/miner.cpp
sieniven: Resolving as the current pipeline is correct. We are wiping all of our indexes whenever we replace the evm tx (both ancestors and descendants), and adding them into the replacebyfee index.

src/miner.cpp
Bushstar: We did not have the nonce before and they should have been in nonce order. Will update this.

src/miner.cpp
```diff
@@ -69,6 +69,40 @@ struct XVM {
...
+    uint64_t nonce;
+
+    bool operator<(const EvmAddressWithNonce& item) const {
+        return std::tie(address, nonce) < std::tie(item.address, item.nonce);
+    }
+};
+
+struct EvmPackageContext {
+    // Used to track EVM TXs by sender and nonce.
+    std::map<EvmAddressWithNonce, uint64_t> feeMap;
```
Jouzo, line 85:
```suggestion
    // Used to track EVM TX fee by sender and nonce.
    std::map<EvmAddressWithNonce, uint64_t> feeMap;
```

Jouzo (dismissed):

Jouzo (dismissed):

Bushstar (approved):

Jouzo (approved):

PR #2221, opened 2023-07-25
Revert buggy miner code and fix segfault

Bushstar:


lib/ain-evm/src/txqueue.rs
```diff
@@ -263,9 +274,20 @@ impl TransactionQueue {
...
+            };
+            if tx_sender == sender {
+                data.total_fees -= item.tx_fee;
+                data.total_gas_used -= item.gas_used;
+            }
+        }
+        data.transactions.retain(|item| {
+            let tx_sender = match &item.queue_tx {
                 QueueTx::SignedTx(tx) => tx.sender,
                 QueueTx::BridgeTx(tx) => tx.sender(),
```
Jouzo, line 292: We could do the fee/gas reduction in the retain loop. Shouldn't be required to clone the vec and iterate twice

src/miner.cpp
```diff
@@ -865,44 +872,37 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
...
-                            addrTxs[addrKey.nonce] = sortedEntries[i];
-                            auto count{addrKey.nonce};
-                            for (const auto& [nonce, entry] : addrTxs) {
+                            checkedTX.erase(evmTXs[txResult.sender][txResult.nonce]->GetTx().GetHash());
+                            evmTXs[txResult.sender][txResult.nonce] = sortedEntries[i];
+                            auto count{txResult.nonce};
+                            for (const auto& [nonce, entry] : evmTXs[txResult.sender]) {
                                 inBlock.erase(entry);
                                 checkedTX.erase(entry->GetTx().GetHash());
                                 replaceByFee.emplace(count, entry);
```
sieniven, line 901: Is there a reason why we do not use the nonce obtained from iterating through (nonce, entry) of the evmTXs map over here?

lib/ain-evm/src/txqueue.rs
sieniven: Resolved in 0d3f42934

src/miner.cpp
```diff
@@ -571,56 +554,93 @@ int BlockAssembler::UpdatePackagesForAdded(const CTxMemPool::setEntries& already
...
+            continue;
+        }
+
         for (size_t i = 0; i < pblock->vtx.size();  ++i) {
-            auto current = pblock->vtx[i];
-            if (current && current->GetHash() == tx.GetHash()) {
+            if (pblock->vtx[i] && pblock->vtx[i]->GetHash() == iter->GetTx().GetHash()) {
                 pblock->vtx.erase(pblock->vtx.begin() + i);
-                nBlockWeight -= iter->GetTxWeight();
-                nBlockSigOpsCost -= iter->GetSigOpCost();
```
sieniven, line 582: We are not removing sigopscost when removing failed transactions

src/miner.cpp
```diff
@@ -865,44 +872,37 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
...
+                            for (const auto& [nonce, entry] : evmTXs[txResult.sender]) {
                                 inBlock.erase(entry);
                                 checkedTX.erase(entry->GetTx().GetHash());
                                 replaceByFee.emplace(count, entry);
                                 ++count;
                             }
-                            evmAddressTxsMap.erase(addrKey.address);
-                            evm_remove_txs_by_sender(evmQueueId, addrKey.address);
+                            evmTXs.erase(txResult.sender);
+                            evm_remove_txs_by_sender(evmQueueId, txResult.sender);
```
sieniven, line 905: When we call evm_remove_txs_by_sender, the current pipeline is such that we are actually removing all transactions from that specific sender in the txqueue. This actually means that if for example we send 5 evm txs from 0 to 5 nonces, and then an evm tx of nonce 3 with higher prepay gas is submitted which replaces the evm tx nonce 3 - we are wiping all txs from 0 to 5 here. This test isnt actually covered in feature_evm, since the nonce replacement is nicely replacing nonce 0 in the test case which misses this specific edge case.

sieniven: Resolving as the current pipeline is correct. We are wiping all of our indexes whenever we replace the evm tx (both ancestors and descendants), and adding them into the replacebyfee index.

src/miner.cpp
prasannavl: Indeed. #2220 is the attempt to fix all that correctly. This PR reverts and tries to be as close as possible to the original to get moving for now. 

(the sigops is OK since EVMTx do not hold sigscript)

src/miner.cpp
prasannavl: If we're at this point, it reflects that we're in the correct nonce (since validate would not have succeeded otherwise). So, all items in the queue needs to be ordered after.

prasannavl: Not the most ideal way, but works for the moment. We'll likely again restructure this in #2220 or later

Jouzo (approved):

PR #2227, opened 2023-07-25
Wipe evm folder on reindex

Jouzo:
## Summary

- Wipe evm folder on reindex
- Introduces IS_INITIALIZED bool. This is done in case an error occurs before `init_services` is called to prevent `stop_network_services` and `stop_services` to initialize SERVICES.


lib/ain-grpc/src/lib.rs
```diff
@@ -57,12 +56,14 @@ pub fn init_logging() {
 
 pub fn init_services() {
     info!("Init rs services");
+    IS_INITIALIZED.store(true, Ordering::Relaxed);
     let _ = &*SERVICES;
```
prasannavl, line 60:
```suggestion
    let _ = &*SERVICES;
    IS_INITIALIZED.store(true, Ordering::Relaxed);
```

prasannavl: Post services, so we know all services are initialized by this point.

lib/ain-grpc/src/lib.rs
```diff
@@ -57,12 +56,14 @@ pub fn init_logging() {
...
 }
 
 pub fn init_network_services(json_addr: &str, grpc_addr: &str) -> Result<()> {
     init_network_json_rpc_service(&SERVICES, json_addr)?;
     init_network_grpc_service(&SERVICES, grpc_addr)?;
+    IS_INITIALIZED.store(true, Ordering::Relaxed);
```
prasannavl, line 66: Perhaps just init in one place, either here or after services // or perhaps granular flags.

Bushstar (approved):

lib/ain-grpc/src/lib.rs
Jouzo: Wanted to cover all usage of SERVICES that would initialize it just to be safe. Another option was to assert it's initialization before calling `the init_network_*`

prasannavl (approved):

PR #2228, opened 2023-07-26
make.sh: Install solc binaries

shohamc1:


sieniven:
To do is to shift installing of solc dependency into build-depends.

sieniven (dismissed):

prasannavl:
Closing in favor of #2237

PR #2183, opened 2023-07-17
Add transfer domain config

Bushstar:

PR #2232, opened 2023-07-26
Split addressmap from vmmap into wallet RPCs

Mixa84:
## Summary

- Move address map out of `vmmap` RPC into it's own. 
- Notes to early users: Please do not rely on this output format yet (More info: https://github.com/DeFiCh/ain/pull/2232#issuecomment-1651946384)

## RPCs

- Changed `vmmap`:
  - Removed address types that doesn't belong here.
  - All vmmap mapping should now be universal.
-  Added `addressmap`: 
   - Separate RPC for wallet address format maps since addresses technically aren't a part of the VM, but rather the wallet.
   - Addresses mapping are only for items in the wallet. 

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


prasannavl (approved):

prasannavl:
- We should add more changes to this later, so it's a proper json response, with all address types after https://github.com/DeFiCh/ain/pull/2164 is merged
- Should use the accepted types from https://github.com/DeFiCh/ain/pull/2183 once it's merged.

Jouzo (approved):

PR #2237, opened 2023-07-27
Fix the toolchain madness

prasannavl:
## Summary

- The largest headaches in the toolchain is often due to the way depends overrides the toolchains
  - For darwin cross compiles it hijacks the compilers entirely
  - For the others, it reuses the env compilers
- Rust and cargo ecosystem will attempt to compile it's native libraries on it's own which will result in calling C/XX compilers resulting in nested compilation env that are highly inconsistent since depends simply overrides it.
- This is further made challenging since some Rust libs will require host builds (eg: build-deps), while some require target builds.
- This retains the overrides from depends and passes it along cleanly in an attempt to resolve this madness. 

### Additional

- ./make.sh respects CC and CXX vars and will not override them
- `MAKE_USE_CLANG` is auto set based on basic heuristics and can now be easily switched off 
- solc:
  - Adds solc to depends system
  - Exports SOLC_PATH var for solc bin path
  - *References*:
    - https://docs.soliditylang.org/en/latest/installing-solidity.html#static-binaries
    - Repo structure: https://github.com/ethereum/solc-bin/

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/depends/packages/solc.mk
+++ b/depends/packages/solc.mk
@@ -32,6 +32,6 @@ endef
 
 define $(package)_build_cmds
   mkdir -p $($(package)_ROOT) && \
-  chmod +x $($(package)_source) && \
-  cp $($(package)_source) $($(package)_ROOT)/$(package)
+  cp $($(package)_source) $($(package)_ROOT)/$(package) && \
+  chmod +x $($(package)_ROOT)/$(package)
 endef
--- a/depends/packages/solc.mk
+++ b/depends/packages/solc.mk
@@ -1,0 +1,45 @@
+package=solc
+$(package)_version=0.8.20
+$(package)_download_path=https://github.com/ethereum/solidity/releases/download/v$($(package)_version)/
+$(package)_file_ext=
+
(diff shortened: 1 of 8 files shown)


depends/packages/solc.mk
```diff
@@ -0,0 +1,37 @@
...
+# NOTE: solc gets invoked on the BUILD OS during compile. So, we don't
+# care about HOST OS, unlike all other dependencies.
+# That is: solc is run on the BUILD OS, not targeted to run on the
+# HOST OS.
+
+ifeq ($(build_os)-$(build_arch),linux-x86_64)
+  $(package)_file_name=solc-static-linux
+  $(package)_sha256_hash=0479d44fdf9c501c25337fdc540419f1593b884a87b47f023da4f1c700fda782
+endif
+ifeq ($(build_os),darwin)
```
prasannavl, line 14:
```suggestion

# Note: solc only provides binaries for darwin amd64, however since rosetta is
# expected to be present, this should work through emulation without additional config.
ifeq ($(build_os),darwin)
```

sieniven (approved):

PR #2248, opened 2023-07-28
feat(src/wallet): `addressmap` - handle auto and ret json

canonbrother:
## Summary

- as per title 

## RPCs

`addressmap`

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4265,11 +4265,12 @@ UniValue addressmap(const JSONRPCRequest &request) {
         },
         RPCResult{
             "{\n"
-            "    input :       \"address\",         (string) The input address to be converted\n"
-            "    type :        \"map type\"\n       (numeric) address map type indicator"
-            "    format : {\n"
-            "       bech32: \"address\"\n           (string, optional) output converted address"
-            "       erc55 : \"address\"\n           ..."
+            "    input :    \"address\",         (string) The input address to be converted\n"
+            "    type :     \"map type\"\n       (numeric) address map type indicator"
+            "    format : { \n"
+            "       bech32: \"address\"\n        (string, optional) output converted address"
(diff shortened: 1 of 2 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -4276,34 +4286,60 @@ UniValue addressmap(const JSONRPCRequest &request) {
...
                 key.Compress();
             }
-            return EncodeDestination(WitnessV0KeyHash(key));
+            format.pushKV("bech32", EncodeDestination(WitnessV0KeyHash(key)));
+            break;
         }
         default:
-            throw JSONRPCError(RPC_INVALID_REQUEST, "Invalid address type passed");
-            break;
+            if (dest.index() == WitV0KeyHashType || dest.index() == PKHashType) {
```
prasannavl, line 4322: Could you pull this logic up and  just set `AddressConversionType` when you can determine it? If not, it still does through the default case, and fails.

prasannavl: If we cannot determine, let's just fail with something like, "Unsupported type or unable to determine conversion type automatically from the input".

prasannavl:
Looks good in general. LGTM once we can refine it a bit

src/wallet/rpcwallet.cpp
canonbrother: > Could you pull this logic up and just set AddressConversionType when you can determine it

hmm.. i don't quite understand.. why do we need to determine `AddressConversionType ` by using that logic since we ady have `type` provided from client??

canonbrother: got it thanks for guidance
auto to determine type -> pass the type to switch flow will do

prasannavl (approved):

PR #2249, opened 2023-07-28
Check output matches pubkey in HasAuth

Bushstar:


src/masternodes/mn_checks.cpp
```diff
@@ -3984,15 +3984,18 @@ Res HasAuth(const CTransaction &tx, const CCoinsViewCache &coins, const CScript
...
                 if (pubkey.Decompress()) {
-                    auto script = GetScriptForDestination(WitnessV16EthHash(pubkey));
-                    if (script == auth)
+                    const auto script = GetScriptForDestination(WitnessV16EthHash(pubkey));
+                    const auto scriptOut = GetScriptForDestination(PKHash(pubkey));
+                    if (script == auth && coin.out.scriptPubKey == scriptOut)
                         return Res::Ok();
                 }
-            } else if (solution == txnouttype::TX_WITNESS_V0_KEYHASH) {
+            } else if (solution == txnouttype::TX_WITNESS_V0_KEYHASH &&
```
prasannavl, line 3992: We don't need the stack.size() == 2, since the Solver already verifies this, and we also do the reverse conversion check now

src/masternodes/mn_checks.cpp
```diff
@@ -3984,15 +3984,18 @@ Res HasAuth(const CTransaction &tx, const CCoinsViewCache &coins, const CScript
...
+                       input.scriptWitness.stack.size() == 2) {
```
prasannavl, line 3993:
```suggestion
            } else if (solution == txnouttype::TX_WITNESS_V0_KEYHASH) {
```

Jouzo (approved):

PR #2258, opened 2023-07-31
evm: Deploy DST20 for loan tokens

shohamc1:


src/masternodes/mn_checks.cpp
```diff
@@ -1069,7 +1069,7 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
 
         if (tokenId && token.IsDAT() && IsEVMEnabled(height, mnview, consensus)) {
```
sieniven, line 1070: Need to do a check for result of CreateToken to ensure that the token to be created passes validation on the DVM side first.

src/masternodes/mn_checks.cpp
```diff
@@ -2652,6 +2655,18 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
             auto attributes  = mnview.GetAttributes();
             attributes->time = time;
 
+            if (tokenId && token.IsLoanToken() && IsEVMEnabled(height, mnview, consensus)) {
```
sieniven, line 2658: Validation check for CreateToken is already done above here, do we need to include tokenId here? the ResVal value is ensured to be non-null

src/masternodes/mn_checks.cpp
sieniven: Note that this is currently on master.

Bushstar: tokenId here is checking the result of CreateToken.

lib/ain-evm/src/evm.rs
```diff
@@ -497,4 +497,37 @@ impl EVMServices {
             storage: vec![(storage_index, ain_contracts::u256_to_h256(new_balance))],
         })
     }
+
+    pub fn dst20_is_deployed(
```
shohamc1, line 501: Conveys functionality more clearly.

```suggestion
    pub fn is_dst20_queued_or_deployed(
```

test/functional/feature_dst20.py
```diff
@@ -488,7 +489,7 @@ def run_test(self):
         )
 
         # Contract ABI
-        self.abi = open("./lib/ain-contracts/dst20/output/abi.json", "r", encoding="utf8").read()
+        self.abi = open(f"{os.path.dirname(__file__)}/../../lib/ain-contracts/dst20/output/abi.json", "r", encoding="utf8").read()
```
prasannavl, line 492: ok, to avoid these, we do need to use `OUT_DIR` convention and build scripts need to use this as well.

https://doc.rust-lang.org/cargo/reference/environment-variables.html 

We already do this in other build scripts. ain-contracts will have to start using CARGO_MANIFEST_DIR for source, and OUT_DIR for target.

src/masternodes/validation.cpp
```diff
@@ -2484,9 +2484,31 @@ static Res ProcessEVMQueue(const CBlock &block, const CBlockIndex *pindex, CCust
     res = cache.AddBalance(minerAddress, {DCT_ID{}, static_cast<CAmount>(blockResult.total_priority_fees)});
     if (!res) return res;
 
-    return Res::Ok();
+    return {blockResult.block_number, Res::Ok()};
+}
+
+
+static Res ProcessDST20Migration(const CBlock &block, const CBlockIndex *pindex, CCustomCSView &cache, const CChainParams& chainparams, const uint64_t evmQueueId, std::array<uint8_t, 20>& beneficiary, const bool evmEnabledOnBlockHead) {
```
Bushstar, line 2491: We do not need block, beneficiary or evmEnabledOnBlockHead passed to this function.

src/masternodes/validation.cpp
```diff
@@ -2483,9 +2484,31 @@ static Res ProcessEVMQueue(const CBlock &block, const CBlockIndex *pindex, CCust
...
+    if (!IsEVMEnabled(pindex->nHeight, cache, chainparams.GetConsensus())) return Res::Ok();
+    auto time = GetTimeMillis();
+    LogPrintf("DST20 migration ...\n");
+
+    CrossBoundaryResult result;
+    cache.ForEachLoanToken([&](DCT_ID const & key, CLoanView::CLoanSetLoanTokenImpl loanToken) {
```
kuegi, line 2497: Should also happen for "normal" DATs, right? Or is the remaining part planned for another PR?

src/masternodes/validation.cpp
```diff
@@ -2486,6 +2488,45 @@ static Res ProcessEVMQueue(const CBlock &block, const CBlockIndex *pindex, CCust
...
+            return result.ok;
+        }
+
+        evm_try_create_dst20(result, evmQueueId, token.creationTx.GetByteArray(),
+                token.name, token.symbol, id.ToString());
+        return result.ok;
+    }, DCT_ID{1});  // start from non-DFI
+    if (!result.ok) return DeFiErrors::DST20MigrationFailure(result.reason.c_str());
+
+    cache.ForEachLoanToken([&](DCT_ID const & id, CLoanView::CLoanSetLoanTokenImpl loanToken) {
```
kuegi, line 2516: loanTokens are DATs, right? so this second loop has no effect as all DATs are already deployed or queued from above.

Jouzo: It should be but there's no check in attributes to assert that every loan tokens are DAT. 
This is just to err on the safe side but it shouldn't affect the overall migration since we guard against `is_dst20_deployed_or_queued`.

Jouzo: Fixed in https://github.com/DeFiCh/ain/pull/2258/commits/fda389e0bc72650ba6c84246250ecdd4dd5faefe and https://github.com/DeFiCh/ain/pull/2258/commits/e30f1faf15ea901c47ba28b69660f198695a1ea4

test/functional/feature_dst20.py
prasannavl: TODO later

PR #2259, opened 2023-07-31
Refactor signing provider get key from destination

sieniven:
## Summary

- Rename GetKeyForDestination to GetKeyOrDefaultFromDestination
- Refactor GetKeyOrDefaultFromDestination to use FromOrDefaultDestionation with new signing provider key type

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/pubkey.h
+++ b/src/pubkey.h
@@ -36,21 +36,18 @@ public:
     static std::optional<CKeyID> TryFromDestination(const CTxDestination &dest, KeyType filter=KeyType::UnknownKeyType) {
         // Explore switching TxDestType to a flag type. Then, we can easily take an allowed
         // flags here and use bit flag logic to decode only specific destinations
-        KeyType destType = FromOrDefaultDestinationTypeToKeyType(dest.index());
-        if ((destType & filter) == KeyType::PKHashKeyType) {
-            return CKeyID(std::get<PKHash>(dest));
-        }
-        else if ((destType & filter) == KeyType::WPKHashKeyType) {
-            return CKeyID(std::get<WitnessV0KeyHash>(dest));
-        }
-        else if ((destType & filter) == KeyType::ScriptHashKeyType) {
-            return CKeyID(std::get<ScriptHash>(dest));
-        }
-        else if ((destType & filter) == KeyType::EthHashKeyType) {
-            return CKeyID(std::get<WitnessV16EthHash>(dest));
-        }
-        else {
-            return {};
+        auto destType = FromOrDefaultDestinationTypeToKeyType(dest.index()) & filter;
+        switch (destType) {
+            case KeyType::PKHashKeyType:
+                return CKeyID(std::get<PKHash>(dest));
+            case KeyType::WPKHashKeyType:
+                return CKeyID(std::get<WitnessV0KeyHash>(dest));
+            case KeyType::ScriptHashKeyType:
+                return CKeyID(std::get<ScriptHash>(dest));
+            case KeyType::EthHashKeyType:
+                return CKeyID(std::get<WitnessV16EthHash>(dest));
+            default:
+                return {};
         }
     }
 
--- a/src/pubkey.h
+++ b/src/pubkey.h
@@ -53,7 +53,7 @@ public:
                 }
                 break;
             case WitV16KeyEthHashType:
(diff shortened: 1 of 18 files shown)


prasannavl:
Looks good. Further, could also abstract the flag checks and remove all the & with a combined singular mask.

prasannavl (dismissed):

src/script/signingprovider.cpp
```diff
@@ -186,28 +186,29 @@ bool FillableSigningProvider::GetCScript(const CScriptID &hash, CScript& redeemS
...
+            break;
+        case WitV16KeyEthHashType:
+            id.type = KeyAddressType::UNCOMPRESSED;
+            break;
+        case ScriptHashType: {
+            CScript script;
+            CScriptID script_id(id);
+            CTxDestination inner_dest;
+            if (store.GetCScript(script_id, script) && ExtractDestination(script, inner_dest)) {
+                id = CKeyID::FromOrDefaultDestination(inner_dest,KeyType::EthHashKeyType);
```
prasannavl, line 206: `            if (auto inner_witness_id = std::get_if<WitnessV0KeyHash>(&inner_dest)) {`

It's checking for V0KeyHash?

src/script/standard.h
```diff
@@ -163,10 +163,11 @@ enum KeyType {
     PKHashKeyType = 1 << 0,
     ScriptHashKeyType = 1 << 1,
     WPKHashKeyType = 1 << 2,
-    EthHashKey = 1 << 3,
+    EthHashKeyType = 1 << 3,
```
prasannavl, line 166: We're using ERC55 terminology now.

src/script/signingprovider.cpp
sieniven: Gosh sorry for this and thanks for the spot.

src/script/standard.h
prasannavl: We can rename the others as well to use the same terminology (in a separate PR), but let's move that direction, than this direction :)

sieniven: Got it. We can do the fix for this in a separate PR.

src/script/standard.h
```diff
@@ -163,10 +163,11 @@ enum KeyType {
...
     MNOperatorKeyType = (1 << 4) | PKHashKeyType | WPKHashKeyType,
     MNOwnerKeyType = (1 << 5) | PKHashKeyType | WPKHashKeyType,
     MNRewardKeyType = (1 << 6) | PKHashKeyType | ScriptHashKeyType | WPKHashKeyType,
+    SigningProviderType = (1 << 7) | PKHashKeyType | ScriptHashKeyType | WPKHashKeyType | EthHashKeyType,
```
prasannavl, line 170: Note: These should just be combinations, Providing it's own values will break many bit flag assumputions

src/pubkey.h
```diff
@@ -36,31 +36,22 @@ class CKeyID : public uint160
...
-                break;
-            case WitV16KeyEthHashType:
-                if ((filter & KeyType::EthHashKeyType) == KeyType::EthHashKeyType) {
-                    return CKeyID(std::get<WitnessV16EthHash>(dest));
-                }
-                break;
-            default: 
-                return {};
+        KeyType destType = FromOrDefaultDestinationTypeToKeyType(dest.index());
+        if ((destType & filter) == KeyType::PKHashKeyType) {
```
prasannavl, line 40: Let's hoist the filtering up and why not just use a switch?

sieniven: Resolved in [f7e7db6](https://github.com/DeFiCh/ain/pull/2259/commits/f7e7db6411a7a5175c58cad5f7ce731e5c0ce4cd)

prasannavl (dismissed):

src/pubkey.h
```diff
@@ -36,31 +36,19 @@ class CKeyID : public uint160
     static std::optional<CKeyID> TryFromDestination(const CTxDestination &dest, KeyType filter=KeyType::UnknownKeyType) {
         // Explore switching TxDestType to a flag type. Then, we can easily take an allowed
```
prasannavl, line 37: Can remove these comments now

Bushstar (dismissed):

PR #2261, opened 2023-07-31
Fix evm tx nonce in feature_evm test

sieniven:
## Summary

- Fix to incorrect evm tx nonce in feature_evm.py

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


sieniven:
MEMPOOL_MAX_ETH_TXS constant param is of value 64, and the mempool will fail to accept a tx into the mempool if the number of txs from that sender in the mempool is >= MEMPOOL_MAX_ETH_TXS.

prasannavl (approved):

PR #2290, opened 2023-08-05
Finalize OP_RETURN default rules

prasannavl:
## Summary

- Finalizes default rule sets of OP_RETURN sizes
  - Can be considered as proxy for EVM, DVM Tx sizes without the envelopes and signatures on DVM

## Protocol

- OP_RETURN sizes on validation:
  - Core: 1k (`v0/rules/tx/core_op_return_max_size_bytes`)
  - DVM: 4k (`v0/rules/tx/dvm_op_return_max_size_bytes`)
  - EVM: 64k  (`v0/rules/tx/evm_op_return_max_size_bytes`)
- Note: The above are part of validation, not just miner policy like the `-datacarriersize`, which still exists and can be used to limit further per node

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


prasannavl:
TODO: 
- Retain the large block tests with a custom attributes set first to allow it. 
- Also retain some tests as copy to that prevent this as expected failures

PR #2317, opened 2023-08-11
Revamp EVM context flow, hash fixes, cleanups

prasannavl:
## Summary

- Uses a single context carry for EVM enable check to simplify things. 
- EVM enables and disables doesn't affect the current block with the `setgov`. Takes effect fully in the next block. 
- Introduces `FromByteArray` to avoid confusion from and to `base_blob` which internally stores bytes in the reversed order. 
- Add more cross boundary helpers.
- Fixes many things, including reversed hashes, miner mining invalid blocks due to missing result checks, etc  
- Few misc cleanups

TODO: 
- Fix migration and block hash change related to it

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None

### DIFF
--- a/src/masternodes/validation.cpp
+++ b/src/masternodes/validation.cpp
@@ -2380,34 +2380,28 @@ static void RevertFailedTransferDomainTxs(const std::vector<std::string> &failed
     }
 }
 
-static Res ValidateCoinbaseXVMOutput(const CScript &scriptPubKey, const FinalizeBlockCompletion &blockResult) {
-    const auto coinbaseBlockHash = uint256(std::vector<uint8_t>(blockResult.block_hash.begin(), blockResult.block_hash.end()));
-    // Miner does not add output on null block
-    if (coinbaseBlockHash.IsNull()) return Res::Ok();
+static Res ValidateCoinbaseXVMOutput(const XVM &xvm, const FinalizeBlockCompletion &blockResult) {
+    const auto blockResultBlockHash = uint256::FromByteArray(blockResult.block_hash);
+    
+    // LogPrintf("DEBUG:: EVMQ Validate: blockResult: %s\n", blockResultBlockHash.ToString());
+    // LogPrintf("DEBUG:: EVMQ Validate: xvm: %s\n", xvm.ToUniValue().write());
 
-    auto res = XVM::TryFrom(scriptPubKey);
-    if (!res.ok) return res;
-
-    auto obj = *res;
-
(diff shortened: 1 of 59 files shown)


src/masternodes/validation.cpp
```diff
@@ -2381,9 +2381,12 @@ static void RevertFailedTransferDomainTxs(const std::vector<std::string> &failed
 }
 
 static Res ValidateCoinbaseXVMOutput(const XVM &xvm, const FinalizeBlockCompletion &blockResult) {
-    const auto coinbaseBlockHash = uint256(std::vector<uint8_t>(blockResult.block_hash.begin(), blockResult.block_hash.end()));
+    const auto blockResultBlockHash = uint256::FromByteArray(blockResult.block_hash);
+    
+    // LogPrintf("DEBUG:: EVMQ Validate: blockResult: %s\n", blockResultBlockHash.ToString());
+    // LogPrintf("DEBUG:: EVMQ Validate: xvm: %s\n", xvm.ToUniValue().write());
```
Jouzo, line 2387:
```suggestion
```

src/miner.cpp
```diff
@@ -389,18 +391,16 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
...
-
-            CDataStream metadata(SER_NETWORK, PROTOCOL_VERSION);
-            metadata << xvm;
-
-            CScript script;
-            script << OP_RETURN << ToByteVector(metadata);
-
-            coinbaseTx.vout[headerIndex].scriptPubKey = script;
+            coinbaseTx.vout[headerIndex].scriptPubKey = xvm.ToScript();
+            // LogPrintf("DEBUG:: CreateNewBlock:: xvm-coinbase-add:: %s\n", XVM::TryFrom(xvm.ToScript())->ToUniValue().write());
```
Jouzo, line 403:
```suggestion
```

src/miner.cpp
```diff
@@ -39,21 +39,21 @@
 #include <random>
 #include <utility>
 
-struct EvmAddressWithNonce {
-    EvmAddress address;
+struct EvmAddressDataWithNonce {
```
Jouzo, line 42: Would keep the names rather than types, i.e. `EvmAddressWithNonce`

Jouzo (approved):

src/miner.cpp
prasannavl: Good catch, agree, will change

sieniven (dismissed):

PR #2321, opened 2023-08-11
Feat: block gas limit by default if not provided in rpc_sendTransaction

canonbrother:
## Summary

- as per title (it was a long time ago "TODO")

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-grpc/src/rpc/eth.rs
+++ b/lib/ain-grpc/src/rpc/eth.rs
@@ -584,9 +584,7 @@ impl MetachainRPCServer for MetachainRPCModule {
         let gas_price = request.gas_price;
         let gas_limit = match request.gas {
             Some(gas_limit) => gas_limit,
-            // TODO(): get the gas_limit from block.header
-            // set 21000 (min gas_limit req) by default first
-            None => U256::from(21000),
+            None => U256::from(self.handler.storage.get_attributes_or_default().block_gas_limit),
         };
         let max_fee_per_gas = request.max_fee_per_gas;
         let message: Option<TransactionMessage> = request.into();


lib/ain-rs-exports/src/evm.rs
```diff
@@ -456,11 +456,17 @@ pub fn evm_unsafe_try_construct_block_in_q(
                 total_priority_fees,
                 block_number,
             }) => {
-                let Ok(total_burnt_fees) = u64::try_from(WeiAmount(total_burnt_fees).to_satoshi()) else {
```
sieniven, line 459: is there a reason for separating the else into a separate line?

lib/ain-rs-exports/src/evm.rs
```diff
@@ -456,11 +456,17 @@ pub fn evm_unsafe_try_construct_block_in_q(
...
+                let Ok(total_burnt_fees) = u64::try_from(WeiAmount(total_burnt_fees).to_satoshi())
+                else {
                     return cross_boundary_error_return(result, "total burnt fees value overflow");
                 };
-                let Ok(total_priority_fees) = u64::try_from(WeiAmount(total_priority_fees).to_satoshi()) else {
```
sieniven, line 462: Same for this? Also the formatting is inconsistent with the above for the ease of reading (although clippy check passes)

lib/ain-rs-exports/src/evm.rs
```diff
@@ -657,26 +663,38 @@ pub fn evm_try_get_tx_by_hash(
             let mut max_priority_fee_per_gas = 0u64;
             match &tx.transaction {
                 TransactionV2::Legacy(transaction) => {
-                    let Ok(price) = u64::try_from(WeiAmount(transaction.gas_price).to_satoshi()) else {
```
sieniven, line 660: Same as above

lib/ain-rs-exports/src/evm.rs
```diff
@@ -657,26 +663,38 @@ pub fn evm_try_get_tx_by_hash(
...
+                    let Ok(price) = u64::try_from(WeiAmount(transaction.gas_price).to_satoshi())
+                    else {
                         return cross_boundary_error_return(result, "tx gas price value overflow");
                     };
                     gas_price = price;
                 }
                 TransactionV2::EIP2930(transaction) => {
                     tx_type = 1u8;
-                    let Ok(price) = u64::try_from(WeiAmount(transaction.gas_price).to_satoshi()) else {
```
sieniven, line 667: Same as above.

lib/ain-rs-exports/src/evm.rs
```diff
@@ -657,26 +663,38 @@ pub fn evm_try_get_tx_by_hash(
...
+                    let Ok(price) = u64::try_from(WeiAmount(transaction.gas_price).to_satoshi())
+                    else {
                         return cross_boundary_error_return(result, "tx gas price value overflow");
                     };
                     gas_price = price;
                 }
                 TransactionV2::EIP1559(transaction) => {
                     tx_type = 2u8;
-                    let Ok(price) = u64::try_from(WeiAmount(transaction.max_fee_per_gas).to_satoshi()) else {
```
sieniven, line 674: Same as above.

lib/ain-rs-exports/src/evm.rs
```diff
@@ -657,26 +663,38 @@ pub fn evm_try_get_tx_by_hash(
...
+                    let Ok(price) =
+                        u64::try_from(WeiAmount(transaction.max_fee_per_gas).to_satoshi())
+                    else {
+                        return cross_boundary_error_return(
+                            result,
+                            "tx max fee per gas value overflow",
+                        );
                     };
                     max_fee_per_gas = price;
-                    let Ok(price) = u64::try_from(WeiAmount(transaction.max_priority_fee_per_gas).to_satoshi()) else {
```
sieniven, line 678: Same as above.

lib/ain-rs-exports/src/evm.rs
canonbrother: sigh.. its `fmt`
https://github.com/DeFiCh/ain/pull/2321/commits/a7ec9f5673089b52f3a203434e96a373cfd9db8f

lib/ain-rs-exports/src/evm.rs
canonbrother: hmm.. is your fmt good which didn't trigger this alert??
lemme drop this.. just with `--no-verify`

sieniven (approved):

shohamc1:
In geth, the default gas limit is set to 90000([ref](https://github.com/ethereum/go-ethereum/blob/e61035c5a3630e4f6fd0fb3e5346a4eed8cedc80/internal/ethapi/api.go#L50)). Wonder if we should adopt the geth standard or stick to this?

canonbrother:
> In geth, the default gas limit is set to 90000([ref](https://github.com/ethereum/go-ethereum/blob/e61035c5a3630e4f6fd0fb3e5346a4eed8cedc80/internal/ethapi/api.go#L50)). Wonder if we should adopt the geth standard or stick to this?

hmm... default by block gas limit is more user friendly tho

prasannavl:
This isn't the correct attribute? This is the block limit. +1 for upping the default TX limit, but let's turn it configurable with a node flag.

canonbrother:
> This isn't the c

canonbrother:
> This isn't the correct attribute? This is the block limit. +1 for upping the default TX limit, but let's turn it configurable with a node flag.

gosh.. lemme grab the block gas limit from header

canonbrother:
> This isn't the correct attribute? This is the block limit. +1 for upping the default TX limit, but let's turn it configurable with a node flag.

hey @prasannavl weird.. isn't it a block gas limit?
https://github.com/DeFiCh/ain/blob/938a56135f5cc4f55f1e35ffd07d5d22523bedee/src/ffi/ffiexports.h#L9
30 mil unit is correct

canonbrother:
> gosh.. lemme grab the block gas limit from header

i can't do this yet before we have block gas limit computation

canonbrother:
(dangling PR): redo if necessary

PR #2324, opened 2023-08-14
Refactor: replace `block_number_to_u256` by `get_block`

canonbrother:
## Summary

- the intention is to have `get_block` utils for happy dev, the `block number` can be grabbed from `get_block`

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-grpc/src/rpc/eth.rs
+++ b/lib/ain-grpc/src/rpc/eth.rs
@@ -5,12 +5,13 @@ use std::sync::Arc;
 use ain_cpp_imports::get_eth_priv_key;
 use ain_evm::bytes::Bytes;
 use ain_evm::core::EthCallArgs;
+use ain_evm::block::BlockError;
 use ain_evm::evm::EVMServices;
 use ain_evm::executor::TxResponse;
 use ain_evm::filters::Filter;
 use ain_evm::storage::traits::{BlockStorage, ReceiptStorage, TransactionStorage};
 use ain_evm::transaction::{SignedTx, TransactionError};
-use ethereum::{EnvelopedEncodable, TransactionV2};
+use ethereum::{BlockAny, EnvelopedEncodable, TransactionV2};
 use jsonrpsee::core::{Error, RpcResult};
 use jsonrpsee::proc_macros::rpc;
 use libsecp256k1::SecretKey;
@@ -274,7 +275,7 @@ impl MetachainRPCModule {
         Self { handler }
     }
 
-    fn block_number_to_u256(&self, block_number: Option<BlockNumber>) -> RpcResult<U256> {
+    fn get_block(&self, block_number: Option<BlockNumber>) -> Option<BlockAny> {
         match block_number.unwrap_or_default() {
             BlockNumber::Hash { hash, .. } => self.handler.storage.get_block_by_hash(&hash),
             BlockNumber::Num(n) => self.handler.storage.get_block_by_number(&U256::from(n)),
@@ -293,12 +294,17 @@ impl MetachainRPCModule {
                         .checked_sub(U256::from(finality_count))?;
                     self.handler.storage.get_block_by_number(&safe_block_number)
                 })
-            }
+            },
             // BlockNumber::Pending => todo!(),
             _ => self.handler.storage.get_latest_block(),
         }
-        .ok_or(Error::Custom(String::from("header not found")))
-        .map(|block| block.header.number)
+    }
+
+    fn try_get_block_number(&self, block_number: Option<BlockNumber>) -> Result<U256, BlockError> {
+        match self.get_block(block_number) {
+            Some(block) => Ok(block.header.number),
+            None => Err(BlockError::BlockNotFound),
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -293,12 +294,17 @@ impl MetachainRPCModule {
                         .checked_sub(U256::from(finality_count))?;
                     self.handler.storage.get_block_by_number(&safe_block_number)
                 })
-            }
+            },
```
sieniven, line 297: Remove changes here which fails rust lint check

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -5,12 +5,13 @@ use std::sync::Arc;
 use ain_cpp_imports::get_eth_priv_key;
 use ain_evm::bytes::Bytes;
 use ain_evm::core::EthCallArgs;
+use ain_evm::block::BlockError;
```
sieniven, line 8: Failing rust lint check here too

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -452,11 +469,7 @@ impl MetachainRPCServer for MetachainRPCModule {
         block_number: BlockNumber,
         full_transactions: Option<bool>,
     ) -> RpcResult<Option<RpcBlock>> {
-        let block_number = self.block_number_to_u256(Some(block_number))?;
-        debug!(target:"rpc", "Getting block by number : {}", block_number);
-        self.handler
-            .storage
-            .get_block_by_number(&block_number)
+        self.get_block(Some(block_number))
```
sieniven, line 472: Use fallible try_get_block method instead here?

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -413,15 +428,17 @@ impl MetachainRPCServer for MetachainRPCModule {
...
-        self.handler
-            .storage
-            .get_block_by_hash(&hash)
-            .map_or(Ok(None), |block| {
-                Ok(Some(RpcBlock::from_block_with_tx(
-                    block,
-                    full_transactions.unwrap_or_default(),
-                )))
-            })
+        self.get_block(Some(BlockNumber::Hash {
```
sieniven, line 431: Use fallible try_get_block instead here?

lib/ain-grpc/src/rpc/eth.rs
sieniven: Resolved in [e7668d5](https://github.com/DeFiCh/ain/pull/2324/commits/e7668d577a89d0f85127fa30efc090f635530d3c)

lib/ain-grpc/src/rpc/eth.rs
sieniven: Resolved in [e7668d5](https://github.com/DeFiCh/ain/pull/2324/commits/e7668d577a89d0f85127fa30efc090f635530d3c)

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -297,8 +298,13 @@ impl MetachainRPCModule {
             // BlockNumber::Pending => todo!(),
             _ => self.handler.storage.get_latest_block(),
         }
-        .ok_or(Error::Custom(String::from("header not found")))
-        .map(|block| block.header.number)
+    }
+
+    fn try_get_block_number(&self, block_number: Option<BlockNumber>) -> Result<U256, BlockError> {
```
sieniven, line 303: Why do we need to wrap this method with the get block implementation which is fallible? We should have a single try_get_block_by_number method that returns BlockError consistently through all RPCs, and propagated accordingly

lib/ain-grpc/src/rpc/eth.rs
sieniven: Correction - get_block is fallible, we should do a proper method try_get_block_from_number, and implementers should propagate the block error accordingly if the block query fails from the db

lib/ain-grpc/src/rpc/eth.rs
sieniven: Same as above ^

lib/ain-grpc/src/rpc/eth.rs
canonbrother: > Why do we need to wrap this method with the get block implementation which is fallible?

hmm... just to have a utils to get block number easily... and to change fallible from `Option<T>` to `Result<T>`

> We should have a single try_get_block_by_number method 

afraid this one requires more refactoring 😅 

> that returns BlockError consistently through all RPCs, and propagated accordingly

the `get_block() -> Option<T>` is actually following ethereum getBlock behavior (throwing null instead of error)
```
➜  metachain-ts-tests git:(main) ✗ curl https://eth.meowrpc.com \
  -H 'content-type:application/json' \
  --data-binary \
  '{
    "jsonrpc":"2.0",
    "id":"2",
    "method":"eth_getBlockByNumber",
    "params":["0x20f3e96",false]
  }'
{"jsonrpc": "2.0", "id": "2", "result": null}%
```

canonbrother:
convert to draft first before fix the conflicts

PR #2325, opened 2023-08-14
evm: Run DST20 migration during mining

shohamc1:
WIP, looking into how we can use `ProcessFallibleEvent` during mining.


prasannavl:
Logic should be equivalent to moving migration into connect block. Merging for now to get going, until it's reworked.

PR #2327, opened 2023-08-14
EVM:  DST20 migration TX

Jouzo:
## Summary

- Create TX and receipt for DST20 migration SystemTx
- Remove redundant pubkey from SignedTx. Only used to retrieve sender which is already indexed alongside the TransactionV2.
- Use transaction idx as nonce and block number as value to obtain unique hash for every DST20 migration SystemTx.
- Additional `ReceiptAndOptionalContractAddress` type to set custom contract address in generate receipt.

## Missing

- TBD whether to switch logic to appending systemtx on genesis block instead of initiating migration from validation.
- Fix `ValidateCoinbaseXVMOutput` issues. Should be fixed in https://github.com/DeFiCh/ain/pull/2317

## Implications

- Storage

- Consensus
  - [x] Includes backward incompatible changes


lib/ain-evm/src/evm.rs
```diff
@@ -563,3 +573,31 @@ impl EVMServices {
...
+
+fn create_deploy_contract_tx(
+    idx: usize,
+    block_number: U256,
+) -> Result<(SignedTx, ReceiptV3), Box<dyn Error>> {
+    let tx = SignedTx {
+        sender: H160::zero(),
+        transaction: TransactionV2::Legacy(LegacyTransaction {
+            nonce: U256::from(idx),
+            gas_price: U256::zero(),
```
shohamc1, line 585: Would be better to set this to base fee instead IMO.

Jouzo:
Needs https://github.com/DeFiCh/ain/pull/2329 to fix migration EVM block hash mismatch

PR #2329, opened 2023-08-15
Bug fixes for endianness between boundaries

sieniven:
## Summary

- Fixes to inconsistencies in converting to U256 type on Rust to pass by u64 instead between cross boundaries.
- Fixes to all inconsistencies in converting to H256 type on Rust to pass hashes by strings instead.
- All EVM sender addresses on consensus and on miner will now be indexed with the sender address which is in string type.
- When passing hash values into EVM side, we will convert them from Rust string types.
- When passing hash values into DVM side, we will convert them from strings to BTC internal hash representation, which is stored in little endian.

## Context

1. Rust U256:
- From: takes a big endian byte array, stores as little endian.
- To: outputs a big endian byte array.

2. BTC U256:
- Stored as little endian, with the exception of Witness16EthHash
- GetByteArray returns byte array in big endian.

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


src/masternodes/mn_checks.cpp
```diff
@@ -3911,22 +3911,18 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
                 CTxDestination dest;
                 ExtractDestination(dst.address, dest);
                 const auto toAddress = std::get<WitnessV16EthHash>(dest);
-                arith_uint256 balanceIn = dst.amount.nValue;
+                const auto balanceIn = static_cast<uint64_t>(dst.amount.nValue);
```
prasannavl, line 3914: Checked numerics

src/masternodes/mn_checks.cpp
```diff
@@ -3939,13 +3935,11 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
                 CTxDestination dest;
                 ExtractDestination(src.address, dest);
                 const auto fromAddress = std::get<WitnessV16EthHash>(dest);
-                arith_uint256 balanceIn = src.amount.nValue;
+                const auto balanceIn = static_cast<uint64_t>(src.amount.nValue);
```
prasannavl, line 3938: Checked numerics

src/masternodes/mn_checks.cpp
```diff
@@ -4004,30 +3996,32 @@ class CCustomTxApplyVisitor : public CCustomTxVisitor {
...
                 LogPrintf("[evm_try_prevalidate_raw_tx] failed, reason : %s\n", result.reason);
                 return Res::Err("evm tx failed to validate %s", result.reason);
             }
         }
 
-        auto txHash = tx.GetHash();
-        auto evmTxHashBytes = std::vector<uint8_t>(validateResults.tx_hash.begin(), validateResults.tx_hash.end());
-        auto evmTxHash = uint256S(HexStr(evmTxHashBytes));
+        auto txHash = tx.GetHash().GetHex();
+        if (validateResults.tx_hash.length() < 2) {
```
prasannavl, line 4014: Can move into the above branch.

PR #2334, opened 2023-08-16
evm: Reserve DST20 and precompile address space

shohamc1:


lib/ain-evm/src/precompiles/mod.rs
```diff
@@ -84,17 +84,11 @@ pub struct MetachainPrecompiles;
...
-            hash(2),
-            hash(3),
-            hash(4),
-            hash(5),
-            hash(6),
-            hash(7),
-            hash(8),
-            hash(9),
-        ]
+        (1..=1024)
```
canonbrother, line 87: not clear what's the rational behind of setting `1024`.. but the return should follow the same??
(1..=1024) -> [H160; 1024]
or
(1..=9) -> [H160; 9]

shohamc1: We are reserving additional addresses for future use.

PR #2337, opened 2023-08-16
EVM: RocksDB as BlockStore backend

Jouzo:
## Summary

- Replace in-memory data handler with rocksdb backed `BlockStore`

## Implications

- Storage

- Consensus
  - [x] Includes backward incompatible changes


prasannavl:
Let's redesign this later so it's easier to keep a unified snapshot view of all items per block number, but LGTM for now.

TODO: Now that things are on disk, we should add a periodic flush

PR #2371, opened 2023-08-23
evm: Save contract artifacts to build tree

shohamc1:


test/functional/feature_dst20.py
```diff
@@ -787,27 +787,48 @@ def run_test(self):
...
                 "r",
                 encoding="utf8",
             ).read()
-        )["object"]
-
-        self.bytecode = json.loads(
-            open(
-                f"{os.path.dirname(__file__)}/../../lib/ain-contracts/dst20/output/bytecode.json",
+            self.bytecode = json.loads(
+                open(
```
prasannavl, line 798: Let's abstract the path finder into a function, instead of if else of the whole codepath.

prasannavl:
Merging, but we should still improve these tests to cut out the noise post PR

PR #2373, opened 2023-08-23
Fix lint

sieniven:
## Summary

- Fix sh lint

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None


prasannavl:
This kinda won't make a lot of sense, since there's no way to coherently pass the same args to test_cpp and test_rs. So might just be clear and not pass args at all on test_unit adding a lint ignore 

But merging this to clear out lints.

PR #2375, opened 2023-08-23
evm: Enable DST20 optimisation

shohamc1:


lib/ain-contracts/build.rs
```diff
@@ -26,8 +28,28 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             .sources(&root)
             .build()?;
 
+        let solc_config = SolcConfig::builder()
+            .settings(Settings {
+                stop_after: None,
+                remappings: vec![],
+                optimizer: Optimizer {
+                    enabled: Some(true),
+                    runs: Some(10000),
```
prasannavl, line 37: Let's use the max val. `2**32-1`, since the optimizer conservatively runs only as much as it needs to

PR #2380, opened 2023-08-24
evm: Add tests related to contracts size

cuongquangnam:
## Summary

- Test deployment of various contract sizes 

## Implications

- Storage
  - [x] Database reindex not required
  - [ ] None

- Consensus
  - [x] None


test/functional/feature_evm_contracts.py
```diff
@@ -99,6 +144,9 @@ def should_create_contract(self):
         self.contract = node.w3.eth.contract(
             address=receipt["contractAddress"], abi=abi
         )
+        size_of_runtime_bytecode = len(node.w3.eth.get_code(self.contract.address))
+        assert_greater_than(size_of_runtime_bytecode, 0)
+        assert_greater_than(1_000, size_of_runtime_bytecode)
```
canonbrother, line 149: can we assert the exact bytecode size?? (which is more informative)

canonbrother:
GREAT for this changes!! https://github.com/DeFiCh/ain/pull/2380/commits/1463572a0215568bed666a18061f925921b5bc0d

canonbrother:
can we add a time measurement in each test?? (more info req)

test/functional/feature_evm_contracts.py
```diff
@@ -60,7 +62,50 @@ def setup(self):
...
+
+        function_template = lambda index: """
+    function func{}() public pure returns(uint256) {{
+        return {};
+    }}""".format(
+            index, index
+        )
+
+        contract_end = """
+}"""
```
shohamc1, line 81:
```suggestion
        contract_start = f"""
        pragma solidity ^0.8.0;
        contract {contract_name} {{
        """

        function_template = (
            lambda index: f"""
        function func{index}() public pure returns(uint256) {{
            return {index};
        }}"""
        )

        contract_end = "\n}"
```

test/functional/feature_evm_contracts.py
```diff
@@ -60,7 +62,50 @@ def setup(self):
...
+
+        list_sig = []
+        contract_body = ""
+
+        for i in range(0, num_functions):
+            func_sig = "func${}()".format(i)
```
shohamc1, line 87:
```suggestion
            func_sig = f"func${i}()"
```

test/functional/feature_evm_contracts.py
```diff
@@ -60,7 +62,50 @@ def setup(self):
...
+            list_sig.append(sig_hash)
+            contract_body += function_template(i)
+
+        utf8SourceCode = contract_start + contract_body + contract_end
+
+        compiled_output = compile_source(
+            source=utf8SourceCode,
+            output_values=["abi", "bin"],
+            solc_version="0.8.20",
+        )
```
shohamc1, line 100: I think it would be better to extend [`EVMContract`](https://github.com/DeFiCh/ain/blob/0b2b7f7eeff05a41bf942e21b73826bef19b0b6a/test/functional/test_framework/evm_contract.py#L7) to have a new `from_str` function instead of calling `solcx` directly.

test/functional/feature_evm_contracts.py
```diff
@@ -181,15 +229,164 @@ def failed_tx_should_increment_nonce(self):
 
         assert_equal(before_tx_count + 1, after_tx_count)
 
+    def should_deploy_contract_1KB_To_10KB(self):
```
shohamc1, line 232: Maybe use the actual contract size here instead of the range?

shohamc1 (changes requested):

cuongquangnam:
> can we add a time measurement in each test?? (more info req)

Why do we need time measurements?

canonbrother:
> > can we add a time measurement in each test?? (more info req)
> 
> Why do we need time measurements? @canonbrother

@cuongquangnam
for benchmarking purpose.. just a reference for us to know deployment time against the bytecode size

easier as
```
start_time = time.time()
// code snippet...
end_time = time.time()
elapsed_time = end_time - start_time
print("Elapsed time: {} s".format(elapsed_time))
```

shohamc1:
Don't think we need this as part of the test? We can add it ad-hoc when needed.

canonbrother (approved):

test/functional/feature_evm_contracts.py
```diff
@@ -181,15 +217,168 @@ def failed_tx_should_increment_nonce(self):
...
+        node = self.nodes[0]
```
shohamc1, line 221: Can set `self.node = self.nodes[0]` so that you don't have to repeat this in every test.

test/functional/feature_evm_contracts.py
```diff
@@ -181,15 +217,168 @@ def failed_tx_should_increment_nonce(self):
...
+                "maxFeePerGas": 10_000_000_000,
+                "maxPriorityFeePerGas": 1_500_000_000,
+            }
+        )
+        # to check the init code is larger than 49152
+        assert_equal((len(tx["data"]) - 2) / 2, 243542)
+        signed = node.w3.eth.account.sign_transaction(tx, self.evm_key_pair.privkey)
+
+        try:
+            node.w3.eth.send_raw_transaction(signed.rawTransaction)
```
shohamc1, line 348: This should always fail, so it should not be in a `try` block.

shohamc1 (changes requested):

test/functional/feature_evm_contracts.py
shohamc1: Fixed: f4aff434c00d716b00880164e1428177d27be655

test/functional/feature_evm_contracts.py
shohamc1: Fixed: b705a6e7247969536aa55740ef05c2e2e7ca15d6

prasannavl:
> https://github.com/DeFiCh/ain/pull/2380/commits/f4aff434c00d716b00880164e1428177d27be655

On this, I'd actually prefer the localized var node than repeated self. 

However @shohamc1 how about extrapolating on your idea and setting self.node directly into the test framework that aliases nodes[0] or is None?

shohamc1:
> > [f4aff43](https://github.com/DeFiCh/ain/commit/f4aff434c00d716b00880164e1428177d27be655)
> 
> On this, I'd actually prefer the localized var node than repeated self. It's a good practise to pull all vars in the beginning of the function into the localized context and use those.
> 
> However @shohamc1 how about extrapolating on your idea and setting self.node directly into the test framework that aliases nodes[0] or is None?

Maybe we can do this on a different PR instead? This one is quite large as it is right now.

prasannavl:
Sure, sounds good to me

test/functional/feature_evm_contracts.py
```diff
@@ -244,91 +244,39 @@ def failed_tx_should_increment_nonce(self):
...
-        assert_equal(receipt["status"], 1)
-        assert_equal(size_of_runtime_bytecode, 21140)
-        # sanity check for the equality between the runtime bytecode generated by the compiler
-        # and the runtime code deployed
-        assert_equal(compiler_runtime_bytecode, runtime_bytecode.hex()[2:])
+    
+    def should_deploy_contract_with_different_sizes(self):
+        iterationToContractName = {128: "ContractSize1KBTo10KB", 256: "ContractSize10KBTo19KB", 400: "ContractSize20KBTo29KB"}
+        iterationToContractSize = {128: 6901, 256: 13685, 400: 21140}
+        for iteration in [128, 256, 400]:
```
shohamc1, line 251:
```python
        test_data = [(6901, "ContractSize1KBTo10KB"), (13685, "ContractSize10KBTo19KB"), (21140, "ContractSize20KBTo29KB")
        for (size, contract_name) in test_data:
```

shohamc1 (dismissed):

PR #2382, opened 2023-08-25
Fix tx context in validate tx pipeline

sieniven:
## Summary

- Fix incorrect transaction context when setting vicinity before calling the tx on EVM in validate_raw_transaction in a queue context.

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None

### DIFF
--- a/lib/ain-evm/src/transaction/mod.rs
+++ b/lib/ain-evm/src/transaction/mod.rs
@@ -303,9 +303,9 @@ impl SignedTx {
 
     pub fn get_tx_type(&self) -> U256 {
         match &self.transaction {
-            TransactionV2::Legacy(tx) => U256::from(0),
-            TransactionV2::EIP2930(tx) => U256::from(1),
-            TransactionV2::EIP1559(tx) => U256::from(2),
+            TransactionV2::Legacy(_) => U256::from(0),
+            TransactionV2::EIP2930(_) => U256::from(1),
+            TransactionV2::EIP1559(_) => U256::from(2),
         }
     }
 }
--- a/lib/ain-evm/src/transaction/mod.rs
+++ b/lib/ain-evm/src/transaction/mod.rs
@@ -300,6 +300,14 @@ impl SignedTx {
             TransactionV2::EIP1559(tx) => tx.hash(),
         }
     }
+
+    pub fn get_tx_type(&self) -> U256 {
+        match &self.transaction {
+            TransactionV2::Legacy(tx) => U256::from(0),
+            TransactionV2::EIP2930(tx) => U256::from(1),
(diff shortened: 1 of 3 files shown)


shohamc1 (approved):

lib/ain-evm/src/transaction/mod.rs
```diff
@@ -300,6 +300,14 @@ impl SignedTx {
...
     }
+
+    pub fn get_tx_type(&self) -> U256 {
+        match &self.transaction {
+            TransactionV2::Legacy(_) => U256::from(0),
+            TransactionV2::EIP2930(_) => U256::from(1),
+            TransactionV2::EIP1559(_) => U256::from(2),
+        }
+    }
 }
```
Jouzo, line 311: TransacationV2 already has a type_id() method

sieniven: Resolved in commit [70c035a](https://github.com/DeFiCh/ain/pull/2382/commits/70c035ad71c3024bc8ddfc755a8e6bfc8ff9e940)

Jouzo (approved):

lib/ain-evm/src/transaction/mod.rs
```diff
@@ -300,6 +300,10 @@ impl SignedTx {
             TransactionV2::EIP1559(tx) => tx.hash(),
         }
     }
+
+    pub fn get_tx_type(&self) -> U256 {
+        U256::from(EnvelopedEncodable::type_id(&self.transaction).unwrap_or_default())
```
shohamc1, line 305: Is it safe to use `unwrap_or_default` here? Maybe this should return a `Result` instead.

sieniven: This is correct. For legacy TransactionV2 txs, None is returned. Thus we unwrap to default (0 value), and Legacy txs are of type value 0.

lib/ain-evm/src/transaction/mod.rs
```diff
@@ -424,7 +428,7 @@ mod tests {
     fn get_test_data(path: &Path) -> Result<Vec<(String, ExpectedTx)>, Box<dyn Error>> {
         let content = fs::read_to_string(path)?;
 
-        let r = serde_json::from_str(&content)?;
+        let r: Vec<(String, ExpectedTx)> = serde_json::from_str(&content)?;
```
prasannavl, line 431: Let's drop the explicit typing here?

prasannavl: It's a test, so moving on

sieniven: resolved

PR #2390, opened 2023-08-29
evm: Add next block baseFeePerGas in `eth_feeHistory` output

shohamc1:
Refining rpc `eth_feeHistory` with
  - [x] add next block base fee
  - [x] change oldest block type from hash to number
  - [x] some percentiles to fine tune

### DIFF
--- a/lib/ain-evm/src/block.rs
+++ b/lib/ain-evm/src/block.rs
@@ -174,6 +174,7 @@ impl BlockService {
             block_number -= U256::one();
         }
 
+        // TODO(): b.hash -> b.number
         let oldest_block = blocks.last().unwrap().header.hash();
 
         let (mut base_fee_per_gas, mut gas_used_ratio): (Vec<U256>, Vec<f64>) = blocks
--- a/lib/ain-evm/src/block.rs
+++ b/lib/ain-evm/src/block.rs
@@ -17,7 +17,7 @@ pub struct BlockService {
 }
 
 pub struct FeeHistoryData {
-    pub oldest_block: H256,
+    pub oldest_block: U256,
     pub base_fee_per_gas: Vec<U256>,
     pub gas_used_ratio: Vec<f64>,
     pub reward: Option<Vec<Vec<U256>>>,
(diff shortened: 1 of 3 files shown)


lib/ain-evm/src/block.rs
```diff
@@ -246,7 +246,25 @@ impl BlockService {
             Some(reward)
         };
 
+        // add another entry for baseFeePerGas
+        let next_block_base_fee = match self
+            .storage
+            .get_block_by_number(&(first_block + U256::one()))?
```
canonbrother, line 252: can you test if the first_block == latest_block?

canonbrother (changes requested):

test/functional/feature_evm_rpc_fee_history.py
```diff
@@ -0,0 +1,141 @@
...
+       node = self.nodes[0]
+       for x in range(count):
+            for y in priority_fees:
+                tx = {
+                    "from": self.ethAddress,
+                    "value": "0x0",
+                    "data": CONTRACT_BYTECODE,
+                    "gas": "0x18e70",  # 102_000
+                    "maxPriorityFeePerGas": hex(16),
+                    # "maxPriorityFeePerGas": hex(y), # <-- too-many-eth-txs-by-sender
```
canonbrother, line 108: edge case

lib/ain-evm/src/block.rs
```diff
@@ -174,6 +174,7 @@ impl BlockService {
             block_number -= U256::one();
         }
 
+        // TODO(): b.hash -> b.number
```
canonbrother, line 177: todo

lib/ain-evm/src/block.rs
canonbrother: resolved at https://github.com/DeFiCh/ain/pull/2390/commits/e8243fb9d8e2baacf2a72758083a40fb946a7d2c

test/functional/feature_evm_rpc_fee_history.py
canonbrother: 64 txs is maximum in mempool for the same sender

lib/ain-evm/src/block.rs
shohamc1: Good catch, thank you.

PR #2396, opened 2023-08-31
evm: Add tests for proxy

cuongquangnam:
## Summary

- Add functional tests related to proxy smart contracts

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


test/functional/feature_evm_proxy.py
```diff
@@ -0,0 +1,281 @@
...
+            "maxFeePerGas": 10_000_000_000,
+            "maxPriorityFeePerGas": 1_500_000_000,
+        }
+        signed = node.w3.eth.account.sign_transaction(
+            transaction, self.evm_key_pair.privkey
+        )
+        hash = node.w3.eth.send_raw_transaction(signed.rawTransaction)
+        node.generate(1)
+        receipt = node.w3.eth.wait_for_transaction_receipt(hash)
+        assert_equal(receipt["status"], 0)
```
canonbrother, line 172: is this how we check whether the tx success or failure?? 
will it have error msg??

test/functional/feature_evm_proxy.py
```diff
@@ -0,0 +1,281 @@
...
+                "nonce": node.w3.eth.get_transaction_count(self.evm_key_pair.address),
+                "gas": 1_000_000,
+            }
+        )
+        signed = node.w3.eth.account.sign_transaction(tx, self.evm_key_pair.privkey)
+        hash = node.w3.eth.send_raw_transaction(signed.rawTransaction)
+        node.generate(1)
+        receipt = node.w3.eth.wait_for_transaction_receipt(hash)
+        assert_equal(receipt["status"], 1)
+        assert_equal(proxy_contract.functions.randomVar().call(), 20)
```
canonbrother, line 153: since its named `should_call_function_in_implementation_contract`
missing the call to implementation contract??

test/functional/feature_evm_proxy.py
cuongquangnam: currently Im not aware of anyway to show the error for contract error. My best guess is that we have to wait for the traceTransaction RPC PR to finish to have better tests

cuongquangnam:
what do you mean by caller -> proxy -> implementation?
the current way is Externally Owned address -> proxy contract -> implementation contract

test/functional/feature_evm_proxy.py
cuongquangnam: Oh, what I mean is proxy delegate call to implementation contract. Changed the name of the test

canonbrother:
> What do you mean by NIT?

NIT: not important

canonbrother (approved):

shohamc1:
> What do you mean by NIT?
> And IMO, we don't need this kind of contract -> proxy -> implementation test when we already have EOA -> proxy -> implementation test

I think there is some value in having a test that checks functionality for different levels of internal calls.

PR #2407, opened 2023-09-04
Remove EVM block size check in validation

Bushstar:


sieniven (approved):

prasannavl:
This breaks protocol going forward, but rollback shouldn't be needed for dev chain.

PR #2443, opened 2023-09-12
Chore: bind eth port

canonbrother:
## Summary

To solve curl eth rpc result "Empty reply from server" from docker


canonbrother:
close as `-ethrpcbind`

PR #2453, opened 2023-09-15
Restore legacy minter info

Bushstar:
Restores the behaviour found in release 3.8.5 and before when printing out minter information for blocks. The previous behaviour can be found in the link below. This prevents masternode being printed in the genesis block, also regardless of a valid MN ID both mintedBlocks and stakeModifier should be added as these are based on the on block index.

https://github.com/DeFiCh/ain/blob/5786fb81d9d679a12ee5cc6d516581af03468c25/src/rpc/blockchain.cpp#L137

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -125,23 +125,26 @@ struct MinterInfo {
     std::string OperatorAddress{};
     std::string RewardAddress{};
     uint64_t MintedBlocks{};
-    uint256 StakeModifier{};
+    std::string StakeModifier{};
 
-    static std::optional<MinterInfo> TryFrom(const CBlock& block, const CBlockIndex* blockindex, const CCustomCSView& view) {
+    static MinterInfo From(const CBlock& block, const CBlockIndex* blockindex, const CCustomCSView& view) {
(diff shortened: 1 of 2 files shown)


prasannavl (approved):

src/rpc/blockchain.cpp
```diff
@@ -125,35 +125,42 @@ struct MinterInfo {
     std::string OperatorAddress{};
     std::string RewardAddress{};
     uint64_t MintedBlocks{};
-    uint256 StakeModifier{};
+    std::string StakeModifier{};
```
prasannavl, line 128: Sure we don't need the StakeModifier in it's raw form here for other use cases? 

Note: `MinterInfo` should be the source for everything where this stakeModifier is used, eventually. We shouldn't be having loose ad-hoc types all over the place. Please make sure that the common denominator is used here, and converted when it's used, rather than directly storing the converted types that's only needed on outer use cases like printing.

prasannavl: Note: Merging, but marking for retrospective check to ensure the base types can absorb the most flexible use cases for future refactors.

PR #2456, opened 2023-09-16
Add fast EVM RBF to mempool

Bushstar:
- Adds the following to the CTxMemPoolEntry
  - EvmAddressWithNonce
  - CustomTxType
  - EVM Prepay Fee
- Adds to the mempool multi-index an order by EvmAddressWithNonce

Using these two when a new TX goes to enter the mempool we can look up any existing entires with EvmAddressWithNonce and check against the existing prepay fee. If the new fee is more then we will replace the existing entry.

When removing entries from the mempool we do not rebuild the attached account view as this would incur a performance cost on each transfer domain transaction. It is likely sufficient to suffer an imperfect attached account view which would not have the transfers in the transfer domains transaction removed. The account view will be rebuilt on each new block so this state should not be in place for long.

### DIFF
--- a/lib/ain-evm/src/core.rs
+++ b/lib/ain-evm/src/core.rs
@@ -345,6 +345,11 @@ impl EVMCoreService {
         })
     }
 
+    pub fn calculate_prepay_gas_fee(&self, signed_tx: &SignedTx) -> Result<U256> {
+        let result = calculate_prepay_gas_fee(&signed_tx)?;
+        Ok(result)
+    }
+
     /// Validates a raw transfer domain tx.
     ///
     /// The pre-validation checks of the tx before we consider it to be valid are:
--- a/lib/ain-rs-exports/src/evm.rs
+++ b/lib/ain-rs-exports/src/evm.rs
@@ -1136,11 +1136,20 @@ pub fn evm_try_get_tx_sender_info_from_raw_tx(
         return cross_boundary_error_return(result, "nonce value overflow");
     };
 
+    let Ok(prepay_fee) = SERVICES.evm.core.calculate_prepay_gas_fee(&signed_tx) else {
+        return cross_boundary_error_return(result, "failed to get fee from transaction");
+    };
+
+    let Ok(prepay_fee) = u64::try_from(prepay_fee) else {
+        return cross_boundary_error_return(result, "prepay fee value overflow");
+    };
+
     cross_boundary_success_return(
         result,
         TxSenderInfo {
             nonce,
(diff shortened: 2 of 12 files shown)


lib/ain-rs-exports/src/evm.rs
```diff
@@ -1136,11 +1136,20 @@ pub fn evm_try_get_tx_sender_info_from_raw_tx(
         return cross_boundary_error_return(result, "nonce value overflow");
     };
 
+    let Ok(prepay_fee) = SERVICES.evm.core.calculate_prepay_gas_fee(&signed_tx) else {
```
Jouzo, line 1139: Can switch fee module to public `pub mod fee;`
and call `calculate_prepay_gas_fee` directly
```
    let Ok(prepay_fee) = calculate_prepay_gas_fee(&signed_tx) else {
        return cross_boundary_error_return(result, "failed to get fee from transaction");
    };
```

lib/ain-evm/src/core.rs
```diff
@@ -345,6 +345,11 @@ impl EVMCoreService {
         })
     }
 
+    pub fn calculate_prepay_gas_fee(&self, signed_tx: &SignedTx) -> Result<U256> {
+        let result = calculate_prepay_gas_fee(signed_tx)?;
+        Ok(result)
+    }
```
Jouzo, line 351: Can remove here and set fee mod public

src/ffi/ffiexports.cpp
```diff
@@ -167,21 +167,26 @@ rust::vec<TransactionData> getPoolTransactions() {
...
                     HexStr(obj.transfers[0].second.data),
                     static_cast<uint8_t>(TransactionDataDirection::DVMToEVM),
                 });
             } else if (obj.transfers[0].first.domain == static_cast<uint8_t>(VMDomain::EVM) && obj.transfers[0].second.domain == static_cast<uint8_t>(VMDomain::DVM)) {
-                poolTransactions.push_back({
+                poolTransactionsByFee.emplace(mi->GetEVMPrePayFee(), TransactionData{
                     static_cast<uint8_t>(TransactionDataTxType::TransferDomain),
-                    HexStr(obj.transfers[0].second.data),
+                    HexStr(obj.transfers[0].first.data),
                     static_cast<uint8_t>(TransactionDataDirection::DVMToEVM),
```
Jouzo, line 179: `static_cast<uint8_t>(TransactionDataDirection::EVMToDVM)`

PR #2473, opened 2023-09-20
Evm: Miner fixes

sieniven:
## Summary

- Fixes EVM block hash mismatch due to evm/transferdomain txs processed in different order on the miner and on connect block.
- Fixes adding incorrect txiter to failed tx set on miner pipeline that causes missing transferdomain txs inside evm block.
- Fixes segfault error on miner happening due to incorrect assertion from evm tx iters being processed again inside failedNonces set when they were already successfully minted into the block from another transferdomain tx iter descendant.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -866,6 +866,7 @@ void BlockAssembler::addPackageTxs(int& nPackagesSelected, int& nDescendantsUpda
                 // Not okay invalidate, undo and skip
                 if (!res.ok) {
                     customTxPassed = false;
+                    failedCustomTx = tx.GetHash();
                     LogPrintf("%s: Failed %s TX %s: %s\n", __func__, ToString(txType), tx.GetHash().GetHex(), res.msg);
                     break;
                 }
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -47,6 +47,7 @@ struct EvmTxPreApplyContext {
     int height;
     uint64_t evmQueueId;
     std::multimap<uint64_t, CTxMemPool::txiter>& failedNonces;
+    std::map<uint256, CTxMemPool::FailedNonceIterator>& failedNoncesLookup;
     CTxMemPool::setEntries& failedTxEntries;
 };
 
@@ -606,6 +607,7 @@ bool BlockAssembler::EvmTxPreapply(EvmTxPreApplyContext& ctx)
     const auto& height = ctx.height;
     const auto& failedTxSet = ctx.failedTxEntries;
     auto& failedNonces = ctx.failedNonces;
+    auto& failedNoncesLookup = ctx.failedNoncesLookup;
 
     if (!CustomMetadataParse(height, Params().GetConsensus(), metadata, txMessage)) {
         return false;
@@ -622,7 +624,8 @@ bool BlockAssembler::EvmTxPreapply(EvmTxPreApplyContext& ctx)
         }
         if (txResult.higher_nonce) {
             if (!failedTxSet.count(txIter)) {
-                failedNonces.emplace(txResult.nonce, txIter);
+                auto it = failedNonces.emplace(txResult.nonce, txIter);
+                failedNoncesLookup.emplace(txIter->GetTx().GetHash(), it);
             }
             return false;
         }
@@ -651,7 +654,8 @@ bool BlockAssembler::EvmTxPreapply(EvmTxPreApplyContext& ctx)
             return false;
         } else if (senderInfo.nonce > expectedNonce) {
             if (!failedTxSet.count(txIter)) {
-                failedNonces.emplace(senderInfo.nonce, txIter);
+                auto it = failedNonces.emplace(senderInfo.nonce, txIter);
+                failedNoncesLookup.emplace(txIter->GetTx().GetHash(), it);
             }
             return false;
         }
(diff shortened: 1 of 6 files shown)


test/functional/feature_evm_miner.py
```diff
@@ -490,20 +489,158 @@ def same_nonce_transferdomain_and_evm_txs(self):
...
+                        "dst": {
+                            "address": self.ethAddress,
+                            "amount": "1@DFI",
+                            "domain": 3,
+                        },
+                        "nonce": start_nonce_erc55 + i,
+                    }
+                ]
+            )
+        self.nodes[0].generate(1)
```
kuegi, line 606: to recreate the crash: do not mine a block, send tx with the reused nonce while first batch is still in the mempool

sieniven: This issue should not be related to mempool but on the miner side.

kuegi: > This issue should not be related to mempool but on the miner side.

could be, but on my side I can not reproduce it if the block is already there. only happens when there is no UTXO left in the address and I reuse the nonce.

kuegi:
FYI @sieniven on my side, if I change the part starting at line 518 to this, it fails (no block gets minted):

```

        for i in range(16):
            # tx call actual used gas - 1_761_626.
            # tx should always pass evm tx validation, but may return early in construct block.
            tx = contract.functions.loop(100_000).build_transaction(
                {
                    "chainId": self.nodes[0].w3.eth.chain_id,
                    "nonce": start_nonce + i,
                    "gasPrice": 25_000_000_000,
                    "gas": 30_000_000,
                }
            )
            signed = self.nodes[0].w3.eth.account.sign_transaction(tx, self.ethPrivKey)
            self.nodes[0].w3.eth.send_raw_transaction(signed.rawTransaction)

        amount= len(self.nodes[0].listunspent())+1
        for i in range(amount):
            self.nodes[0].transferdomain(
                [
                    {
                        "src": {
                            "address": self.address,
                            "amount": "1@DFI",
                            "domain": 2,
                        },
                        "dst": {
                            "address": self.ethAddress,
                            "amount": "1@DFI",
                            "domain": 3,
                        },
                        "nonce": start_nonce_erc55 + i,
                    }
                ]
            )

        start = self.nodes[0].getblockcount()
        self.nodes[0].generate(1)
        block_height = self.nodes[0].getblockcount()
        assert_equal(block_height, start + 1)
```

kuegi:
Seems to be some randomness to my test. it works 50% of the runs, others fail.

src/miner.cpp
```diff
@@ -873,12 +889,48 @@ void BlockAssembler::addPackageTxs(int& nPackagesSelected, int& nDescendantsUpda
...
             }
-            failedTxSet.insert(iter);
+
+            // Remove from checked TX set
+            for (const auto &entry : sortedEntries) {
+                checkedDfTxHashSet.erase(entry->GetTx().GetHash());
+            }
+
+            // Remove entries from queue if first EVM TX is not the failed TX.
+            if (sortedEntries.size() > 1) {
```
prasannavl, line 899: Needed?

prasannavl: Ok, never mind. Needed

src/miner.cpp
```diff
@@ -873,15 +888,56 @@ void BlockAssembler::addPackageTxs(int& nPackagesSelected, int& nDescendantsUpda
...
+            for (const auto& entry : sortedEntries) {
+                auto entryTxType = entry->GetCustomTxType();
+                auto entryHash = entry->GetTx().GetHash();
+
+                if (entryTxType == CustomTxType::EvmTx || entryTxType == CustomTxType::TransferDomain) {
+                    // If the first TX in a failed set is not the failed TX
+                    // then remove from queue, otherwise it has not been added.
+                    if (entryHash != failedCustomTx) {
+                        CrossBoundaryResult result;
+                        auto hashes = evm_unsafe_try_remove_txs_above_hash_in_q(result, evmQueueId, entry->GetTx().GetHash().ToString());
```
prasannavl, line 911:
```suggestion
                        auto hashes = evm_unsafe_try_remove_txs_above_hash_in_q(result, evmQueueId, entryHash.ToString());
```

src/miner.cpp
```diff
@@ -873,15 +888,56 @@ void BlockAssembler::addPackageTxs(int& nPackagesSelected, int& nDescendantsUpda
...
+                auto entryHash = entry->GetTx().GetHash();
+
+                if (entryTxType == CustomTxType::EvmTx || entryTxType == CustomTxType::TransferDomain) {
+                    // If the first TX in a failed set is not the failed TX
+                    // then remove from queue, otherwise it has not been added.
+                    if (entryHash != failedCustomTx) {
+                        CrossBoundaryResult result;
+                        auto hashes = evm_unsafe_try_remove_txs_above_hash_in_q(result, evmQueueId, entryHash.ToString());
+                        if (!result.ok) {
+                            LogPrintf("%s: Unable to remove %s from queue. Will result in a block hash mismatch.\n", __func__, entry->GetTx().GetHash().ToString());
```
prasannavl, line 913:
```suggestion
                            LogPrintf("%s: Unable to remove %s from queue. Will result in a block hash mismatch.\n", __func__, entryHash.ToString());
```

PR #2107, opened 2023-06-22
EVM Web3 RPC

Jouzo:
## Summary

- Adds web3 RPC namespace

## RPCs

- web3_sha3
- web3_clientVersion

PR #2482, opened 2023-09-22
chore: Read client version in `web_clientVersion` RPC from `cpp` ffi 

andyrobert3:
## Summary

- As per title, instead of reading client version from environment variable flags, we read it from `cpp` ffi  
- As per request from @prasannavl https://github.com/DeFiCh/ain/pull/2107#discussion_r1239250167 

## RPCs
- `web_clientVersion` 
image

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-cpp-imports/src/lib.rs
+++ b/lib/ain-cpp-imports/src/lib.rs
@@ -61,7 +61,7 @@ mod ffi {
     pub fn pastChangiIntermediateHeight3() -> bool {
         unimplemented!("{}", UNIMPL_MSG)
     }
-    pub fn getClientVersion() -> [u64; 3] {
+    pub fn getClientVersion() -> String {
         unimplemented("{}", UNIMPL_MSG)
     }
 }
@@ -71,9 +71,8 @@ pub fn get_chain_id() -> Result<u64, Box<dyn Error>> {
     Ok(chain_id)
 }
 
-pub fn get_client_version() -> Result<[u64; 3], Box<dyn Error>> {
-    let version = ffi::getClientVersion();
-    Ok(version)
+pub fn get_client_version() -> String {
+    ffi::getClientVersion()
 }
 
 pub fn is_mining() -> Result<bool, Box<dyn Error>> {
--- a/lib/ain-cpp-imports/src/lib.rs
+++ b/lib/ain-cpp-imports/src/lib.rs
@@ -61,11 +61,17 @@ mod ffi {
     pub fn pastChangiIntermediateHeight3() -> bool {
         unimplemented!("{}", UNIMPL_MSG)
     }
+    pub fn getClientVersion() -> [u64; 3] { unimplemented("{}", UNIMPL_MSG) }
 }
 
 pub fn get_chain_id() -> Result<u64, Box<dyn Error>> {
     let chain_id = ffi::getChainId();
     Ok(chain_id)
+}
+
+pub fn get_client_version() -> Result<[u64; 3], Box<dyn Error>> {
+    let version = ffi::getClientVersion();
+    Ok(version)
 }
 
 pub fn is_mining() -> Result<bool, Box<dyn Error>> {
--- a/lib/ain-cpp-imports/src/lib.rs
+++ b/lib/ain-cpp-imports/src/lib.rs
@@ -61,7 +61,9 @@ mod ffi {
     pub fn pastChangiIntermediateHeight3() -> bool {
         unimplemented!("{}", UNIMPL_MSG)
     }
-    pub fn getClientVersion() -> [u64; 3] { unimplemented("{}", UNIMPL_MSG) }
+    pub fn getClientVersion() -> [u64; 3] {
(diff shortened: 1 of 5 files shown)


lib/ain-cpp-imports/src/lib.rs
```diff
@@ -61,13 +61,21 @@ mod ffi {
...
+    }
 }
 
 pub fn get_chain_id() -> Result<u64, Box<dyn Error>> {
     let chain_id = ffi::getChainId();
     Ok(chain_id)
 }
 
+pub fn get_client_version() -> Result<[u64; 3], Box<dyn Error>> {
+    let version = ffi::getClientVersion();
```
Jouzo, line 75: Can return `ffi::getClientVersion()` without wrapping in a Result since it does not throw

src/ffi/ffiexports.cpp
```diff
@@ -208,6 +209,16 @@ int getHighestBlock() {
                             : (int) ::ChainActive().Height(); // return current block count if no peers
 }
 
+// Returns Major, Minor, Revision
+std::array<uint64_t, 3> getClientVersion() {
```
Jouzo, line 213: Couldn't we use existing `std::string FormatVersion(int nVersion)` here ?

lib/ain-cpp-imports/src/lib.rs
andyrobert3: Can this said of the same for the other functions in the `ffi` bridge such as `get_chain_id()`?

src/ffi/ffiexports.cpp
andyrobert3: woops, didn't see this XD

lib/ain-grpc/src/rpc/web3.rs
```diff
@@ -19,8 +19,9 @@ pub struct MetachainWeb3RPCModule {}
 
 impl MetachainWeb3RPCServer for MetachainWeb3RPCModule {
     fn client_version(&self) -> RpcResult<String> {
-        let version = env!("CARGO_PKG_VERSION");
-        let commit = option_env!("GIT_HASH").unwrap_or("unknown");
+        let version: String = ain_cpp_imports::get_client_version();
+        let commit = option_env!("GIT_HASH")
+            .ok_or_else(|| Error::Custom(format!("missing GIT_HASH env var")))?;
```
shohamc1, line 24: No point in throwing a runtime error on something that fails during compile time.

```suggestion
        let commit = env!("GIT_HASH", ""missing GIT_HASH env var");
```

shohamc1 (commented):
Wonder if we can just return the result of [`FormatVersionAndSuffix()`](https://github.com/DeFiCh/ain/blob/master/src/clientversion.cpp#L84) for this RPC.

lib/ain-grpc/src/rpc/web3.rs
```diff
@@ -19,8 +19,9 @@ pub struct MetachainWeb3RPCModule {}
 
 impl MetachainWeb3RPCServer for MetachainWeb3RPCModule {
     fn client_version(&self) -> RpcResult<String> {
-        let version = env!("CARGO_PKG_VERSION");
-        let commit = option_env!("GIT_HASH").unwrap_or("unknown");
+        let version: String = ain_cpp_imports::get_client_version();
+        let commit = option_env!("GIT_HASH")
```
sieniven, line 23: We can remove the need to get the hash from the local env during runtime, when we can directly get it from the compiled binary through the cpp side and expose it on the FFI as well.

andyrobert3: agree, good point 

lemme make a new PR for this

lib/ain-grpc/src/rpc/web3.rs
andyrobert3: as per @sieniven 's comment, i think we can expose the `GIT_HASH` via the ffi C++ bridge

lib/ain-grpc/src/rpc/web3.rs
sieniven: No worries I'm making the fixes directly, all good.

andyrobert3: any good first issues that i can look into :) 

@sieniven

lib/ain-grpc/src/rpc/web3.rs
sieniven: Resolved directly in PR #2107. web3_clientVersion RPC returns the format `Metachain/{client_version}/{target_os}-{target_arch}/{rustc_version}`.

lib/ain-cpp-imports/src/lib.rs
sieniven: @andyrobert3 Good spot. This should be a refactor after RC.

lib/ain-grpc/src/rpc/web3.rs
sieniven: Resolved directly on #2107.

PR #2488, opened 2023-09-25
evm: Use effective gas price for gas refunds

shohamc1:


Jouzo (approved):

lib/ain-evm/src/transaction/mod.rs
```diff
@@ -255,6 +255,15 @@ impl SignedTx {
         }
     }
 
+    pub fn effective_gas_price(&self, base_fee: U256) -> U256 {
+        match &self.transaction {
+            TransactionV2::Legacy(_) | TransactionV2::EIP2930(_) => self.gas_price(),
```
cuongquangnam, line 260: TransactionV2::EIP2930 ?

lib/ain-evm/src/core.rs
```diff
@@ -304,7 +304,8 @@ impl EVMCoreService {
 
             // Execute tx
             let mut executor = AinExecutor::new(&mut backend);
-            let (tx_response, ..) = executor.exec(&signed_tx, signed_tx.gas_limit(), prepay_fee);
```
sieniven, line 307: Should not be None as this is in validate evm tx pipeline

lib/ain-evm/src/executor.rs
```diff
@@ -187,7 +188,10 @@ impl<'backend> AinExecutor<'backend> {
                 signed_tx.sender,
                 signed_tx.gas_limit(),
                 U256::from(used_gas),
-                signed_tx.gas_price(),
+                match base_fee {
+                    Some(base_fee) => signed_tx.effective_gas_price(base_fee),
```
sieniven, line 192: We should be refactoring refund_unused_gas total gas fee calculation to use the 'calculate_gas_fee' helper function in fee lib.

PR #2496, opened 2023-09-26
evm: Return effective gas price in `eth_getTransaction`

shohamc1:


Jouzo (dismissed):

cuongquangnam:
Should we fix other endpoints that are related to blocks too?
Btw, just thinking whether we can fix the root cause which is EthTransactionInfo?

shohamc1:
What is the issue with `eth_getBlockByHash` and `eth_getBlockByNumber`? What is the problem with `EthTransactionInfo`?

cuongquangnam:
in eth_getBlockByHash there is an option to ask for full-transactions, which will return ethTransactionInfo later
reference: 
https://github.com/DeFiCh/ain/blob/ba17cf1d3ff50749e12bbb5288c868d0daae3867/lib/ain-grpc/src/rpc/eth.rs#L442

shohamc1:
Ran through the pipeline and found one bug with the RPCs, fixed in 97f298f4fdba4e6ac5bbcaf4f1d465f2ac8d08a6. Is this the bug you were talking about, or is there something else wrong?

PR #2498, opened 2023-09-26
Get gas fee in mempool and skip on mining if over block limit

Bushstar:
## Summary

- When entering the mempool get the gas cost of the EVM TXs
- Skip TXs if they will go over the block gas limit
- Moves some performance cost to the mempool, but should prevent miners becoming unresponsive when the mempool as many TXs with high gas fees.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-rs-exports/src/evm.rs
+++ b/lib/ain-rs-exports/src/evm.rs
@@ -1187,6 +1187,17 @@ pub fn evm_try_get_tx_sender_info_from_raw_tx(
     )
 }
 
+pub fn evm_try_get_total_gas_used(result: &mut ffi::CrossBoundaryResult, queue_id: u64) -> String {
+    unsafe { cross_boundary_success_return(result, SERVICES.evm.core.get_total_gas_used(queue_id)) }
+}
+
+pub fn evm_try_get_block_limit(result: &mut ffi::CrossBoundaryResult) -> u64 {
+    match SERVICES.evm.get_block_limit() {
+        Ok(block_limit) => cross_boundary_success_return(result, block_limit),
+        Err(e) => cross_boundary_error_return(result, e.to_string()),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     #[test]
--- a/lib/ain-rs-exports/src/evm.rs
+++ b/lib/ain-rs-exports/src/evm.rs
@@ -1191,7 +1191,7 @@ pub fn evm_try_get_tx_info_from_raw_tx(
         caller: Some(signed_tx.sender),
         to: signed_tx.to(),
         value: signed_tx.value(),
-        data: &signed_tx.data(),
+        data: signed_tx.data(),
         gas_limit: u64::try_from(signed_tx.gas_limit()).unwrap_or(u64::MAX),
         gas_price: Some(signed_tx.effective_gas_price(base_fee)),
         max_fee_per_gas: signed_tx.max_fee_per_gas(),
--- a/lib/ain-rs-exports/src/evm.rs
+++ b/lib/ain-rs-exports/src/evm.rs
@@ -1,7 +1,8 @@
 use ain_contracts::{get_transfer_domain_contract, FixedContract};
 use ain_evm::{
-    core::{ValidateTxInfo, XHash},
+    core::{EthCallArgs, ValidateTxInfo, XHash},
     evm::FinalizedBlockInfo,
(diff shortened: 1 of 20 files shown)


lib/ain-rs-exports/src/evm.rs
```diff
@@ -1187,16 +1187,43 @@ pub fn evm_try_get_tx_sender_info_from_raw_tx(
         return cross_boundary_error_return(result, "prepay fee value overflow");
     };
 
+    let Ok(TxResponse { used_gas, .. }) = SERVICES.evm.core.call(EthCallArgs {
```
sieniven, line 1190: Note that the gas used calculated here is based on the state of the current tip, which means that we cannot make the assumption that this gas used is actually correct. The only pipeline where we will be able to correctly get the gas used calculation is only on validate evm tx pipeline.

sieniven (approved):

lib/ain-rs-exports/src/evm.rs
Bushstar: We do execute the TX to get the gas, it should be approximately correct. 

Problem we were having is the miners becoming unresponsive while creating blocks due to lots of large fee TXs in the mempool. The mempool was spammed with thousands of 10m fee TXs which took down the miner nodes. This PR prevents those TXs from being processed after the block is full, it does not need to be perfect.

Currently lots of 10M fee TXs act like a DoS on staking nodes.

src/miner.cpp
```diff
@@ -619,6 +626,10 @@ bool BlockAssembler::EvmTxPreapply(EvmTxPreApplyContext& ctx)
         return false;
     }
 
+    if (blockGasUsed + sortedGasUsed + txIter->GetEVMGasUsed() > blockGasLimit) {
```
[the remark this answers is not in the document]
Bushstar: We could try to add a TX that does not fit or skip one that could have, but yes, it is a trade off to protect the stakers.

lib/ain-rs-exports/src/evm.rs
sieniven: `We do execute the TX to get the gas, it should be approximately correct.`
This might not be true. A evm tx that executes with an early return could give a vastly different gas used from an evm tx that executes fully.

sieniven: But the assumption should be good enough to protect the minter from DoS attack, while trading off some performance.

src/validation.cpp
```diff
@@ -943,6 +944,16 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
...
+            // Block gas limit
+            auto blockGasLimit = evm_try_get_block_limit(result);
+            if (!result.ok) {
+                blockGasLimit = DEFAULT_EVM_BLOCK_GAS_LIMIT;
+            }
+
+            if (txResult.used_gas > blockGasLimit) {
+                return state.Invalid(ValidationInvalidReason::TX_NOT_STANDARD, error("evm tx over block gas limit. Gas used: %d Limit: %d", txResult.used_gas, blockGasLimit), REJECT_INVALID, "evm-over-limit");
+            }
+
```
sieniven, line 956: This check is not required since it would have already been caught in pre-validation pipeline. Used gas will never exceed gas limit specified in the evm tx: https://github.com/DeFiCh/ain/blob/master/lib/ain-evm/src/core.rs#L267-L273.

src/validation.cpp
```diff
@@ -938,17 +939,28 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
...
             EvmAddressWithNonce evmAddrAndNonce{txResult.nonce, txResult.address.c_str()};
 
             const auto prePayFee = isEVMTx ? txResult.prepay_fee : std::numeric_limits<uint64_t>::max();
 
             entry.SetEVMAddrAndNonce(evmAddrAndNonce);
             entry.SetEVMPrePayFee(prePayFee);
+            entry.SetEVMGasUsed(txResult.used_gas);
```
sieniven, line 963: txResult for the transferdomain transparent evm tx should be set to zero.

src/validation.cpp
Bushstar: Thanks.

sieniven (dismissed):

PR #2500, opened 2023-09-26
EVM: Cache ValidateTxInfo

Jouzo:
## Summary

- TxValidationCache implementation
- Cache stateless/stateful TX validation


lib/ain-evm/src/evm.rs
```diff
@@ -396,6 +394,7 @@ impl EVMServices {
         }
         self.core.tx_queues.remove(queue_id);
         self.core.clear_account_nonce();
+        self.core.clear_transaction_cache();
```
sieniven, line 397: Should we clear the transaction cache on disconnect block as well?

sieniven (approved):

sieniven (approved):

PR #2501, opened 2023-09-27
evm: EIP1559 fees tests

shohamc1:


sieniven (commented):
Any reason to not put these tests into feature_evm_fee instead?

shohamc1:
> Any reason to not put these tests into feature_evm_fee instead?

`feature_evm_fee` is being used for legacy transactions, this is for EIP1559 transactions,

sieniven (approved):

sieniven (approved):

PR #2521, opened 2023-10-02
Set correct timestamp in queue backend

Jouzo:
## Summary

- Pass down timestamp when creating EVM queue.
- Use timestamp as backend vicinity to use set timestamp as in construct_block
- Some smart contract are using https://github.com/DeFiPages/Decentralized-Crypto-Casino/blob/main/src/backend/contracts/Roulette.sol timestamp as randomness and this leads to different result/gas usage between TX applied in queue and in construct_block. 
- Fixes TXs going above block gas limit. As we were only checking TXs total gas used while queueing the TXs, it was within bounds in queue but going randomly above while constructing block.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2369,7 +2369,7 @@ static void LogApplyCustomTx(const CTransaction &tx, const int64_t start) {
  *  Validity checks that depend on the UTXO set are also done; ConnectBlock ()
  *  can fail if those validity checks fail (among other reasons). */
 bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pindex,
-                  CCoinsViewCache& view, CCustomCSView& mnview, const CChainParams& chainparams, bool & rewardedAnchors, uint64_t evmQueueId, bool fJustCheck)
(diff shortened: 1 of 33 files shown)


sieniven (dismissed):

src/validation.cpp
```diff
@@ -2650,6 +2649,13 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
     const auto consensus = chainparams.GetConsensus();
     auto isEvmEnabledForBlock = IsEVMEnabled(pindex->nHeight, mnview, consensus);
 
+    uint64_t evmQueueId{};
```
sieniven, line 2652: With queue creation moved to inside ConnectBlock from ConnectTip how do we delete the queue if ConnectBlock returns early with a failure? We may end up with dangling queues. cc/ @Bushstar

Bushstar: Thanks. This was the reason for having the queue in ConnectTip, so if ConnectBlock returned early we could clean up the queue as construct block would be skipped at the end of ConnectBlock.

sieniven: Connect block changes are reverted commit 037d513

sieniven (approved):

Bushstar (approved):

PR #2526, opened 2023-10-03
EVM: Clean up construct block validation

sieniven:
## Summary

- This PR **depends on PR #2521 and PR #2516** to be merged in first.
- Adds additional layer of check to ensure that block gas limit is not exceeded when executing txs in the txqueue.
- This PR safeguards and removes the assumption that the state should match on consensus when adding txs to the queue vs the state on actual construct block
- EVM txs are now allowed to fail if they exceed block gas limit in construct block pipeline on the miner, which will subsequently be removed on the block template on the DVM miner side.
- On connect block, EVM txs should never fail from block gas limit now. Thus block will be invalidated if any EVM tx fails on construct block pipeline.
- Refactor block gas limit check into executor

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


src/miner.cpp
```diff
@@ -270,13 +270,43 @@ ResVal<std::unique_ptr<CBlockTemplate>> BlockAssembler::CreateNewBlock(const CSc
...
+                }
 
+                // Get all EVM Txs
+                CTxMemPool::setEntries failedEVMTxs;
+                for (const auto& iter : inBlock) {
+                    auto tx = iter->GetTx();
+                    if (!evmTxsToUndo.count(tx.GetHash()))
+                        continue;
+                    std::vector<unsigned char> metadata;
+                    const auto txType = GuessCustomTxType(tx, metadata, false);
```
Bushstar, line 297: From the iter you can get the custom TX type.

src/miner.cpp
```diff
@@ -270,13 +270,43 @@ ResVal<std::unique_ptr<CBlockTemplate>> BlockAssembler::CreateNewBlock(const CSc
...
+                    if (txType == CustomTxType::EvmTx) {
```
Bushstar, line 298: What about TransferDomain TXs?

src/validation.cpp
```diff
@@ -2650,6 +2649,13 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
     const auto consensus = chainparams.GetConsensus();
     auto isEvmEnabledForBlock = IsEVMEnabled(pindex->nHeight, mnview, consensus);
 
+    uint64_t evmQueueId{};
+    if (isEvmEnabledForBlock) {
+        auto r = XResultValueLogged(evm_try_unsafe_create_queue(result, pindex->GetBlockTime()));
```
Bushstar, line 2654: With queue creation moved to inside ConnectBlock from ConnectTip how do we delete the queue if ConnectBlock returns early with a failure? We may end up with dangling queues.

sieniven: These changes are in PR #2521. Let me shift these into that PR instead.

src/miner.cpp
```diff
@@ -270,13 +270,43 @@ ResVal<std::unique_ptr<CBlockTemplate>> BlockAssembler::CreateNewBlock(const CSc
...
+                        LogPrintf("%s: Unable to remove from block, not EVM tx. Will result in a block hash mismatch.\n", __func__);
+                    }
+                }
+                RemoveFromBlock(failedEVMTxs, true);
+            }
+        }
+
+        auto r = XResultStatusLogged(evm_try_unsafe_remove_queue(result, evmQueueId));
+        if (!r) return Res::Err("Failed to remove queue");
         xvm = XVM{0, {0, std::string(blockResult.block_hash.data(), blockResult.block_hash.length()).substr(2), blockResult.total_burnt_fees, blockResult.total_priority_fees, evmBeneficiary}};
```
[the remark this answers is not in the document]
sieniven: These changes are in PR https://github.com/DeFiCh/ain/pull/2521. Let me shift these into that PR instead.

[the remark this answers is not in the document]
sieniven: This is intended. In consensus when adding txs, we should not guarantee or make the assumption that the state of the txqueue vs the state in construct block is the same. Thus, we should not ever reach this pipeline, but if we do, we guarantee that blocks created are strictly below block gas limit.

src/miner.cpp
sieniven: Only evmtxs have gas usage on the executor. This is guarded with the system_tx flag that is passed into the executor: https://github.com/DeFiCh/ain/pull/2526/files#diff-6f8e31143f7885918b267c7d971ca0193faeba1b27ef2229f2c8ee6ec4889336R190.

sieniven: We should not ever go into this loop: https://github.com/DeFiCh/ain/pull/2526/files#diff-6bf9d2e37d503f2f3381cd3a41c26e0b9670f26c2cd4571e98d2af9e3767c1a1R298-R300, and we should not be seeing theis log message.

src/miner.cpp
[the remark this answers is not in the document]
Bushstar: Then there is no point in removing TXs from the block as the block will be rejected as soon as you do due to EVM hash mismatch,

src/miner.cpp
Bushstar: evm_try_unsafe_remove_txs_above_hash_in_q will remove TD TXs from the queue. If you could regenerate the block hash for the queue after removal for the EVM hash it would not match due to the block still having the TD TXs in.

All we are going to do here is set the miner up to create invalid blocks in a loop.

sieniven: This PR should not do that. To summarize, the pipeline is:
1. consensus builds the txqueue
2. construct block is called -> we do a final check here to ensure block doesnt exceed limit
3. if it somehow does, this safety layer catches it and fails that tx (and everything else that comes after)
4. kick out the failed evm txs from dvm block template

Transferdomain txs uses zero gas so that will never ever enter this safety net pipeline. EVM txs are never committed on the EVM block when they fail to add into the block due to the block size limit, and are returned as failed txs. Failed txs are subsequently removed from the block template on the miner, and will be included in the next block.

No invalid blocks should be minted since all txs are validated through consensus pipeline. EVM block hash mismatch will not happen either as long as we ensure proper handling of the txs to match correctly in both the DVM and EVM block.

src/miner.cpp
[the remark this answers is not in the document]
sieniven: https://github.com/DeFiCh/ain/pull/2526#discussion_r1344447136 should resolve this as well.

Bushstar (dismissed):

Bushstar (approved):

PR #2527, opened 2023-10-03
Cargo tweaks for CI

prasannavl:
## Summary

- Adds cargo caches on the CI to speed up deps compilations expect on staging or release builds. 
- Disable incremental cargo builds on CI to avoid unnecessary data. 
- Reduces rust debuginfo level to 1, as that's sufficient for most use cases, but will considerably reduce binary size

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -12,6 +12,9 @@ rustflags = [
   # "-C", "force-unwind-tables=y",
 ]
 
+[profile.dev]
+debug = 1
+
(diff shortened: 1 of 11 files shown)


prasannavl:
Edit: This won't work on docker builds.

Related: https://github.com/DeFiCh/ain/pull/2114

prasannavl:
Full caching to be addressed later in https://github.com/DeFiCh/ain/pull/2114. This caches non dockerized builds until then.

PR #2014, opened 2023-05-25
EVM Pub/Sub

Jouzo:
## Summary

- Adds websocket server
- Adds `eth_subscribe` and `eth_unsubscribe` RPC calls
- Supports `syncing`, `newHeads`, `logs` and `newPendingTransactions` subscriptions. Reference : https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub

## RPCs

To be used via websocket connection :

Syncing call :
```
{"id": 1, "jsonrpc": "2.0", "method": "eth_subscribe", "params": ["syncing"]}
```
outputs :
```
{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":"0xe743129ac2b693fa07228ad2e51460fd","result":{"syncing":true,"startingBlock":3976,"currentBlock":3810,"highestBlock":3809}}}
```

NewHeads call :
```
{"id": 1, "jsonrpc": "2.0", "method": "eth_subscribe", "params": ["newHeads"]}
```
outouts:
```
[...]

PR #2528, opened 2023-10-04
evm: Subscriptions

shohamc1:
Supersedes #2014

### DIFF
--- a/lib/ain-evm/src/evm.rs
+++ b/lib/ain-evm/src/evm.rs
@@ -36,6 +36,11 @@ use crate::{
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 use tokio::sync::RwLock;
 
+pub struct NotificationChannel<T> {
+    pub sender: UnboundedSender<T>,
+    pub receiver: RwLock<UnboundedReceiver<T>>,
+}
+
 pub struct EVMServices {
     pub core: EVMCoreService,
     pub block: BlockService,
@@ -43,10 +48,7 @@ pub struct EVMServices {
     pub logs: LogService,
     pub filters: FilterService,
     pub storage: Arc<Storage>,
-    pub channel: (
-        UnboundedSender<Notification>,
-        RwLock<UnboundedReceiver<Notification>>,
-    ),
+    pub channel: NotificationChannel<Notification>,
 }
 
 pub struct FinalizedBlockInfo {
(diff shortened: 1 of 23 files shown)


Jouzo (approved):

lib/ain-evm/src/evm.rs
```diff
@@ -418,6 +436,11 @@ impl EVMServices {
                 .generate_logs_from_receipts(&receipts, block.header.number)?;
             self.receipt.put_receipts(receipts)?;
             self.filters.add_block_to_filters(block.header.hash());
+
+            self.channel
+                .sender
+                .send(Notification::Block(block.header.hash()))
+                .map_err(|e| format_err!(e.to_string()))?;
```
sieniven, line 443: This is an incorrect place to notify that a new block is created. This is still in consensus, and there could be downstream failures that may potentially happen. We should expose the send notification as an FFI, and call it at the end of connectblock pipeline on the CPP side.

lib/ain-evm/src/evm.rs
```diff
@@ -512,6 +535,11 @@ impl EVMServices {
 
         self.filters.add_tx_to_filters(signed_tx.transaction.hash());
 
+        self.channel
+            .sender
+            .send(Notification::Transaction(signed_tx.transaction.hash()))
+            .map_err(|e| format_err!(e.to_string()))?;
+
```
sieniven, line 542: Likewise for here. At this point, it could be called on the miner or on connectblock when an incoming block is still be validated. Txs pushed into the queue at this point havent been executed yet on construct block and it is an incorrect event trigger to notify subscribers of the signed tx being minted.

sieniven: We should expose an FFI at the end of connect block that then iterates through the block and publishes the txs inside the block.

sieniven: Pending transactions should be obtained from the mempool instead. It should be inside the mempool pipeline, and we expose an FFI call for this

shohamc1: Fixed: cbc5cab

src/validation.cpp
```diff
@@ -966,6 +966,12 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
             } else {
                 ethSender = txResultSender;
             }
+
+
+            evm_try_dispatch_pending_transactions_event(result, rawEVMTx);
+            if (!result.ok) {
+                return state.Invalid(ValidationInvalidReason::CONSENSUS, error("evm tx failed to generate events %s", result.reason.c_str()), REJECT_INVALID, "evm-events");
```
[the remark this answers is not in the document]
shohamc1: Fixed: https://github.com/DeFiCh/ain/pull/2528/commits/387c653f1c314726f4e4286ef322362ec1c324ef

sieniven (approved):

Jouzo (approved):

PR #2533, opened 2023-10-04
Remove redundant validation checks

Jouzo:
## Summary

- `evm_try_unsafe_validate_raw_tx_in_q` has seen some major overhaul recently. Since then, it's only checking for nonce and total_gas_used. These checks are also part of `evm_try_unsafe_push_tx_in_q`.
Both these calls are called one after the other and executing the same checks. TX validation is now handled only by `evm_try_unsafe_push_tx_in_q`.
- Port TD DST20 nonce check to `apply_queue_tx`.
- Remove `evm_try_unsafe_validate_raw_tx_in_q` which was always called before `evm_try_unsafe_push_tx_in_q`.
- Non-consensus since this PR only removes redundant checks. 

- Depends on https://github.com/DeFiCh/ain/pull/2532

### DIFF
--- a/lib/ain-evm/src/core.rs
+++ b/lib/ain-evm/src/core.rs
@@ -307,7 +307,7 @@ impl EVMCoreService {
 
     /// Validates a raw tx.
     ///
-    /// The pre-validation checks of the tx before we consider it to be valid are:
+    /// The validation checks of the tx before we consider it to be valid are:
     /// 1. Gas price check: verify that the maximum gas price is minimally of the block initial base fee.
     /// 2. Gas price and tx value check: verify that amount is within money range.
     /// 3. Intrinsic gas limit check: verify that the tx intrinsic gas is within the tx gas limit.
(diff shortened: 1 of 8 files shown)


lib/ain-evm/src/core.rs
```diff
@@ -337,32 +336,24 @@ impl EVMCoreService {
...
     /// across all usages. Note: To be replaced with a proper lock flow later.
     ///
-    pub unsafe fn validate_raw_tx(
-        &self,
-        tx: &str,
-        queue_id: u64,
-        pre_validate: bool,
-        block_fee: U256,
-    ) -> Result<ValidateTxInfo> {
+    pub unsafe fn prevalidate_raw_tx(&self, tx: &str, queue_id: u64) -> Result<ValidateTxInfo> {
```
sieniven, line 339: Instead of changing the naming convention, can we just call this validate_raw_tx to maintain consistency with the validate_transferdomain_raw_tx?

sieniven (dismissed):

lib/ain-evm/src/executor.rs
```diff
@@ -349,6 +349,17 @@ impl<'backend> AinExecutor<'backend> {
...
+                let nonce = self.backend.get_nonce(&signed_tx.sender);
+                if nonce != signed_tx.nonce() {
+                    return Err(format_err!(
+                        "[apply_queue_tx] nonce check failed. Account nonce {}, signed_tx nonce {}",
+                        nonce,
+                        signed_tx.nonce(),
+                    )
+                    .into());
+                }
+
```
sieniven, line 362: Is this nonce check necessary? Since it is already handled in exec pipeline when we actually execute the transferdomain tx itself. Also, we dont have to worry about handling BlockSizeLimit EVM error since system tx flag is passed for DST20 bridging.

lib/ain-evm/src/core.rs
sieniven: Resolved in [95c4c7f](https://github.com/DeFiCh/ain/pull/2533/commits/95c4c7f0c7ab57f42974defbc87c3a49082f7437)

Bushstar (approved):

PR #2537, opened 2023-10-05
Scoped EVM queue ID class

Bushstar:
## Summary

- Creates a class that will create and destroy an EVM queue ID
- Only create a queue if EVM is enabled, allows the replacement, where applicable, of the isEvmEnabledForBlock check with a valid EVM queue ID check

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


prasannavl:
@Bushstar nice work. I think we can do better though. 

I think this might have complicated things a bit with isValid, as well as checking the count manually; and what happens when the count isn't. 

However, this might be perfect for shared_ptr itself, who's whole goal is to keep track of these and destroy when everything goes out of scope automatically. By wrapping it and adding a destructor on top, I think we've actually increased complexity than simplify it. 

Rather, how about, create a simple scope wrapper, who's destructor will always destroy the queue. Now, we keep track of that in the shared_ptr and pass the shared_ptr around. So, it has to be:

shared_ptr<ScopedCQueue> -- this automatically reference tracks the queue, and destroys it when all references go out of scope.

sieniven (approved):

PR #2538, opened 2023-10-05
EVM: Mempool improvements to RBF strategy and housekeeping

sieniven:
### Mempool Improvements
- PR aligns the node's mempool replace-by-fee strategy with [Ethereum standards](https://github.com/ethereum/go-ethereum/issues/23616)
- The conditions for replace-by-fee in the mempool will now **require the sender to bump the promise tip fees by minimally 10%, and the increase in promised tip fee per gas has to be of minimally 1 gwei**.
- The minimum RBF promised tip fees calculation details:
   - For legacy / EIP2930 txs: **(oldGasPrice - INITIAL_BLOCK_FEES) * tx_gas_limit * 1.10**
   - For EIP1559 txs: **oldMaxPriorityFeePerGas * tx_gas_limit * 1.10**
- Remove RBF of transferdomain EVM txs.

### Housekeeping for RPCs 
- Fixes FFI calls used for miner and RPCs to get values in satoshis as FP for higher precision
- Fixes reduces the loss in precision when getting EVM tx attributes such as gas prices which may be specified in values less than 1 satoshi

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -1179,7 +1179,7 @@ bool CTxMemPool::getAccountViewDirty() const {
     return accountsViewDirty;
 }
 
-bool CTxMemPool::checkAddressNonceAndFee(const CTxMemPoolEntry &pendingEntry, const EvmAddressData &txSender, bool &senderLimitFlag) {
+bool CTxMemPool::checkAddressNonceAndFee(const CTxMemPoolEntry &pendingEntry, const uint64_t &entryFee, const EvmAddressData &txSender, bool &senderLimitFlag) {
     if (pendingEntry.GetCustomTxType() != CustomTxType::EvmTx &&
         pendingEntry.GetCustomTxType() != CustomTxType::TransferDomain) {
         return true;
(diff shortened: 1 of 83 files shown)


prasannavl:
This is great. Love the changes overall, but not convinced on the need for floating point.

src/txmempool.cpp
```diff
@@ -1197,32 +1197,14 @@ bool CTxMemPool::checkAddressNonceAndFee(const CTxMemPoolEntry &pendingEntry, co
...
-                    }
-                }
-                txsToRemove.insert(tx);
-            }
+        const auto entryType = entry.GetCustomTxType();
+        if (entryType == CustomTxType::EvmTx && entryFee >= entry.GetEVMRbfMinTipFee()) {
+            auto txIter = mapTx.project<0>(it);
+            itersToRemove.insert(txIter);
 
             // We might want to set accountsViewDirty to true here but
```
Bushstar, line 1209: We can remove this comment now, it is redundant since we will not be removing TransferDomain TXs. The account view will not be affected by EVM removal.

Bushstar (approved):

PR #2544, opened 2023-10-09
Cache Keccak Hashes per SignedTx

prasannavl:
## Summary

- Cache tx hash computation, as the pipeline computes hash multiple times and Keccak hashes make for the most expensive computation in the current pipeline. 

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


lib/ain-evm/src/transaction/mod.rs
```diff
@@ -179,6 +180,7 @@ impl TryFrom<TransactionV2> for SignedTx {
         Ok(SignedTx {
             transaction: src,
             sender: public_key_to_address(&pubkey),
+            hash_cache: Cell::new(None),
         })
```
Jouzo, line 184:
```suggestion
        Ok(SignedTx {
            hash: src.hash(),
            transaction: src,
            sender: public_key_to_address(&pubkey),
        })
```

lib/ain-evm/src/transaction/mod.rs
```diff
@@ -100,6 +100,7 @@ impl From<&LegacyTransaction> for LegacyUnsignedTransaction {
 pub struct SignedTx {
     pub transaction: TransactionV2,
     pub sender: H160,
+    hash_cache: Cell<Option<H256>>,
```
Jouzo, line 103:
```suggestion
    hash: H256,
```

lib/ain-evm/src/transaction/mod.rs
```diff
@@ -317,11 +319,16 @@ impl SignedTx {
...
+        let h = &self.hash_cache;
+        if h.get().is_none() {
+            let val = match &self.transaction {
+                TransactionV2::Legacy(tx) => tx.hash(),
+                TransactionV2::EIP2930(tx) => tx.hash(),
+                TransactionV2::EIP1559(tx) => tx.hash(),
+            };
+            h.set(Some(val));
         }
+        h.get().unwrap()
```
Jouzo, line 331:
```suggestion
       self.hash
```

Jouzo (commented):
Missed the pipeline where we never use the hash(). Reverted the lazy changes

PR #2546, opened 2023-10-09
evm: Do not recompute block hashes

shohamc1:


Jouzo:
Handled via https://github.com/DeFiCh/ethereum/pull/1

PR #2563, opened 2023-10-10
Add block and transaction context

Bushstar:
## Summary

- Add related function argument to a context to reduce the number of arguments passed through various pipelines.

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


src/dfi/consensus/txvisitor.cpp
```diff
@@ -96,19 +97,17 @@ CCustomTxVisitor::CCustomTxVisitor(const CTransaction &tx,
...
       coins(coins),
       consensus(consensus),
       time(time),
       txn(txn),
-      evmQueueId(evmQueueId),
-      isEvmEnabledForBlock(isEvmEnabledForBlock),
-      evmPreValidate(evmPreValidate) {}
+      evmQueueId(blockCtx.evmQueueId),
+      isEvmEnabledForBlock(blockCtx.isEvmEnabledForBlock),
+      evmPreValidate(blockCtx.evmPreValidate) {}
```
prasannavl, line 110: evmPrevalidate would be tx context, and not block context?

Bushstar: At what point do we switch to a transaction context?

PR #2565, opened 2023-10-11
Check ATTRIBUTE export keys. Do not import empty set.

Bushstar:
## Summary

- Do not allow setting of an empty attribute map. It is possible to send an ATTRIBUTE Gov var TX for a yet to be defined value as Export without knowing that value will export an empty ATTRIBUTE map which will be accepted by Import and will be accepted. This PR rejects empty ATTRIBUTE imports.
- Before exporting ATTRIBUTES to import check the keys are valid as invalid keys are currently ignored when they should generate an error.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] Network upgrade required (rollback on testnet)
  - [x] Includes backward compatible changes (maybe for mainnet)
  - [ ] None


prasannavl (approved):

prasannavl:
Full sync tests: 

Fork guarded, but affected blocks: 

- Testnet:
  - [1366006](https://defiscan.live/blocks/1366006?network=TestNet)
  - [1850197](https://defiscan.live/blocks/1850197?network=TestNet) [Original culprit] 
  
- Mainnet - Sync in progress.

PR #2579, opened 2023-10-13
evm: Use checked operations

shohamc1:


lib/ain-evm/src/block.rs
```diff
@@ -80,30 +80,30 @@ impl BlockService {
...
                     gas_used_delta,
                     parent_gas_target,
                     parent_base_fee,
                     base_fee_max_change_denominator,
                 )?;
-                Ok(max(
-                    parent_base_fee + base_fee_per_gas_delta,
-                    initial_base_fee,
-                ))
+                Ok(match parent_base_fee.checked_add(base_fee_per_gas_delta) {
```
prasannavl, line 90: Can do a saturating_add, but do we really have to? Why don't we just return error. Simplifies it and makes sure nothing on the chain can result in overflow/underflows.

lib/ain-evm/src/block.rs
```diff
@@ -80,30 +80,30 @@ impl BlockService {
...
                     gas_used_delta,
                     parent_gas_target,
                     parent_base_fee,
                     base_fee_max_change_denominator,
                 )?;
-                Ok(max(
-                    parent_base_fee - base_fee_per_gas_delta,
-                    initial_base_fee,
-                ))
+                Ok(match parent_base_fee.checked_sub(base_fee_per_gas_delta) {
```
prasannavl, line 103: Same as above.

lib/ain-evm/src/block.rs
shohamc1: Throwing an error would mean that block production would stop for a while until a sufficiently full/empty block is mined. Doing the `saturating_add` would allow the block production to remain unaffected while respecting U256 bounds.

prasannavl: We're at the edge case where base fee jumps so high to overflow these numbers. If we're at the point, we likely will have to dealing with economical decision well before. I'd suggest just error-ing out and let things cool down.

prasannavl: Reason being: whatever we do post this stage, error-ing out will make for better options, than saturating that will change fee characteristic.

prasannavl: Max base fee is a good alternative for now. Looks good.

prasannavl (approved):

PR #2592, opened 2023-10-17
Fix mempool accounts view state mismatch

sieniven:
## Summary

- Remove accounts view rebuild pipeline on mempool reorg
- Add graceful fallbacks to create new block pipeline error failure - on first failure, mempool will do a rebuild of the accounts view state, and on second failure clear the mempool to prevent the miner from being stuck in a loop creating invalid blocks.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -1145,7 +1145,7 @@ void Staker::withSearchInterval(F&& f, int64_t height)
 
 void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams chainparams)
 {
-    std::set<CKeyID> nClearFlag;
+    std::map<CKeyID, uint32_t> nClearFlag;
     std::map<CKeyID, int32_t> nMinted;
     std::map<CKeyID, int32_t> nTried;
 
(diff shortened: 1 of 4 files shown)


src/miner.cpp
```diff
@@ -1214,9 +1215,15 @@ void ThreadStaker::operator()(std::vector<ThreadStaker::Args> args, CChainParams
             } catch (const std::runtime_error& e) {
                 LogPrintf("ThreadStaker: (%s) runtime error: %s\n", e.what(), operatorName);
 
-                // Could be failed TX in mempool, wipe mempool and allow loop to continue.
-                LOCK(cs_main);
-                mempool.clear();
+                if (nClearFlag.find(arg.operatorID) == nClearFlag.end()) {
```
prasannavl, line 1218: How about a map of CKeyId and int for consecutive failures? It's easy to track how many failure this way. 

Note: Also, reset to 0 on successful block.

src/txmempool.cpp
```diff
@@ -658,6 +645,19 @@ void CTxMemPool::removeForBlock(const std::vector<CTransactionRef>& vtx, unsigne
     blockSinceLastRollingFeeBump = true;
 }
 
+void CTxMemPool::rebuildAccountsView() {
+    if (pcustomcsview) {
```
prasannavl, line 649: Can flip condition to un-nest

src/txmempool.cpp
```diff
@@ -658,6 +645,19 @@ void CTxMemPool::removeForBlock(const std::vector<CTransactionRef>& vtx, unsigne
     blockSinceLastRollingFeeBump = true;
 }
 
+void CTxMemPool::rebuildAccountsView() {
```
prasannavl, line 648: Perhaps call this as rebuildCustomCSView to avoid confusion?

prasannavl (dismissed):

PR #2603, opened 2023-10-20
Remove BlockTemplateMap

Jouzo:
## Summary

- Remove blockTemplateId and return a ptr to the template itself.
- Remove need for locks and sync primitives


Jouzo:
Switching to draft as EVM e2e fail on CI, needs  further investigation

sieniven (dismissed):

PR #2611, opened 2023-10-23
EVM backend overlay

Jouzo:
## Summary

- Keeps track of block changes in in-memory overlay layer
- Keeps `EVMBackend` in `BlockTemplate`
- Commit backend a single time on `construct_block`
- Remove cache by state root
- Keep vector of changeset to be able to revert to an overlay state at a particular TX (used by `remove_txs_above_hash`)


lib/ain-evm/src/backend.rs
```diff
@@ -62,9 +141,14 @@ impl EVMBackend {
             trie_store,
             storage,
             vicinity,
+            overlay: Overlay::new(),
         })
     }
 
+    pub fn increase_tx_count(&mut self) {
```
sieniven, line 148: Will renaming this method to something like cache_state make it more readable?

lib/ain-evm/src/backend.rs
```diff
@@ -38,11 +42,86 @@ pub struct Vicinity {
...
+        code: Option<Vec<u8>>,
+        mut storage: HashMap<H256, H256>,
+        reset_storage: bool,
+    ) {
+        if !reset_storage {
+            if let Some(existing_storage) = self.storage(&address) {
+                for (k, v) in existing_storage {
+                    storage.entry(*k).or_insert_with(|| *v);
+                }
+            }
```
sieniven, line 81: Not sure if I am seeing this correctly, but it seems like by doing this, we could potentially overwrite the current storage state with an old value with the same key in the existing storage?

sieniven (dismissed):

lib/ain-evm/src/backend.rs
Jouzo: Yes, in the case where we would write multiple time to the same storage slot in a block (such as TD total_supply), we would overwrite the current storage with the updated value

sieniven: Isn't existing_storage the old value? Meaning that we are overwriting the updated value with the old value?

Jouzo: So it keeps the `storage` entry as is with `.entry(*k)` or if the key does not exist in storage, insert the existing key val from existing_storage. Feels correct but we can add some test if needed

Jouzo: Logic could probably be reversed since we can assume that `existing_storage` will grow and be on average bigger than `storage`

PR #2616, opened 2023-10-24
EVM: Fixes to backend state in storage and fix db dump debug RPC

sieniven:
## Summary

- Fix rollback pipeline on disconnect block to remove block logs and codehash index in EVM storage
- Add index for codehashes deployed in the block in storage for disconnect block pipeline
- Fix db dump RPC to return db state in RPC instead of logging on the node
- Enable db dump RPC check in feature_evm_rollback

## Implications

- Storage
  - [x] Database reindex optional
  - [ ] None

- Consensus
  - [x] None


lib/ain-evm/src/storage/block_store.rs
```diff
@@ -257,25 +257,28 @@ impl BlockStore {
...
+        let mut out = String::new();
         for arg in &[
             DumpArg::Blocks,
             DumpArg::Txs,
             DumpArg::Receipts,
             DumpArg::BlockMap,
             DumpArg::Logs,
         ] {
-            self.dump(arg, None, limit);
+            out.push_str(self.dump(arg, None, limit).as_str());
```
prasannavl, line 269: We're constructing a massive string here. This needs to be chunked. Better to use a buffered writer that just keeps writing out

sieniven: Switched to using fmt write to write to the string stream instead. To do in the future will be to shift out and use a streaming output on our RPC server.

lib/ain-evm/src/backend.rs
```diff
@@ -195,6 +195,10 @@ impl EVMBackend {
         self.overlay.state.clear()
     }
 
+    pub fn reset(&mut self) {
```
Jouzo, line 198:
```suggestion
    pub fn reset_to_last_changeset(&mut self) 
```
First though on reset is that we reset to initial state

Jouzo: Resolved on https://github.com/DeFiCh/ain/pull/2625

sieniven: Will make the switches on https://github.com/DeFiCh/ain/pull/2625

sieniven: Changes are in #2625

lib/ain-evm/src/backend.rs
```diff
@@ -249,8 +249,12 @@ impl EVMBackend {
             }
 
             if let Some(code) = code {
-                self.storage
-                    .put_code(self.vicinity.block_number, account.code_hash, code)?;
+                self.storage.put_code(
+                    self.vicinity.block_number,
```
Jouzo, line 253: Can use self.block_number()

sieniven: This isnt possible because of the mutable borrow that occurs in the drain method which prevents us from calling the method which creates another immutable borrow....

Jouzo (approved):

PR #2623, opened 2023-10-25
Evm state override

### LINKED ISSUE #2122: Add State Override Support for eth_call
## Summary

Extend `eth_call` RPC to include a `state_overrides` parameter for more flexible contract testing.

## Rationale

This allows end users to specify a temporary state against which testing their contract deployment or interaction. This override should be temporary and should not affect the actual state of the blockchain. This ensures that developers can run these simulations without unintentionally altering the state of the blockchain.
This helps developer with smart contract debugging and analysis.
It can potentially reduce the amount of code deployed on-chain. Code that simply returns internal state or does pre-defined validations can be kept off chain and fed to the node on-demand.

## Design

Additional state override struct :
```
struct StateOverride {
	pub balance: Option<U256>, 
	pub nonce: Option<U256>,
	pub code: Option<Bytes>, // code of the contract, if any
	pub state: Option<BTreeMap<H256, H256>>, // state of the contract
	pub state_diff: Option<BTreeMap<H256, H256>>, // the difference in state from the original
}
```
```
BTreeMap<H160, StateOverride>
```

## RPC

`eth_call` example :

```
eth_call { from: "0xd9c9cd5f6779558b6e0ed4e6acf6b1947e7fa1f3", to: <contract_address>, input: <data> } "latest"
{
  "0xd9c9cd5f6779558b6e0ed4e6acf6b1947e7fa1f3": {
    "balance": "0xde0b6b3a7640000"
  }
}
```
The balance of the `from` address is set to "0xde0b6b3a7640000" for the context of this call

--- 
Ref : https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth

Jouzo:
## Summary

This PR adds support for overriding the contract code, storage, account balance and nonce for simulations with `eth_call` and `eth_estimateGas`. This feature is similarly [supported by geth](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set).

The eth_call RPC method now accepts a third override parameter which is a map from an address to the desired state of that address. The state object has the following keys, all of which are optional:

    balance: QUANTITY - The balance to set for the account before executing the call.
    nonce: QUANTITY - The nonce to set for the account before executing the call.
    code: DATA - The EVM bytecode to set for the account before executing the call.
    state*: OBJECT - Key-value mapping of storage slot to value that will clear all slots and then override individual slots in the account storage before executing the call.
    stateDiff*: OBJECT - Key-value mapping of storage slot to value that will override individual slots in the account storage before executing the call.
    *Note - stateDiff is either applied to state or existing state

So, an example override object using all fields overriding multiple addresses could look like the following:
```
{
  "0xd9c9cd5f6779558b6e0ed4e6acf6b1947e7fa1f3": { 
    "nonce": "0xa", 
    "balance": "0xffff", 
    "state": { 
      "0000000000000000000000000000000000000000000000000000000000000001": "0xbaddad42baddad42baddad42baddad42baddad42baddad42baddad42baddad42"
    } 
  },
  "0xebe8efa441b9302a0d7eaecc277c09d20d684540": { 
    "code": "0x123456", 
    "stateDiff": {
      "0000000000000000000000000000000000000000000000000000000000000002": "0xbaddad42baddad42baddad42baddad42baddad42baddad42baddad42baddad42"
    } 
  }
}
```
Like all state changes made with `eth_call` and `eth_estimateGas`, the changes are ephemeral and only last for that specific run.

Fixes #2122


lib/ain-grpc/src/call_request.rs
```diff
@@ -127,3 +129,53 @@ impl CallRequest {
...
+    pub code: Option<Bytes>,
+    /// Fake key-value mapping to override all slots in the account storage before
+    /// executing the call.
+    pub state: Option<BTreeMap<H256, H256>>,
+    /// Fake key-value mapping to override individual slots in the account storage before
+    /// executing the call.
+    pub state_diff: Option<BTreeMap<H256, H256>>,
+}
+
+pub fn override_to_overlay(overide: BTreeMap<H160, CallStateOverride>) -> Overlay {
```
sieniven, line 152:
```suggestion
pub fn override_to_overlay(override: BTreeMap<H160, CallStateOverride>) -> Overlay {
```

lib/ain-grpc/src/call_request.rs
```diff
@@ -127,3 +129,53 @@ impl CallRequest {
...
+    let mut overlay = Overlay::default();
+
+    for (address, state_override) in overide {
```
sieniven, line 155:
```suggestion
    for (address, state_override) in override {
```

lib/ain-grpc/src/call_request.rs
sieniven, line 152: Will switching this into a from trait implementation be better?

sieniven (approved):

sieniven (approved):

PR #2631, opened 2023-10-27
EVM: Housekeeping and add parameter for eccache and EVM cache size limit

sieniven:
## Summary

- Implement cache size limit as a configurable parameter, set with -evmcachesize
- Implement the use of worker thread pool of parallel ec recovery cache of signed EVM txs as a configurable flag to turn on/off, set with -eccprecache
- Cleanup all eth references in source code to use evm instead.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-cpp-imports/src/bridge.rs
```diff
@@ -29,17 +29,17 @@ pub mod ffi {
 
         fn getChainId() -> u64;
         fn isMining() -> bool;
-        fn publishEthTransaction(data: Vec<u8>) -> String;
+        fn publishEvmTransaction(data: Vec<u8>) -> String;
```
prasannavl, line 32: These are ETH transactions, as they are TXs of the format of Ethereum. EVM has not nothing to do with this, nor defines a TX spec.

lib/ain-cpp-imports/src/bridge.rs
```diff
@@ -29,17 +29,17 @@ pub mod ffi {
...
         fn getAccounts() -> Vec<String>;
         fn getDatadir() -> String;
         fn getNetwork() -> String;
-        fn getEthMaxConnections() -> u32;
+        fn getEvmMaxConnections() -> u32;
```
prasannavl, line 36: Three is no EVM Connection. The connections are for ETH-JSON RPC. And as such I'd argue the original naming is technically more accurate.

lib/ain-cpp-imports/src/bridge.rs
```diff
@@ -29,17 +29,17 @@ pub mod ffi {
...
         fn getDifficulty(block_hash: [u8; 32]) -> u32;
         fn getChainWork(block_hash: [u8; 32]) -> [u8; 32];
         fn getPoolTransactions() -> Vec<TransactionData>;
         fn getNativeTxSize(data: Vec<u8>) -> u64;
         fn getMinRelayTxFee() -> u64;
-        fn getEthPrivKey(key: String) -> [u8; 32];
+        fn getEvmPrivKey(key: String) -> [u8; 32];
```
prasannavl, line 42: Again, EVM defines no concept of EVM address nor EVM keys. 

EVM doesn't really care for a format really. It's either ERC55, or Ethereum address format. We do follow the ethereum format. So, again, the original would be more accurate.

lib/ain-cpp-imports/src/bridge.rs
```diff
@@ -50,5 +50,6 @@ pub mod ffi {
         fn getNumCores() -> i32;
         fn getCORSAllowedOrigin() -> String;
         fn getNumConnections() -> i32;
+        fn getEvmCacheSizeLimit() -> usize;
```
prasannavl, line 53: Would just call it as ECC pre-cache limit to side step this naming issue completely.

src/init.cpp
```diff
@@ -424,6 +424,8 @@ void SetupServerArgs()
     gArgs.AddArg("-datadir=<dir>", "Specify data directory", ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-dbbatchsize", strprintf("Maximum database write batch size in bytes (default: %u)", nDefaultDbBatchSize), ArgsManager::ALLOW_ANY | ArgsManager::DEBUG_ONLY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-dbcache=<n>", strprintf("Maximum database cache size <n> MiB (%d to %d, default: %d). In addition, unused mempool memory is shared for this cache (see -maxmempool).", nMinDbCache, nMaxDbCache, nDefaultDbCache), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
+    gArgs.AddArg("-evmcachesize=<n>", strprintf("Maximum EVM database cache size <n> items (default: %d).", DEFAULT_EVM_CACHE_SIZE_LIMIT), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
```
prasannavl, line 427: Would suggest not to use EVM terminology at all. And EVM cache size is also misleading here and we're not caching anything else on EVM. 

Just `ecccachesize` or `ecclrusize` would be closer to what we do it's cache is primarily intended for ECC key recovery.

src/init.cpp
```diff
@@ -524,7 +526,7 @@ void SetupServerArgs()
     gArgs.AddArg("-dexstats", strprintf("Enable storing live dex data in DB (default: %u)", DEFAULT_DEXSTATS), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-blocktimeordering", strprintf("(Deprecated) Whether to order transactions by time, otherwise ordered by fee (default: %u)", false), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-txordering", strprintf("Whether to order transactions by entry time, fee or both randomly (0: mixed, 1: fee based, 2: entry time) (default: %u)", DEFAULT_TX_ORDERING), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
-    gArgs.AddArg("-ethstartstate", strprintf("Initialise Ethereum state trie using JSON input"), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
+    gArgs.AddArg("-evmstartstate", strprintf("Initialise EVM state trie using JSON input"), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
```
prasannavl, line 529: This is arguably the one part I'm more inclined to agree, though can be argued that eth will remain more consistent in line with other references. 

However, we have set an rc. So, for better or worse, the window to discuss is past, unless you intend to alias this and consider deprecation, which IMO isn't worth it.

lib/ain-cpp-imports/src/bridge.rs
sieniven: Naming ECC pre-cache may not be accurate since the parameter isnt just used for signed tx cache, as it is also used for the validation cache.

sieniven:
Discarding PR since the clean up doesnt help cleanup naming but creates more confusion instead.

PR #2632, opened 2023-10-28
EVM: Add caching and concurrency control flags

sieniven:
## Summary

- add config parameters for using parallel ecc recovery on validation pipeline and LRU cache size limit for validation and signedtx caches

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


src/init.cpp
```diff
@@ -424,6 +424,8 @@ void SetupServerArgs()
     gArgs.AddArg("-datadir=<dir>", "Specify data directory", ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-dbbatchsize", strprintf("Maximum database write batch size in bytes (default: %u)", nDefaultDbBatchSize), ArgsManager::ALLOW_ANY | ArgsManager::DEBUG_ONLY, OptionsCategory::OPTIONS);
     gArgs.AddArg("-dbcache=<n>", strprintf("Maximum database cache size <n> MiB (%d to %d, default: %d). In addition, unused mempool memory is shared for this cache (see -maxmempool).", nMinDbCache, nMaxDbCache, nDefaultDbCache), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
+    gArgs.AddArg("-lrucache=<n>", strprintf("Maximum LRU cache size <n> items (default: %d).", DEFAULT_LRU_CACHE_SIZE_LIMIT), ArgsManager::ALLOW_ANY, OptionsCategory::OPTIONS);
```
prasannavl, line 427: How about we use `ecclrucache`? LRU is rather too generic, and we have many LRUs all around, so it might the scoping would put it in a more accurate phase. The cache is indeed related to ECC. SignedTX cache also exists purely because of the ECC.

sieniven: Hmm but the config is also uses for VailidationTxCache. Wont it be abit confusing?

prasannavl: ValidationTxCache's purpose is the same as well. It's for ECC.

prasannavl: I see what you mean. But is the validation txcache even that significant? I'm not sure it that really even matters to carry the confg. We can keep that a default 4k or so. I don't think it matters as much?

PR #2633, opened 2023-10-29
Fix eth RPC call context and fix gas estimate pipeline

sieniven:
## Summary

- This PR fixes all eth RPC call request contexts to align with geth interface.
- Fix estimate_gas pipeline to execute the requested code and determining the highest and lowest possible gas limits to binary search between them. (ref: https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/bind/backends/simulated.go#L537-L639)
- estimate_gas pipeline now aligns with geth - if the tx fails state execution, estimate_gas RPC will throw an error on the specified call context.
- Fix processing data and input fields inside all eth calls - RPCs should not allow both data and input fields to be set, and data field or input field should be used to set the transaction data. (ref: https://github.com/fjl/go-ethereum/commit/c20147392245250de1efb2122979b69ae3a5bff5)
- Fix gas price calculation inside all eth calls - RPCs should not allow if gasPrice and (maxFeePerGas or maxPriorityFeePerGas) fields are both set. Correct effective gas price calculation is now handled inside the call context.
- Cleanup eth call in core services - EthCallArgs should contain only the caller's gas price to be passed into executor to be executed on the VM.
- Refactor and clean up all eth, debug, net RPCs for better and proper return of error messages

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-grpc/src/errors.rs
+++ b/lib/ain-grpc/src/errors.rs
@@ -1,44 +1,40 @@
 use ain_evm::EVMError;
 use jsonrpsee::core::Error;
 
+pub enum RPCError {
+    EvmCall(EVMError),
+    NoSenderAddress,
+    InsufficientFunds,
+    ValueOverflow,
+    InvalidGasPrice,
+    InvalidTransactionType,
+    InvalidDataInput,
+    TxExecutionFailed,
+    GasCapTooLow(u64),
+}
+
+impl From<RPCError> for Error {
+    fn from(e: RPCError) -> Self {
+        match e {
+            RPCError::EvmCall(e) => Error::Custom(format!("error calling EVM : {e:?}")),
+            RPCError::NoSenderAddress => to_custom_err("no from address specified"),
+            RPCError::InsufficientFunds => to_custom_err("insufficient funds for transfer"),
+            RPCError::ValueOverflow => to_custom_err("value overflow"),
+            RPCError::InvalidGasPrice => {
+                to_custom_err("both gasPrice and (maxFeePerGas or maxPriorityFeePerGas) specified")
+            }
+            RPCError::InvalidTransactionType => to_custom_err("invalid transaction type specified"),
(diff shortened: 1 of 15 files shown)


lib/ain-grpc/src/errors.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub fn invalid_data_err() -> Error {
+    Error::Custom(String::from("data and input fields are mutually exclusive"))
+}
+
+pub fn tx_execution_failure_err() -> Error {
+    Error::Custom(String::from("transaction execution failed"))
+}
+
+pub fn gas_cap_too_low_err(cap: u64) -> Error {
```
Jouzo, line 42: Could use an enum that implements Into<jsonrpsee::core::Error::Custom>.

sieniven: Resolved in PR [8a88018](https://github.com/DeFiCh/ain/pull/2633/commits/8a880188c3b0717cf8f402ea45114a385b31feab)

lib/ain-grpc/src/errors.rs
```diff
@@ -2,33 +2,57 @@ use ain_evm::EVMError;
...
-    GasCapTooLow(u64),
+    ValueOverflow,
 }
 
 impl From<RPCError> for Error {
     fn from(e: RPCError) -> Self {
         match e {
-            RPCError::EvmCall(e) => Error::Custom(format!("error calling EVM : {e:?}")),
+            RPCError::AccountError => to_custom_err("error getting account"),
+            RPCError::BlockNotFound => to_custom_err("block not found"),
```
Jouzo, line 28: Any reason to change error message here ? `header not found` keeps us in sync with `geth`

sieniven: No reason, i'll switch it back

PR #2634, opened 2023-10-30
evm: `debug_traceTransaction`

shohamc1:
Supersedes #2007


lib/ain-cpp-imports/src/bridge.rs
```diff
@@ -50,5 +50,7 @@ pub mod ffi {
         fn getNumCores() -> i32;
         fn getCORSAllowedOrigin() -> String;
         fn getNumConnections() -> i32;
+        fn isDebugEnabled() -> bool;
```
prasannavl, line 53: Let's retrospectively change this to isEthDebugRPCEnabled, isEthDebugTraceEnabled, etc.

lib/ain-evm/src/core.rs
```diff
@@ -740,6 +755,88 @@ impl EVMCoreService {
...
+        debug!(
+            "Calling EVM at block number : {:#x}, state_root : {:#x}",
+            block_number, state_root
+        );
+
+        let vicinity = Vicinity {
+            block_number,
+            origin: caller,
+            ..Default::default()
+        };
```
sieniven, line 785: Not setting the full vicinity state of the backend with the block context could lead to incorrect state execution

lib/ain-cpp-imports/src/bridge.rs
sieniven: Resolved on PR https://github.com/DeFiCh/ain/pull/2641

lib/ain-evm/src/core.rs
sieniven: Resolved on PR https://github.com/DeFiCh/ain/pull/2641

PR #2638, opened 2023-10-30
EVM: Prevent overriding overlay contract code

Jouzo:
## Summary

- Check if a code has been deployed on a previous changeset and keep it set if present.
- Prevent TD state injection from overriding deployed contract code in a same block


prasannavl:
TODO: Let's add a test here as well for Token creation and TransferDomain in the same block.

PR #2652, opened 2023-11-01
evm: Add max size test

shohamc1:


test/functional/feature_evm_contracts.py
```diff
@@ -438,6 +438,21 @@ def non_payable_proxied_contract(self):
...
+    # should fail since it is larger than DVM TX size limit
+    def fail_send_large_tx(self):
+        assert_raises_web3_error(
+            -32001,
+            "reason: scriptpubkey",
+            self.node.w3.eth.send_transaction,
+            {
+                "to": "0x0000000000000000000000000000000000000000",
+                "from": self.evm_key_pair.address,
+                "data": "0x" + "f" * (32000 * 5),
```
prasannavl, line 450: Why not use `ff` so it's 1 byte. And then go right above 64k? (DVM Envelope would take up a bit more, but should be minimal for EVM TX)

shohamc1: Makes sense, updated in 2c85aa1

prasannavl (approved):

prasannavl:
Blocked on https://github.com/DeFiCh/ain/pull/2653 for CI runs

test/functional/feature_evm_contracts.py
```diff
@@ -438,6 +438,21 @@ def non_payable_proxied_contract(self):
...
+    # should fail since it is larger than DVM TX size limit
+    def fail_send_large_tx(self):
+        assert_raises_web3_error(
+            -32001,
+            "reason: scriptpubkey",
+            self.node.w3.eth.send_transaction,
+            {
+                "to": "0x0000000000000000000000000000000000000000",
+                "from": self.evm_key_pair.address,
+                "data": "0x" + "ff" * 64000,
```
prasannavl, line 450: The failure is not due to size. This will fail regardless of the size. Will fail even if it's just set to ff.

prasannavl: Updated in https://github.com/DeFiCh/ain/pull/2652/commits/9dca6dc55da3269030d7a02e3fbba2aed07c862d

PR #2657, opened 2023-11-01
test: Add EVM minimum RBF test

shohamc1:


prasannavl:
Blocked on https://github.com/DeFiCh/ain/pull/2653 for CI runs

test/functional/feature_evm.py
```diff
@@ -1603,6 +1603,77 @@ def print_fee_info():
...
+            self.eth_address,
+            nonce,
+            180,
+            21001,
+            self.to_address,
+            1,
+        )
+
+        # Send tx with more than 100% in increase fees
+        self.nodes[0].evmtx(self.eth_address, nonce, 220, 21001, self.to_address, 1)
```
sieniven, line 1660: Can we do a stricter check here to send the exact min replacement tip fee?

sieniven: Since in the above 50% fee increment test, a tip fee of more than the specified is already tested

Bushstar (approved):

sieniven (approved):

prasannavl:
We should clean up all the `getAttributesValues`, do it once and pass the Attributes along later. 

LGTM now.

PR #2669, opened 2023-11-02
Dev: enable rust debug

canonbrother:
## Summary

- Set `debug=2` by default instead of `1`, dev friendly for debugger. Even though can remove it since debug=2 is default for dev, btw explicitly set is preferable
ref: https://doc.rust-lang.org/cargo/reference/profiles.html

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -13,7 +13,7 @@ rustflags = [
 ]
 
 [profile.dev]
-debug = 1
+debug = 2
 
 [target.x86_64-apple-darwin]
 rustflags = [


prasannavl:
This was specifically pulled down as the CIs run out of space without. And debug=1 retains sufficient enough symbols, however diff. in space is rather large making it inefficient unless needed.

defid binary out should also be significantly different: 

defid with debug=2 ~= 600-700mb
defid with debug=1 ~= 20mb. 

Was there a specific reason you needed debug=2? 1 should work with perf, most OS tools for backtraces as it'll retain the key symbols.

prasannavl:
#2527 was the PR that brought it down intentionally. Closing as at it's current stage, it makes all CI runner fail to be the absurd disk requirements of debug=1.

canonbrother:
> Was there a specific reason you needed debug=2? 1 should work with perf, most OS tools for backtraces as it'll retain the key symbols.

the variable values is only shown on `debug=2` on debugger
btw its alright due to large significant size built.. will set manually if needed on my end

PR #2675, opened 2023-11-04
Index EVM txs' block hash and index

Jouzo:
## Summary

- Index EVM TXs' block hash and index by TX hash
- Prevent indexing TransactionV2 twice


Jouzo:
Expected results: 10% disk usage reduction on changi at block height 1692012

sieniven (dismissed):

PR #2686, opened 2023-11-07
EVM: Fix create token

sieniven:
## Summary

- Add UTF-8 encoding check on cpp string to rust string conversions
- Limit token name length
- Add additional layer of UTF-8 encoding check on genesis block DST20 migration
- Add unit test for UTF-8 validity checks

## Implications

- Storage
  - [x] None

- Consensus
  - [x] Includes backward compatible changes
  - [ ] None

### DIFF
--- a/src/dfi/consensus/tokens.cpp
+++ b/src/dfi/consensus/tokens.cpp
@@ -90,16 +90,22 @@ Res CTokensConsensus::operator()(const CCreateTokenMessage &obj) const {
         return res;
     }
 
+    const auto &consensus = txCtx.GetConsensus();
+    const auto height = txCtx.GetHeight();
+    const auto &tx = txCtx.GetTransaction();
+    auto &mnview = blockCtx.GetView();
+
     CTokenImplementation token;
     static_cast<CToken &>(token) = obj;
 
     auto tokenSymbol = trim_ws(token.symbol).substr(0, CToken::MAX_TOKEN_SYMBOL_LENGTH);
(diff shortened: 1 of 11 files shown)


src/dfi/consensus/tokens.cpp
```diff
@@ -90,16 +90,22 @@ Res CTokensConsensus::operator()(const CCreateTokenMessage &obj) const {
...
 
     auto tokenSymbol = trim_ws(token.symbol).substr(0, CToken::MAX_TOKEN_SYMBOL_LENGTH);
     auto tokenName = trim_ws(token.name).substr(0, CToken::MAX_TOKEN_NAME_LENGTH);
-
-    const auto &consensus = txCtx.GetConsensus();
-    const auto height = txCtx.GetHeight();
-    const auto &tx = txCtx.GetTransaction();
-    auto &mnview = blockCtx.GetView();
+    if (height >= consensus.DF22MetachainHeight) {
+        tokenName = (tokenName).substr(0, CToken::MAX_DST20_TOKEN_NAME_LENGTH);
```
Jouzo, line 104: Wouldn't it be better to fail here rather than trim the `tokenName` and have unexpected results ?

prasannavl: Definitely agree, this would be way better and more predictive behaviour.

sieniven: Resolved - added token name byte size limit check in e32f472

Jouzo (approved):

PR #2692, opened 2023-11-13
EVM: Add Code to storage cache

Jouzo:
## Summary

- Add code to storage cache
- Switch storage cache to mutex only since RwLock is always write locked to update LRU


sieniven (dismissed):

lib/ain-evm/src/storage/cache.rs
```diff
@@ -3,30 +3,35 @@ use std::{borrow::ToOwned, num::NonZeroUsize, sync::RwLock};
...
 #[derive(Debug)]
 pub struct Cache {
-    transactions: RwLock<LruCache<H256, TransactionV2>>,
-    blocks: RwLock<LruCache<U256, BlockAny>>,
-    block_hashes: RwLock<LruCache<H256, U256>>,
+    transactions: Mutex<LruCache<H256, TransactionV2>>,
+    blocks: Mutex<LruCache<U256, BlockAny>>,
+    block_hashes: Mutex<LruCache<H256, U256>>,
     latest_block: RwLock<Option<BlockAny>>,
+    codes: Mutex<LruCache<H256, Vec<u8>>>,
```
shohamc1, line 17: I think this would be more descriptive.
```suggestion
    contract_code: Mutex<LruCache<H256, Vec<u8>>>,
```

shohamc1 (approved):

lib/ain-evm/src/storage/mod.rs
```diff
@@ -193,7 +193,17 @@ impl FlushableStorage for Storage {
 
 impl Storage {
     pub fn get_code_by_hash(&self, address: H160, hash: H256) -> Result<Option<Vec<u8>>> {
```
sieniven, line 195: Should we implement a CodeStorage trait since its used by both the cache and blockstore?

PR #2693, opened 2023-11-14
Fix EVM block invalidation

Jouzo:
## Summary

- Check if EVM block hash exists in `vmmap` index before invalidating

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -3809,7 +3809,11 @@ bool CChainState::DisconnectTip(CValidationState &state,
             mnview.GetHistoryWriters().DiscardDB();
             return error("DisconnectTip(): DisconnectBlock %s failed", pindexDelete->GetBlockHash().ToString());
         }
-        XResultThrowOnErr(evm_try_disconnect_latest_block(result));
+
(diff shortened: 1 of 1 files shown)


src/validation.cpp
```diff
@@ -3809,7 +3809,12 @@ bool CChainState::DisconnectTip(CValidationState &state,
             mnview.GetHistoryWriters().DiscardDB();
             return error("DisconnectTip(): DisconnectBlock %s failed", pindexDelete->GetBlockHash().ToString());
         }
-        XResultThrowOnErr(evm_try_disconnect_latest_block(result));
+
+        if (auto res =
+                pcustomcsview->GetVMDomainBlockEdge(VMDomainEdge::DVMToEVM, pindexDelete->GetBlockHash().ToString())) {
```
sieniven, line 3814: Would it be better to use the tryfrom XVM static method on the block coinbase tx to try to get the XVM struct information instead to determine if the DVM block has an EVM block?

Jouzo: Sounds better, let me update

sieniven: LGTM 👍

sieniven (approved):

PR #2697, opened 2023-11-17
evm: Fix divide by zero error in `eth_estimateGas`

shohamc1:


lib/ain-grpc/src/call_request.rs
```diff
@@ -39,6 +39,25 @@ pub struct CallRequest {
 }
 
 impl CallRequest {
+    pub fn guess_tx_type(&mut self) -> Result<(), Error> {
```
Jouzo, line 42: What's the context here ? Why do we want to re-assign to transaction_type ?

sieniven: @shohamc1 I guess the tradeoff in abstracting this out of th effective_gas_price calculation is that we cannot keep CallRequest immutable. Not sure if that is the best way although i think the idea is nice, to set a tx type. We could abstract this into our custom deserialize implementation actually which may keep it cleaner. Alternatively not sure if abstracting this here is cleaner - maybe keeping it inside effective_gas_price calculation, since the tx type in the call context is really only relevant for gas price calculations, may be safer.

sieniven (dismissed):

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -846,7 +849,7 @@ impl MetachainRPCServer for MetachainRPCModule {
 
         while lo + 1 < hi {
             let sum = hi.checked_add(lo).ok_or(RPCError::ValueOverflow)?;
-            let mid = sum.checked_div(2u64).ok_or(RPCError::ValueOverflow)?;
+            let mid = sum.checked_div(2u64).ok_or(RPCError::DivideByZero)?;
```
canonbrother, line 852: hmm.. its clearly state `divide by 2`.. should name the error not `DivideByZero`??
suggestion:
```suggestion
            let mid = sum.checked_div(2u64).ok_or(RPCError::ValueUnderflow)?;
```

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -799,12 +802,12 @@ impl MetachainRPCServer for MetachainRPCModule {
...
                     return Err(RPCError::InsufficientFunds.into());
                 }
-                available = balance.checked_sub(value).ok_or(RPCError::ValueOverflow)?;
+                available = balance.checked_sub(value).ok_or(RPCError::ValueUnderflow)?;
             }
 
             let allowance = available
                 .checked_div(fee_cap)
-                .ok_or(RPCError::ValueOverflow)?;
+                .ok_or(RPCError::DivideByZero)?;
```
canonbrother, line 810: shouldn't validate the `fee_cap` is not zero before get divided for error `DividedByZero`??

shohamc1: That is what `checked_div` does.

lib/ain-grpc/src/rpc/eth.rs
shohamc1: `checked_div` only throws when the divisor is 0.

canonbrother: yea but the divisor has been hardcoded `2u64` right??

lib/ain-grpc/src/rpc/eth.rs
canonbrother: hmm.. i mean that should have 2 validations:
i. fee_cap is not allowed to zero
ii. check_div for underoverflow checking

lib/ain-grpc/src/call_request.rs
shohamc1: Made `guess_tx_type` not mutate `CallRequest` in 506043e as a temporary fix to unblock RPCs. We can merge this and then explore what the best way to handle `transaction_type` will be.

lib/ain-grpc/src/rpc/eth.rs
prasannavl: @canonbrother's point is correct. The error type shouldn't be dividebyzero, but rather something more generic like, division error.

lib/ain-grpc/src/rpc/eth.rs
prasannavl: Agree. And in relation to the other point, should be more generic. 

fee_cap check, ideal to have it checked, but probably not necessary if it isn't used directly later, and can just rely on checked_div

prasannavl:
Going ahead and merging this to unblock a patch release today, but let's fix the error type in another PR to cleanup. DivideByZero is not the most accurate error to be returned.

lib/ain-grpc/src/call_request.rs
```diff
@@ -39,6 +39,27 @@ pub struct CallRequest {
 }
 
 impl CallRequest {
+    pub fn guess_tx_type(&self) -> Result<Self, Error> {
```
prasannavl, line 42: Let's rename this. The response does co-relate well with the guess name. 

Perhaps `clone_with_guessed_tx_type`

prasannavl: Or better yet - even better is `guess_tx_type` simply just returns the guess. 

And then do a clone internally in the method.

lib/ain-grpc/src/rpc/eth.rs
shohamc1: Addressed in 9d2cd5cf2ab4a0975829a87b0d79b5f7f305bff0

lib/ain-grpc/src/call_request.rs
shohamc1: Renamed to `clone_with_guessed_tx_type` for now. We can decide on how to properly handle the mutation after we have unblocked the RPC users.

lib/ain-grpc/src/call_request.rs
```diff
@@ -39,6 +39,27 @@ pub struct CallRequest {
 }
 
 impl CallRequest {
+    pub fn clone_with_guessed_tx_type(&self) -> Result<Self, Error> {
```
Jouzo, line 42: This is only used by `get_effective_gas_price` where we can switch on the tx type itself. We shouldn't need to clone nor re-assign

Jouzo: Sure but this can be directly called by `get_effective_gas_price` and only return the tx_type itself

Bushstar (approved):

PR #2709, opened 2023-11-21
RPC: Add single key check for source and destination addresses in transferdomain RPC

sieniven:
## Summary

- Add optional parameter for single key check (singlekeycheck) that run on default in transferdomain RPC
- Single key check pipeline will check that the source and destination addresses specified in the transferdomain RPC matches the same pubkey
- Guards against unplanned loss in funds to ensure destination address specified in the call is to the correct corresponding address
- To do a transferdomain to a different destination address that does not correspond to the same key, the RPC will need to explicitly set the singlekeycheck parameter to false
- Fix all functional tests with the new parameter, and add single key check tests in feature_evm_transferdomain
- Add legacy addresses into addressmap RPC

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/ci/ethlibs_test/main.sh
+++ b/ci/ethlibs_test/main.sh
@@ -79,7 +79,7 @@ setup_fixtures() {
                     "v0/params/feature/transferdomain": "true"}}'
     $DEFI_CLI_BIN -regtest generatetoaddress 2 "$OWNERAUTHADDR"
 
-    $DEFI_CLI_BIN -regtest transferdomain '[{"src":{"address":"'"$OWNERAUTHADDR"'", "amount":"200@DFI", "domain":2}, "dst":{"address":"'"$ALICE"'", "amount":"200@DFI", "domain":3}}]'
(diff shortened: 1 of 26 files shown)


Bushstar (dismissed):

ci/ethlibs_test/main.sh
```diff
@@ -79,7 +79,7 @@ setup_fixtures() {
                     "v0/params/feature/transferdomain": "true"}}'
     $DEFI_CLI_BIN -regtest generatetoaddress 2 "$OWNERAUTHADDR"
 
-    $DEFI_CLI_BIN -regtest transferdomain '[{"src":{"address":"'"$OWNERAUTHADDR"'", "amount":"200@DFI", "domain":2}, "dst":{"address":"'"$ALICE"'", "amount":"200@DFI", "domain":3}}]'
+    $DEFI_CLI_BIN -regtest transferdomain '[{"src":{"address":"'"$OWNERAUTHADDR"'", "amount":"200@DFI", "domain":2}, "dst":{"address":"'"$ALICE"'", "amount":"200@DFI", "domain":3}, "singlekeycheck": false}]'
```
canonbrother, line 82: the second param should be nonce

Bushstar (approved):

PR #2717, opened 2023-11-23
CI: Shift workflows to local build, cleanup pipelines

sieniven:
## Summary

- 

## RPCs

- 

## Flags

- 

## Protocol

- 

## Storage

- 

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None


Bushstar (approved):

contrib/dockerfiles/defi.dockerfile
```diff
@@ -0,0 +1,24 @@
...
+FROM --platform=linux/amd64 ubuntu:latest as defi
+ARG TARGET
+ARG PACKAGE
+ENV PATH=/app/bin:$PATH
+LABEL org.defichain.name="defichain"
+LABEL org.defichain.arch=${TARGET}
+
+WORKDIR /app
+COPY ${PACKAGE} ./
+RUN tar -xvzf ${PACKAGE} --strip-components 1
```
prasannavl, line 12: This puts both on the image, the tar as well extracted versions, perhaps add another build state to have it clean?

.github/workflows/build-dev.yaml
```diff
@@ -25,74 +25,71 @@ env:
...
+    strategy:
+      matrix:
+        target: [x86_64-pc-linux-gnu, x86_64-w64-mingw32, x86_64-apple-darwin]
     env:
-      TARGET: x86_64-pc-linux-gnu
+      TARGET: ${{matrix.target}}
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
```
prasannavl, line 38: Think we can use the options for checkout action to `set-safe-directory: *` instead of the manual command.

contrib/genbuild.sh
```diff
@@ -25,7 +25,7 @@ git_check_in_repo() {
 DESC=""
 SUFFIX=""
 CURRENT_BRANCH=""
-if [ -z "$BUILD_VERSION" ] && [ "${BITCOIN_GENBUILD_NO_GIT}" != "1" ] && [ -e "$(command -v git)" ] && [ "$(git rev-parse --is-inside-work-tree 2>/dev/null)" = "true" ] && git_check_in_repo contrib/genbuild.sh; then
+if [ "${BITCOIN_GENBUILD_NO_GIT}" != "1" ] && [ -e "$(command -v git)" ] && [ "$(git rev-parse --is-inside-work-tree 2>/dev/null)" = "true" ] && git_check_in_repo contrib/genbuild.sh; then
```
prasannavl, line 28: Is there a specific reason to remove this?

sieniven: Skipping the dirty tag check when BUILD_VERSION is specified is incorrect in our previous docker build workflow because BUILD_VERSION was not passed on into the docker image itself which resulted in the defid version still being tagged as dirty. This was removed and reverted because with CI on local builds inside the runner, build should be clean and there is no need to explicitly skip the check.

In addition, BUILD_VERSION specified in CI will vary depending on if it is a tagged commit. Thus on build-release workflow, it should be passed directly as BUILD_DESC, while on merge to master, BUILD_VERSION should be defined as BUILD_SUFFIX. It would be easier to simply revert back to the old genbuild pipeline, which will resolve the incorrect dirty tag once we shift our build locally.

.github/workflows/build-dev.yaml
sieniven: This will not work. Seems like set-safe-directory is a bool config which defaults to true. Ref: https://github.com/actions/checkout/pull/770/.

sieniven: Seems like it is a known issue that when running jobs inside docker containers, which makes the config not work as desired. The known workaround is to explicitly call git config after the action checkout.

contrib/dockerfiles/defi.dockerfile
sieniven: Fixed in PR #2721

PR #2724, opened 2023-11-27
Ocean archive indexer and api server

Jouzo:
## Summary

- Introduces ocean indexes and API

## Flags :

```
"-oceanarchive", Enable ocean archive indexer
"-oceanarchiveserver", Enable ocean archive server
"-oceanarchiveport=<port>", Listen for ocean archive connections on <port>
"-oceanarchivebind=<addr>[:port]", Bind to given address to listen for Ocean connections. Do not expose the Ocean server to untrusted networks such as the public internet! This option is ignored unless -rpcallowip is also passed. Port is optional and overrides -oceanarchiveport. This option can be specified multiple times (default: 127.0.0.1 i.e., localhost)"
```


Jouzo:
List of RPC queries done by Ocean that should be exposed directly by FFI: 
```
account.getAccount
account.getAccountHistory
account.getBurnInfo
account.listAccountHistory
blockchain.getBlockchainInfo
loan.getLoanInfo
loan.getLoanScheme
masternode.getGov
net.getNetworkInfo
poolpair.getPoolPair
poolpair.listPoolPairs
token.getToken
token.listTokens
```

lib/ain-evm/src/storage/block_store.rs
```diff
@@ -95,52 +96,50 @@ impl TransactionStorage for BlockStore {
...
         }
     }
 
     fn put_latest_block(&self, block: Option<&BlockAny>) -> Result<()> {
         if let Some(block) = block {
             let latest_block_cf = self.column::<columns::LatestBlockNumber>();
             let block_number = block.header.number;
-            latest_block_cf.put(&"", &block_number)?;
+            // latest_block_cf.put(&"", &block_number)?;
+            latest_block_cf.put(&String::new(), &block_number)?;
```
canonbrother, line 142: Noted for reverting changes from #2783 cc @Jouzo

prasannavl (dismissed):

lib/ain-grpc/src/lib.rs
```diff
@@ -135,7 +135,7 @@ pub async fn init_ocean_server(addr: String) -> Result<()> {
     let (user, pass) = ain_cpp_imports::get_rpc_auth().map_err(|e| format_err!("{e}"))?;
     let client = Arc::new(
         Client::new(
-            &format!("localhost:{}", ain_cpp_imports::get_rpc_port()),
+            &format!("http://127.0.0.1:{}", ain_cpp_imports::get_rpc_port()),
```
prasannavl, line 138: Note, we might need to let this be configured from the flags similar to the rest, as it's not possible to expose externally otherwise.

prasannavl:
Going ahead and merging this in, as even though it doesn't affect consensus and the rest is also completed isolated and flagged - it's good to have the core part of it in for the testnet release before further refinements and fixes that target the optional ocean archive.

prasannavl:
TODO: Hoist the IP binding all the way up to config.

src/dfi/validation.cpp
```diff
@@ -4666,6 +4685,15 @@ Res ProcessDeFiEventFallible(const CBlock &block,
     // Construct undo
     FlushCacheCreateUndo(pindex, mnview, cache, uint256S(std::string(64, '1')));
 
+    // Ocean archive
```
Bushstar, line 4688: Maybe this should be after evm_try_unsafe_commit_block in ConnectBlock, that's the very last point of failure when connecting new blocks.

PR #2725, opened 2023-11-28
Fix calls to CreateToken

Bushstar:
## Summary

- CreateToken was updated to take a height argument but not all calls were updated at the same time. This PR updates all remaining calls which passed a bool set to false before to max int which results in the same logic as before 4.0.4.
- On mainnet creating a loan token or pool pair will fork 4.0.4 from other nodes.
- On mainnet token splits would also fail but this is a currently disabled feature.

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


sieniven (approved):

sieniven:
Closing - fixes are #2726.

PR #2730, opened 2023-11-29
Remove consortium

Bushstar:
## Summary

- Remove the unused consortium which has never been enabled on any network and is disabled after the next fork.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


src/dfi/consensus/tokens.cpp
```diff
@@ -64,23 +64,20 @@ ResVal<CScript> CTokensConsensus::MintableToken(DCT_ID id,
...
-        // Is a DAT, check founders auth
-        if (height < static_cast<uint32_t>(consensus.DF20GrandCentralHeight) && !HasFoundationAuth()) {
-            return Res::Err("token is DAT and tx not from foundation member");
-        }
-    } else {
+    if (!token.IsDAT()) {
         return Res::Err("tx must have at least one input from token owner");
     }
 
+    // It is a DAT, check founders auth
```
prasannavl, line 76: Is the reason to drop the https://github.com/DeFiCh/ain/pull/2730/files#diff-48843c84274310df9b298a031159a1961a0b0cdb4e85b2601def1305bee1febaL77 check intentional?

Bushstar: Yes. GrandCentral introduced consortium, so that if a DAT token without foundation auth did not fail at this point as we would then check for a consortium member. Now we are removing consortium we restore the previous behaviour, simply that, a none-DAT token must have an input from the token owner and a DAT token must have HasFoundationAuth.

Refer to the code linked below.

https://github.com/DeFiCh/ain/blob/17c7d0085c13a21e78e368b1c808299c4dfabf85/src/masternodes/mn_checks.cpp#L773-L779

sieniven (approved):

PR #2732, opened 2023-11-30
Remove usage of Discard() from codebase

Bushstar:
## Summary

- Prevent usage of Discard on flushable storage. A cache copy of CCustomCSView or a snapshot will have a copy of the flushable storage. This should not be discarded as this flushable storage may have data not yet written to disk.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -612,10 +612,16 @@ CTxMemPool::~CTxMemPool() {}
 
 CCustomCSView &CTxMemPool::accountsView() {
     if (!acview) {
+        LOCK(cs_main);
         assert(pcustomcsview);
         acview = std::make_unique<CCustomCSView>(*pcustomcsview);
     }
     return *acview;
+}
+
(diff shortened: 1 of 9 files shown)


prasannavl (approved):

prasannavl:
Let's do a full-sync to verify past consistency before merging.

src/txmempool.cpp
```diff
@@ -612,12 +612,18 @@ CTxMemPool::~CTxMemPool() {}
 
 CCustomCSView &CTxMemPool::accountsView() {
     if (!acview) {
+        LOCK(cs_main);
```
sieniven, line 615: Could this potentially deadlock the node?

Bushstar: cs_main is already held there. I'll remove the lock for this function and resetAccountsView() as they are redundant.

sieniven (approved):

PR #2733, opened 2023-11-30
EVM: Refactor FFI from rust strings to passing byte arrays

sieniven:
## Summary

This PR refactors all exposed rust FFIs to pass byte vectors instead of rust strings, specifically when passing BTC internal hash storage representation into our Rust interface.
- Fixes unsafe cpp std::string conversions to rust string, which will panic the node if it is not UTF-8 valid.
- Cleans up and passes all byte vectors as BE representation with the exposed FFI.
- Methods abstract conversions between BTC internal hash representations and BE byte arrays.
- XVM struct is maintained to store miner address and EVM block hash as strings to maintain backwards compatibility. The hashes are only generated internally from BTC hash representations, and are guaranteed to be UTF-8 valid.
- VM mapping stored on disk is also maintained to store hashes as string representation so that a re-index is not required. String conversions are safely generated by BTC hash representation.

## Context
1. Rust hashes are a fixed-sized uninterpreted hash types. By default we will pass byte vector in BE representation and stores it as big endian.
2. Rust hashes as_fixed_bytes method returns the internal byte array containing the fixed hash.
3. BTC hash representation internally stores bytes in little endian.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-evm/src/core.rs
```diff
@@ -159,9 +159,9 @@ pub struct ValidateTxInfo {
 }
 
 pub struct TransferDomainTxInfo {
-    pub from: String,
-    pub to: String,
-    pub native_address: String,
+    pub from: [u8; 20],
```
Jouzo, line 162: Can we type this ? 
Something like `pub type XAddress = [u8; 20];`

sieniven: Sounds good. Will add this in

sieniven: Switched in c6ff318

Jouzo (dismissed):

Jouzo (dismissed):

Jouzo (approved):

PR #2736, opened 2023-12-04
EVM: Revamp and fix filters system

sieniven:
## Summary

This PR cleans up the node's filter system, aligning ourselves as much as possible with geth.

## Filter System
Ref: https://github.com/ethereum/go-ethereum/blob/master/eth/filters/filter_system.go
1. Switch to LRU cache with cache size limit (defaults to 5000) to safeguard against attack vectors.
2. Revamp design of how we accumulate our filters - this PR shifts filter update out of validation pipeline, and is only accumulated on every call request.
4. Add block range limit and response log limit, defaulting to 2000 and 10_000 respectively. Aligns with Alchemy, BNB, etc for maximum reliability on calls.
5. Exit early if topics filter has more than 4 topics - Ref: https://github.com/ethereum/go-ethereum/pull/28494
6. Use FilterCriteria context to be passed directly when querying db for node information.
7. Discontinue updating filter cache on commit block pipeline and block template creation pipeline.
8. Cleanup and add more test coverage on filters RPC

## RPCs
1. Fixes to eth_getLogs and eth_newFilter RPCs handling of topics optional argument - each event should be able to digest a list of possible topics, matching a prefix of that list.
2. Shift handling of logic of RPC into filter service from RPC.
3. eth_newFilter - should allow fromBlock and toBlock specified to be future blocks.

## To do:
- Implementation of pending transactions filter change RPC.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -984,160 +983,139 @@ impl MetachainRPCServer for MetachainRPCModule {
...
-                        .checked_add(U256::one())
-                        .ok_or(RPCError::ValueOverflow)?;
-                }
-                Ok::<Vec<U256>, Error>(block_numbers)
+        let from_block = if input.from_block.is_some() {
+            if let Some(BlockNumber::Num(block_num)) = input.from_block {
+                // Allow future block number to be specified
+                Some(U256::from(block_num))
+            } else {
+                Some(self.get_block(input.from_block)?.header.number)
```
Jouzo, line 991: Can simplify with map here

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -984,160 +983,139 @@ impl MetachainRPCServer for MetachainRPCModule {
...
+        };
+        let topics = if let Some(topics) = input.topics {
+            match topics {
+                LogRequestTopics::VecOfHashes(inputs) => {
+                    Some(inputs.into_iter().map(|input| vec![input]).collect())
+                }
+                LogRequestTopics::VecOfHashVecs(inputs) => Some(inputs),
+            }
+        } else {
+            None
```
Jouzo, line 1014: Same as above

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -984,160 +983,139 @@ impl MetachainRPCServer for MetachainRPCModule {
...
+        };
+        let topics = if let Some(topics) = input.topics {
+            match topics {
+                LogRequestTopics::VecOfHashes(inputs) => {
+                    Some(inputs.into_iter().map(|input| vec![input]).collect())
+                }
+                LogRequestTopics::VecOfHashVecs(inputs) => Some(inputs),
+            }
+        } else {
```
Jouzo, line 1063: Same as above

Jouzo (dismissed):

lib/ain-grpc/src/rpc/eth.rs
sieniven: Refactored topics pipeline to use map in [e7ae914](https://github.com/DeFiCh/ain/commit/e7ae914e0564533816ef18415a3266cd5fe2b413)

lib/ain-grpc/src/rpc/eth.rs
sieniven: Refactored topics pipeline to use map in [e7ae914](https://github.com/DeFiCh/ain/commit/e7ae914e0564533816ef18415a3266cd5fe2b413)

Jouzo (approved):

PR #2748, opened 2023-12-14
WIP: Ocean API database implementation to all modules

nagarajm22:
## Summary

- Ocean API `CRUD` implementation to all module.

## Storage

- `rocks db` crud operation.

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None


Jouzo:
Can we close https://github.com/DeFiCh/ain/pull/2746 for this PR ?

lib/ain-ocean/src/data_acces/masternode.rs
```diff
@@ -1,19 +1,49 @@
...
+            }
+        }
+
+        // Sort blocks based on the specified sort order
+        match sort_order {
+            SortOrder::Ascending => master_node.sort_by(|a, b| a.block.height.cmp(&b.block.height)),
+            SortOrder::Descending => {
+                master_node.sort_by(|a, b| b.block.height.cmp(&a.block.height))
+            }
+        }
```
Jouzo, line 44: We would need a From impl from `SortOrder` to RocksDB Iterator mode. 
In both SortOrder cases, we iterate from the end, get the batch and sort it. We should be iterating from start for `SortOrder::Ascending`.

lib/ain-ocean/src/data_acces/masternode_states.rs
```diff
@@ -13,25 +14,71 @@ pub struct MasterStatsDb {
...
+        match self.db.get("block", &latest_height.to_be_bytes()) {
+            Ok(Some(value)) => {
+                let block: MasternodeStats =
+                    serde_json::from_slice(&value).map_err(|e| anyhow!(e))?;
+                Ok(Some(block))
+            }
+            Ok(None) => Ok(None), // No block found for the latest height
+            Err(e) => Err(anyhow!(e)),
+        }
     }
```
Jouzo, line 45: Couldn't we iterator from End with limit = 1 and take the latest entry here ? This would bypass the need to query the block db as well

lib/ain-ocean/src/data_acces/masternode.rs
```diff
@@ -1,19 +1,49 @@
...
 impl MasterNodeDB {
-    pub async fn query(&self, limit: i32, lt: i32) -> Result<Vec<Masternode>> {
-        todo!()
+    pub async fn query(
+        &self,
+        limit: i32,
+        lt: i32,
+        sort_order: SortOrder,
+    ) -> Result<Vec<Masternode>> {
+        let iterator = self.db.iterator("masternode", IteratorMode::End)?;
```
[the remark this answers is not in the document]
Jouzo: We need to dig through Ocean spec and check all the models that need a secondary sort index (I guess all the one that are currently indexed by TX id but needs to be iterable via block height). I'll add this requirement to the main tracking PR

lib/ain-ocean/src/data_acces/masternode.rs
```diff
@@ -1,19 +1,49 @@
...
+                Ok((key, value)) => (key, value),
+                Err(err) => return Err(anyhow!("Error during iteration: {}", err)),
+            };
+
+            let master_stats: Masternode = serde_json::from_slice(&value)?;
+            master_node.push(master_stats);
+
+            if master_node.len() == limit as usize {
+                break;
+            }
```
Jouzo, line 35: Can change this recurring pattern to use `.take(limit as usize)` directly on the iterator

lib/ain-ocean/src/data_acces/masternode_states.rs
```diff
@@ -13,25 +14,71 @@ pub struct MasterStatsDb {
...
+            }
+        }
+
+        // Sort blocks based on the specified sort order
+        match sort_order {
+            SortOrder::Ascending => master_node.sort_by(|a, b| a.block.height.cmp(&b.block.height)),
+            SortOrder::Descending => {
+                master_node.sort_by(|a, b| b.block.height.cmp(&a.block.height))
+            }
+        }
```
Jouzo, line 79: Once we convert the `lt` and `sort_order` to rocksdb iterator mode, most of the list logic can be handled with something such as :
```
self.db
            .iterator(<cf_name>, iter_mode)?
            .into_iter()
            .take(limit as usize)
            .map(|result| {
                result
                    .map_err(|e| anyhow!("Error during iteration: {}", e))
                    .and_then(|(_key, value)| serde_json::from_slice(&value).map_err(Into::into))
            })
            .collect()
```

This cleans it up and removes the need for a mut vector and sorting post collecting

Jouzo: Can apply to other query impl as well

lib/ain-ocean/src/database/db_manager.rs
```diff
@@ -281,3 +287,12 @@ impl ColumnFamilyOperations for RocksDB {
         }
     }
 }
+
+impl<'a> From<SortOrder> for IteratorMode<'a> {
+    fn from(sort_order: SortOrder) -> Self {
+        match sort_order {
+            SortOrder::Ascending => IteratorMode::Start,
+            SortOrder::Descending => IteratorMode::From(b"", Direction::Reverse),
```
Jouzo, line 295:
```suggestion
            SortOrder::Descending => IteratorMode::End,
```

lib/ain-ocean/src/data_acces/masternode.rs
```diff
@@ -1,19 +1,47 @@
...
+            .take(limit as usize)
+            .map(|result| {
+                result
+                    .map_err(|e| {
+                        anyhow!("Error during iteration: {}", e)
+                            .context("error master_node query error")
+                    })
+                    .and_then(|(_key, value)| {
+                        let stats: Masternode = serde_json::from_slice(&value)?;
+                        if stats.block.height < lt {
```
Jouzo, line 34: I'm not sure this `lt` will do what we expect on both SortOrder.
I think we'd better add a `From<(SortOrder, StartKey)>` to `Itermode::From(<StartKey.as_slice()>, <SortOrder into Direction>)`, drop the `lt` altogether and replace it with the start index.

lib/ain-ocean/src/data_acces/masternode.rs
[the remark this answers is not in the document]
Jouzo: Let's take care of the rest and implement these secondary indexes in another dedicated PR.

lib/ain-ocean/src/data_acces/masternode.rs
nagarajm22: I am facing issue whenever I use `lt` for Ascending order. Its doesn`t loop whole collection, its just throw the error. I ll update with `startKey`. Rocks db doesn't store data in order.

Jouzo: Keys are sorted. That's why we need a secondary sort index by block height + txno to have an iteration that makes sense (Retrieve masternodes by order of creation).

PR #2753, opened 2023-12-18
Fix release workflow

canonbrother:
## Summary

- as per title


Bushstar (approved):

.github/workflows/build-dev.yaml
```diff
@@ -25,12 +25,33 @@ env:
   GIT_VERSION: 1
 
 jobs:
+  win-x64:
```
sieniven, line 28: This doesnt fix anything since the changes are on build-dev workflow

PR #2754, opened 2023-12-19
Debug building workflow

Bushstar:
## Summary

- Workaround to get the Windows release working

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   create-release:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/build-release.yaml
```diff
@@ -13,7 +13,7 @@ env:
 
 jobs:
   create-release:
-    runs-on: ubuntu-latest
```
sieniven, line 16: Maybe we should not get this in since it doesnt resolve the win-x64 bug, and switching to this shouldnt matter anyway since we run our builds entirely within the builder container.

PR #2770, opened 2024-01-08
evm: `eth_createAccessList`

shohamc1:


lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
 
         Ok((listener.trace, execution_success, return_value, used_gas))
     }
+
+    #[allow(clippy::too_many_arguments)]
```
Jouzo, line 820: Unnecessary

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
...
+        let mut backend = EVMBackend::from_root(
+            state_root,
+            Arc::clone(&self.trie_store),
+            Arc::clone(&self.storage),
+            vicinity.clone(),
+            None,
+        )
+        .map_err(|e| format_err!("Could not restore backend {}", e))?;
+        backend.vicinity.origin = caller;
+        backend.vicinity.gas_price = arguments.gas_price;
```
Jouzo, line 854:
```suggestion
        let mut vicinity = Vicinity::from(block_header);
        vicinity.origin = caller;
        vicinity.gas_price = arguments.gas_price;
        let backend = EVMBackend::from_root(
            state_root,
            Arc::clone(&self.trie_store),
            Arc::clone(&self.storage),
            vicinity,
            None,
        )?;
```

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
...
+
+        static CONFIG: Config = Config::shanghai();
```
Jouzo, line 856: We should look into sharing this.
We could make AinExecutor CONFIG pub and  use `static CONFIG: Config = AinExecutor::CONFIG;` or use a single global const

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
...
+            .into_iter()
+            .map(|(address, storage_keys)| AccessListItem {
+                address,
+                storage_keys: Vec::from_iter(storage_keys),
+            })
+            .collect();
+
+        al_executor.transact_call(
+            caller,
+            to.unwrap(),
```
Jouzo, line 893: Is it safe to unwrap here ?

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
 
         Ok((listener.trace, execution_success, return_value, used_gas))
     }
+
+    pub fn create_access_list(&self, arguments: EthCallArgs) -> Result<(AccessList, u64)> {
```
sieniven, line 820: trace_transaction and create_access_list should not be placed under block template method implementations. We should not be placing the implementation of these methods anyway inside the core module, and perhaps shift the implementation abstracted inside the executor, while the core service wraps the call to the executor.

lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -1133,6 +1141,41 @@ impl MetachainRPCServer for MetachainRPCModule {
...
+        let block = self.get_block(block_number)?;
+        let block_base_fee = block.header.base_fee;
+        let gas_price = call.get_effective_gas_price(block_base_fee)?;
+
+        Ok(self
+            .handler
+            .core
+            .create_access_list(EthCallArgs {
+                caller,
+                to: call.to,
```
sieniven, line 1168: The to field in eth_createAccessList RPC is a required field when passing the call context for this RPC. We should check that the field is populated and if it isnt, return the appropriate RPC error

lib/ain-evm/src/core.rs
sieniven: Since there isnt a check on the to field in the call context, it will cause the node to panic on this unwrap

test/functional/feature_evm_createaccesslist.py
```diff
@@ -0,0 +1,163 @@
...
+        }
+        access_list = self.node.eth_createAccessList(transaction)
+        al = access_list["accessList"][0]
+
+        assert_equal(len(al["storageKeys"]), 3)
+        assert_equal(al["address"], "0xff00000000000000000000000000000000000001")
+
+
+if __name__ == "__main__":
+    AccessListTest().main()
```
sieniven, line 163: Can we write test pipelines similar to other functional tests where we abstract test setup, and each functional test into methods? Each method should also do a rollback on the chain state so that we could add more test coverage quickly without the need for each test to be dependent on each other

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
...
+        let metadata = StackSubstateMetadata::new(gas_limit, &CONFIG);
+        let state = MemoryStackState::new(metadata.clone(), &backend);
+        let al_state = MemoryStackState::new(metadata, &backend);
+        let precompiles = MetachainPrecompiles;
+        let mut executor = StackExecutor::new_with_precompiles(state, &CONFIG, &precompiles);
+        let mut al_executor = StackExecutor::new_with_precompiles(al_state, &CONFIG, &precompiles);
+
+        let mut listener = crate::eventlistener::StorageAccessListener::new();
+
+        runtime_using(&mut listener, move || {
```
sieniven, line 866: Does the RPC depend on the execution result of the transaction call?

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
...
+            let access_list = access_list
+                .into_iter()
+                .map(|x| (x.address, x.storage_keys))
+                .collect::<Vec<_>>();
+
+            executor.transact_call(
+                caller,
+                to.unwrap(),
```
sieniven, line 874: Unsafe unwrap here

lib/ain-evm/src/core.rs
```diff
@@ -816,6 +816,91 @@ impl EVMCoreService {
...
+        let al: AccessList = listener
+            .access_list
+            .into_iter()
+            .map(|(address, storage_keys)| AccessListItem {
+                address,
+                storage_keys: Vec::from_iter(storage_keys),
+            })
+            .collect();
+
+        al_executor.transact_call(
```
sieniven, line 891: Why are we executing another transaction call here while not doing anything with the execution results?

sieniven (changes requested):

lib/ain-evm/src/core.rs
shohamc1: https://github.com/DeFiCh/ain/blob/04c36dc6d08082d1d60bc0822a70151213580df4/lib/ain-evm/src/core.rs#L924

It is part of the results returned to the user.

lib/ain-evm/src/core.rs
shohamc1: While I agree, this is beyond the scope for this PR since `create_access_list` uses multiple executors which could introduce some complexity to moving them inside the executor.

lib/ain-evm/src/core.rs
shohamc1: https://github.com/DeFiCh/ain/blob/117b9cd26aa25a4b8813f4b967670eb759cb2b28/lib/ain-evm/src/core.rs#L888-L895

lib/ain-evm/src/core.rs
shohamc1: Addressed in [117b9cd](https://github.com/DeFiCh/ain/pull/2770/commits/117b9cd26aa25a4b8813f4b967670eb759cb2b28)

lib/ain-evm/src/core.rs
sieniven: This doesnt resolve anything. The result of the transaction_call from the executor is never processed.

lib/ain-evm/src/core.rs
sieniven: Likewise here

lib/ain-evm/src/core.rs
sieniven: We are making some assumptions that the execution result is successful which this pipeline isnt covered

lib/ain-evm/src/core.rs
sieniven: Resolved in PR #2824

lib/ain-evm/src/core.rs
```diff
@@ -727,6 +729,113 @@ impl EVMCoreService {
...
+                    gas_limit,
+                    al.iter()
+                        .map(|item| (item.address, item.storage_keys.clone()))
+                        .collect(),
+                );
+            }
+        };
+
+        Ok((al, al_executor.used_gas()))
+    }
```
sieniven, line 838: Why is there a need to re-execute the tx on the executor instance again?

sieniven: From what I am seeing, the results are never propagated either way from this execution nor does it change the access list obtained from the intial runtime listener that accumulates the access list.

sieniven: Ah, sorry cancel what I am saying. It is to obtain the execution gas usage with the access list populated.

sieniven:
Closing this PR since https://github.com/DeFiCh/ain/pull/2827 fixes multiple issues on this PR.

PR #2774, opened 2024-01-10
EVM: Add pending transactions filter

sieniven:
## Summary

- Add support for pending transcations filter to the current filters system.
- Note that due to mempool design, we do not currently track pending/queued evm transaction state in the mempool. Pending transactions using the filter changes will also include queued transactions.
- Shift in-memory signed/validation tx caches out of EVM core service for filters service to interface with it.
- Add documetation for filters system.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-grpc/src/rpc/eth.rs
```diff
@@ -1125,14 +1129,10 @@ impl MetachainRPCServer for MetachainRPCModule {
         let logs = self
             .handler
             .filters
-            .get_filter_logs_from_id(filter_id, curr_block)
+            .get_logs_filter_from_id(filter_id, curr_block)
```
shohamc1, line 1132: Doesn't the new name imply that we are getting the filter, but we are actually getting the logs of the filter here? Perhaps `get_logs_from_filter_id` would be best.

Jouzo (approved):

PR #2775, opened 2024-01-10
evm: Propagate `updatetoken` changes to EVM

shohamc1:


test/functional/feature_evm_dst20.py
```diff
@@ -988,6 +988,34 @@ def test_dst20_back_and_forth(self):
             Decimal(0),
         )
 
+    def test_update_token(self):
+        self.node.updatetoken("USDT", {"name": "USDT"})
```
sieniven, line 992: Can we please do a state rollback so that each functional test methods are indepedent of the chain state from each other?

prasannavl:
Superseded by https://github.com/DeFiCh/ain/pull/2850

PR #2783, opened 2024-01-17
Ensure EVM database integrity and consistency with DVM database

sieniven:
## Summary

This PR fixes issues that cause database integrity issues on the EVM side, resulting in a corrupted rocksdb.
1. ADD EVM db flush in DVM's FlushStateToDisk pipeline to keep EVM and DVM DB flushes as consistent as possible.
2. Expose EVM db flush on FFI.
3. Propagate errors on EVM services shutdown pipeline correctly and do not panic the node on failed DB flush.
4. Add pipeline in init to check for DVM and EVM database state inconsistencies. If state inconsistency is present, the startup pipeline will trigger re-indexing pipeline.
5. Propagate errors on RPC servers shutdown pipeline correctly instead of panicking the node. Previous pipeline panics the node early which prevents rocksdb from properly flushing in-memory buffer on shutdown, causing db corruption on subsequent startups.
6. Propagate errors from RPC servers correctly and set initialization failure instead of panicking the node on start up.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-evm/src/evm.rs
+++ b/lib/ain-evm/src/evm.rs
@@ -29,7 +29,10 @@ use crate::{
     filters::FilterService,
     log::{LogService, Notification},
     receipt::ReceiptService,
-    storage::{traits::BlockStorage, Storage},
+    storage::{
+        traits::{BlockStorage, FlushableStorage},
+        Storage,
+    },
     transaction::{cache::TransactionCache, SignedTx},
     trie::GENESIS_STATE_ROOT,
     Result,
@@ -269,7 +272,7 @@ impl EVMServices {
             .sender
             .send(Notification::Block(block.header.hash()))
             .map_err(|e| format_err!(e.to_string()))?;
-        
+
         self.core.clear_account_nonce();
         self.core.flush()?;
         self.storage.flush()?;
--- a/lib/ain-evm/src/evm.rs
+++ b/lib/ain-evm/src/evm.rs
@@ -272,10 +272,7 @@ impl EVMServices {
             .sender
             .send(Notification::Block(block.header.hash()))
             .map_err(|e| format_err!(e.to_string()))?;
-
         self.core.clear_account_nonce();
-        self.core.flush()?;
-        self.storage.flush()?;
         Ok(())
     }
 
@@ -466,6 +463,17 @@ impl EVMServices {
         }
         template.backend.increase_tx_count();
         Ok(())
+    }
+
+    ///
+    /// # Safety
+    ///
+    /// Result cannot be used safety unless `cs_main` lock is taken on C++ side
+    /// across all usages. Note: To be replaced with a proper lock flow later.
+    ///
+    pub unsafe fn flush_state_to_db(&self) -> Result<()> {
+        self.core.flush()?;
+        self.storage.flush()
     }
 }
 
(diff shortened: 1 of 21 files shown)


src/init.cpp
```diff
@@ -2044,6 +2044,32 @@ bool AppInitMain(InitInterfaces& interfaces)
...
+                    // After EVM activation
+                    auto evmBlockHash = uint256::FromByteArray(*res).GetHex();
+                    auto dvmBlockHash = pcustomcsview->GetVMDomainBlockEdge(VMDomainEdge::EVMToDVM, evmBlockHash);
+                    if (!dvmBlockHash.val.has_value()) {
+                        strLoadError = _("Unable to get DVM block hash from latest EVM block hash, inconsistent chainstate detected. "
+                                        "This may be due to corrupted block databases between DVM and EVM, and you will need to "
+                                        "rebuild the database using -reindex.").translated;
+                        break;
+                    }
+                    CBlockIndex *pindex = LookupBlockIndex(uint256S(*dvmBlockHash.val));
```
Bushstar, line 2061: Check that pindex is not a nullptr.

sieniven: Resolved in d48109084b

Bushstar (approved):

lib/ain-evm/src/evm.rs
```diff
@@ -252,27 +255,27 @@ impl EVMServices {
...
+            .generate_logs_from_receipts(&receipts, block.header.number)?;
+        self.receipt.put_receipts(receipts)?;
+        self.channel
+            .sender
+            .send(Notification::Block(block.header.hash()))
+            .map_err(|e| format_err!(e.to_string()))?;
+
+        self.core.clear_account_nonce();
+        self.core.flush()?;
+        self.storage.flush()?;
```
Jouzo, line 278: Any idea of the impact on sync ? We could do it periodically instead of every block, especially since we add a consistency check with the DVM db

Jouzo (approved):

lib/ain-evm/src/evm.rs
sieniven: Didnt check the perf tradeoff on the sync for this, but we likely may need to do this since we implement a flush on the DVM side on every state update from connect block. If flush happens periodically, unpredictable node crashes will cause state mismatches between the DVM and EVM side, which i guess this PR tradeoffs the consistency for performance

sieniven: Not sure what are your thoughts on this tradeoff?

sieniven: Although on init, we do check for state mismatch between the 2 dbs, but the current pipeline goes into reindex which wont be ideal at all.

Jouzo: Sounds good/safe to me.

sieniven:
EVM db flush is shifted out of every tip change on the rust side with commit c86861f. Instead, the EVM db flush is now exposed on the FFI, and is called in the same pipeline on the DVM side to keep the DVM and EVM databases consistent in FlushStateToDisk.

Bushstar (approved):

Jouzo (approved):

PR #2784, opened 2024-01-17
RPC: Fix eth_feeHistory pipeline

sieniven:
## Summary

- PR contains the fixes to correct the erroneous fee_history pipeline.
- Check block range argument limits passed in the RPC.
- Check priority fee percentile vector to be monotonically increasing and not exceed max vec size limit.
- Proper priority fee calculation from the block transactions, taking to account the block base fee and the maxFeePerGas limit set for EIP1559 transactions.
- Proper priority fee calculation for non-EIP1559 transactions (gasPrice - block base fee).
- Clean up and add more test coverage to feature_evm_rpc_fee_history functional test.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-evm/src/block.rs
+++ b/lib/ain-evm/src/block.rs
@@ -198,7 +198,7 @@ impl BlockService {
         &self,
         block_count: U256,
         highest_block: U256,
-        priority_fee_percentile: Option<Vec<usize>>,
+        priority_fee_percentile: Vec<usize>,
         block_gas_target_factor: u64,
     ) -> Result<FeeHistoryData> {
         // Validate block_count input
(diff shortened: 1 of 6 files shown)


lib/ain-evm/src/block.rs
```diff
@@ -182,136 +191,146 @@ impl BlockService {
...
-            */
-
-            let mut reward = Vec::new();
-
-            for block_eip_tx in eip_transactions {
-                if block_eip_tx.is_empty() {
-                    reward.push(vec![U256::zero()]);
-                    continue;
-                }
+        let mut blocks = Vec::new();
```
Jouzo, line 249: Restoring previous init
```suggestion
        let mut blocks = Vec::with_capacity(block_count);
```

sieniven: Resolved in commit d14f60b0db

Jouzo (approved):

PR #2812, opened 2024-02-02
RPC: Fix eth subscription notifications

sieniven:
## Summary

This PR fixes multiple stability issues on the node's subscription system.
- Switch from unbounded mpsc channel to bounded broadcast channel to safeguard against bloated buffer.
- Resolve incorrect pipeline of single consumer, used on subscription threads.
- Implement broadcast receivers in every subscription thread, and trigger response on notifications.
- Correct handling of errors in each subscription thread, and propagate send results of subscription sink in each thread to ensure proper shutdown.
- Shift runtime of subscriptions thread to EVM service threadpool.
- Refactor and clean up param structs.
- Add config to set notification channel's buffer size (-evmnotificationchannel), defaults to 10_000.
- Add config to turn on/off Eth subscription notification server (-ethsubscription), defaults to true.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


lib/ain-grpc/src/subscription/eth.rs
```diff
@@ -43,148 +48,159 @@ impl MetachainPubSubServer for MetachainPubSubModule {
...
+                        let (current, highest) = ain_cpp_imports::get_sync_status()
+                            .map_err(|_| format_err!("failed to get sync status"))?;
+                        let current_block = handler
+                            .storage
+                            .get_latest_block()?
+                            .ok_or(format_err!("Unable to find latest block"))?
+                            .header
+                            .number;
+                        let diff = U256::from(highest.saturating_sub(current));
+                        let highest_block = current_block.saturating_add(diff);
```
Jouzo, line 163: Shouldn't we use highest directly here ?

sieniven: Nope, highest refers to the highest DVM block height, while we are trying to get the highest EVM block height that the node is trying to sync to.

Jouzo (approved):

PR #2816, opened 2024-02-07
Automatic Setting of Negative Interest

Bushstar:
## Summary

- Set the negative interest rate automatically.
- Introduces a new RPC call to estimate the negative interest rate.

## RPCs

- New estimatenegativeinterest RPC call
```
defi-cli estimatenegativeinterest
{
  "dusdBurned": 2463205.39186885,
  "dusdLoaned": 94262062.35901081,
  "negativeInterest": -15.88792800
}
```
## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None

### DIFF
--- a/src/dfi/govvariables/attributes.cpp
+++ b/src/dfi/govvariables/attributes.cpp
@@ -61,33 +61,35 @@ const std::map<uint8_t, std::string> &ATTRIBUTES::displayVersions() {
 
 const std::map<std::string, uint8_t> &ATTRIBUTES::allowedTypes() {
     static const std::map<std::string, uint8_t> types{
-        {"locks",          AttributeTypes::Locks     },
-        {"oracles",        AttributeTypes::Oracles   },
-        {"params",         AttributeTypes::Param     },
-        {"poolpairs",      AttributeTypes::Poolpairs },
-        {"token",          AttributeTypes::Token     },
-        {"gov",            AttributeTypes::Governance},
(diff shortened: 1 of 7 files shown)


kuegi:
IMHO this way of doing it can not work and should not be merged.

The goal is to payout half of the stabilization fee. It is currently done as a workaround manually by setting the interest rate of DUSD negatively. The workaround is done via averages and max values cause of the limitations of the manualy workaround.

If the goal of this PR is to stop the need for manual setting of values, keep in mind that this will be in the code forever. I am not sure if its worth it to add such code and the according risk and maintenance need.

If we decide to accept that, we still need to rework this IMHO:

do not base calculations on time because of the workaround. (In theory it would be best to just distribute the burn directly, even if we do it via negative interest rates. But I understand that users are now used to small volatility which would change if we go from average to daily direct update)

We now have an additional discount fee, with "messes" this calculation up:
The DUSD fee is now the combination of `basefee` and `discount fee`

BaseFee is the stabilization fee which is based on the algo ratio, currently set to a fixed value of 30%. half of the DUSD burned via this fee is supposed to be distributed to loan holders

DiscountFee is  an additional fee which is burned completely. this fee is set based on the DUSD price in the stablecoin pools.

So IF we want to go forward with this PR, we need to add parameters telling the chain how much of the current fee should be distributed in payout. and the node summing it up accordingly each block.

prasannavl:
Closing as no longer relevant.

PR #2824, opened 2024-02-18
EVM: Refactor trace tx pipeline into executor

sieniven:
## Summary

- Resolves issue raised in https://github.com/DeFiCh/ain/pull/2770#discussion_r1449950193
- Shifts execution tx with trace into ain executor. EVM executor instance should be abstracted from caller and inside ain executor with this PR refactor

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/lib/ain-evm/src/executor.rs
+++ b/lib/ain-evm/src/executor.rs
@@ -249,7 +249,7 @@ impl<'backend> AinExecutor<'backend> {
         ))
     }
 
-    pub fn exec_trace_tx(
+    pub fn exec_with_tracer(
         &mut self,
         signed_tx: &SignedTx,
     ) -> Result<(Vec<ExecutionStep>, bool, Vec<u8>, u64)> {
--- a/lib/ain-evm/src/executor.rs
+++ b/lib/ain-evm/src/executor.rs
@@ -5,8 +5,10 @@ use ethereum_types::{Bloom, H160, H256, U256};
 use evm::{
     backend::{ApplyBackend, Backend},
     executor::stack::{MemoryStackState, StackExecutor, StackSubstateMetadata},
+    gasometer::tracing::using as gas_using,
     Config, CreateScheme, ExitReason,
 };
+use evm_runtime::tracing::using as runtime_using;
 use log::{debug, trace};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


lib/ain-evm/src/executor.rs
```diff
@@ -246,6 +249,67 @@ impl<'backend> AinExecutor<'backend> {
...
     }
 
+    pub fn exec_with_tracer(
+        &mut self,
+        signed_tx: &SignedTx,
+    ) -> Result<(Vec<ExecutionStep>, bool, Vec<u8>, u64)> {
+        let to = signed_tx.to().ok_or(format_err!(
+            "debug_traceTransaction does not support contract creation transactions",
+        ))?;
+        self.backend.update_vicinity_from_tx(signed_tx)?;
```
Jouzo, line 259: Could we lift this so we can remove the `&mut self` and rename it to `call_with_tracer`. This is replaying a TX and should be a read-only call.

lib/ain-evm/src/executor.rs
```diff
@@ -246,6 +249,67 @@ impl<'backend> AinExecutor<'backend> {
...
+        trace!(
+            "[Executor] Executing trace EVM TX with vicinity : {:?}",
+            self.backend.vicinity
+        );
+        let ctx = ExecutorContext {
+            caller: signed_tx.sender,
+            to: Some(to),
+            value: signed_tx.value(),
+            data: signed_tx.data(),
+            gas_limit: u64::try_from(signed_tx.gas_limit())?, // Sets to u64 if overflows
```
Jouzo, line 269: Don't we want to unwrap_or(u64::MAX) in line with the comment ?

sieniven: The comments were inaccurate, my bad. We should propagate the overflow error (which shouldnt ever happen) back to the RPC call.

lib/ain-evm/src/executor.rs
sieniven: Makes sense.

sieniven: Resolved in 35e1f6caa6

Jouzo (approved):

PR #2833, opened 2024-02-26
Fix: ocean api tx get vins

canonbrother:
## Summary

- Ocean get vins api


canonbrother:
missing few handling on get_vouts.. mv it to other PR

PR #2834, opened 2024-02-27
Fix: ocean api tx get vouts

canonbrother:
## Summary

- as per title

#### checked
```
getVouts
    ✓ should return list of vout (11 ms)
    ✓ should return empty page if txid is not valid (2 ms)
    ○ skipped should return list of vout when next is out of range
```

### DIFF
--- a/lib/ain-ocean/src/api/transactions.rs
+++ b/lib/ain-ocean/src/api/transactions.rs
@@ -74,18 +74,13 @@ async fn get_vouts(
     Query(query): Query<PaginationQuery>,
     Extension(ctx): Extension<Arc<AppContext>>,
 ) -> Result<ApiPagedResponse<TransactionVout>> {
-    let next = query
-        .next
-        .as_deref()
-        .unwrap_or("0")
-        .parse::<usize>()?;
-    let vout_id = format!("{id}{next:08x}");
+    let next = query.next.as_deref().unwrap_or("0").parse::<usize>()?;
 
     let list = ctx
         .services
         .transaction
         .vout_by_id
-        .list(Some(vout_id), SortOrder::Descending)?
+        .list(Some((id, next)), SortOrder::Descending)?
         .paginate(&query)
         .take_while(|item| match item {
             Ok((_, vout)) => vout.txid == id,
--- a/lib/ain-ocean/src/api/transactions.rs
+++ b/lib/ain-ocean/src/api/transactions.rs
@@ -74,11 +74,14 @@ async fn get_vouts(
     Query(query): Query<PaginationQuery>,
     Extension(ctx): Extension<Arc<AppContext>>,
 ) -> Result<ApiPagedResponse<TransactionVout>> {
+    let next = query.next.clone().unwrap_or(format!("0"));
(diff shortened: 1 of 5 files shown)


lib/ain-ocean/src/api/transactions.rs
```diff
@@ -74,11 +74,18 @@ async fn get_vouts(
     Query(query): Query<PaginationQuery>,
     Extension(ctx): Extension<Arc<AppContext>>,
 ) -> Result<ApiPagedResponse<TransactionVout>> {
+    let next = query
+        .next
+        .clone()
+        .unwrap_or(format!("0"))
+        .parse::<usize>()?;
```
Jouzo, line 81:
```suggestion
let next = query
    .next
    .as_ref()
    .map(String::as_str)
    .unwrap_or("0")
    .parse::<usize>()?;
```

lib/ain-ocean/src/model/transaction_vout.rs
```diff
@@ -1,10 +1,9 @@
 use bitcoin::Txid;
 use serde::{Deserialize, Serialize};
 
-pub type TransactionVoutKey = (Txid, usize);
-
 #[derive(Debug, Serialize, Deserialize)]
 pub struct TransactionVout {
+    pub id: String, // format!({}{:08x}, txid, usize")
```
Jouzo, line 6: Why are we reverting back to String here ?

canonbrother: the old one usize wasn't padded
references: https://github.com/BirthdayResearch/jellyfishsdk/blob/767aa7199458153486350a6804fa0851932d91ae/apps/whale-api/src/module.model/_hex.encoder.ts#L37

lib/ain-ocean/src/api/transactions.rs
canonbrother: fixed at [f06e14e](https://github.com/DeFiCh/ain/pull/2834/commits/f06e14ea39dd7af7f86a961499224c38743375e7)

clippy updates:
```
let next = query
        .next
        .as_deref()
        .unwrap_or("0")
        .parse::<usize>()?;
```

lib/ain-ocean/src/model/transaction_vout.rs
Jouzo: We should fully implement `Column` trait for `TransactionVout` column instead

Jouzo: Can check `TransactionByBlockHash` implementation for ref

lib/ain-ocean/src/api/transactions.rs
Jouzo: Even better 👍

lib/ain-ocean/src/model/transaction_vout.rs
canonbrother: fixed at [feac743](https://github.com/DeFiCh/ain/pull/2834/commits/feac74377bf6dc59556672799e6c9d744ec132ed)

PR #2835, opened 2024-02-27
fix: ocean api typings token

canonbrother:
## Summary

- [x] api typings correction
- [ ] Vec<TokenData> sort by id <-- `todo: rm sort and use BTreeMap on TokenResult`
- [ ] return `undefined`


canonbrother:
refer to https://github.com/DeFiCh/ain/pull/2838

PR #2843, opened 2024-03-04
Future Swap limitation

Bushstar:
## Summary

- Limits the amount that can be swapped from loan token to DUSD to 10% of the 28 day average liquidity of that loan token in its DUSD pool. Actual percentage and period is configurable, but defaults to the values above.
- Average liquidity will only populate on a Future Swap event if there is enough liquidity data present, which there should be as long as the limitation has been enabled for the full 28 days period before the event.
- If there is no average liquidity present then Future Swap will behave as normal without any limits.
- Implements https://github.com/DeFiCh/dfips/issues/238

## RPCs

- `listloantokenliquidity` returns a list of all present average liquidity.

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


src/dfi/validation.cpp
```diff
@@ -846,47 +848,116 @@ static void ProcessLoanEvents(const CBlockIndex *pindex, CCustomCSView &cache, c
...
+    CDataStructureV0 activeKey{AttributeTypes::Param, ParamIDs::DFIP2211F, DFIPKeys::Active};
+    if (!attributes->GetValue(activeKey, false)) {
         return;
     }
 
-    CDataStructureV0 startKey{AttributeTypes::Param, ParamIDs::DFIP2203, DFIPKeys::StartBlock};
-    const auto startBlock = attributes->GetValue(startKey, CAmount{});
-    if (pindex->nHeight < startBlock) {
+    CDataStructureV0 blockKey{AttributeTypes::Param, ParamIDs::DFIP2211F, DFIPKeys::BlockPeriod};
+    const auto blockPeriod = attributes->GetValue(blockKey, DEFUALT_FS_LIQUIDITY_BLOCK_PERIOD);
```
kuegi, line 868: typo DEFUALT_...

src/dfi/validation.cpp
```diff
@@ -846,47 +848,116 @@ static void ProcessLoanEvents(const CBlockIndex *pindex, CCustomCSView &cache, c
...
         return;
     }
 
-    const auto blockPeriod = attributes->GetValue(blockKey, CAmount{});
-    if ((pindex->nHeight - startBlock) % blockPeriod != 0) {
+    const auto &dusdID = dusdToken->first;
+
+    std::set<DCT_ID> tokens;
+    for (const auto &[id, loanToken] : loanTokens) {
+        tokens.insert(id);
```
kuegi, line 879: imho need to filter DUSD here. its used as "tokens == dTokens, not DUSD", but DUSD is also a loantoken. right?

test/functional/feature_future_swap_limitation.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # Try and swap above new limit
+        assert_raises_rpc_error(
+            -32600,
+            "Swap amount exceeds 10% of average pool liquidity limit. Available amount to swap: 7.24862451@META",
+            self.nodes[0].futureswap,
+            self.address,
+            "7.24862452@META",
+        )
+
```
kuegi, line 289: missing test to swap DUSD->META. limitation should apply on both directions.

kuegi (changes requested):

src/dfi/validation.cpp
```diff
@@ -896,9 +899,15 @@ static void LiquidityForFuturesLimit(const CBlockIndex *pindex,
...
+            cache.SetLoanTokenLiquidityPerBlock(
+                {static_cast<uint32_t>(pindex->nHeight), poolPair.idTokenA.v, poolPair.idTokenB.v}, poolPair.reserveA);
+            cache.SetLoanTokenLiquidityPerBlock(
+                {static_cast<uint32_t>(pindex->nHeight), poolPair.idTokenB.v, poolPair.idTokenA.v}, poolPair.reserveB);
         } else {
-            cache.SetLoanTokenLiquidityPerBlock(pindex->nHeight, poolPair.idTokenB.v, poolPair.reserveB);
+            cache.SetLoanTokenLiquidityPerBlock(
+                {static_cast<uint32_t>(pindex->nHeight), poolPair.idTokenB.v, poolPair.idTokenA.v}, poolPair.reserveB);
+            cache.SetLoanTokenLiquidityPerBlock(
+                {static_cast<uint32_t>(pindex->nHeight), poolPair.idTokenA.v, poolPair.idTokenB.v}, poolPair.reserveA);
```
kuegi, line 910: if is unnecessary. both blocks do exactly the same.

kuegi (commented):
small comment, rest looks good to me

src/dfi/validation.cpp
```diff
@@ -846,47 +848,118 @@ static void ProcessLoanEvents(const CBlockIndex *pindex, CCustomCSView &cache, c
...
+        const auto dusdB = poolPair.idTokenB == dusdID;
+        if (!dusdA && !dusdB) {
+            return true;
+        }
+
+        cache.SetLoanTokenLiquidityPerBlock(
+            {static_cast<uint32_t>(pindex->nHeight), poolPair.idTokenA.v, poolPair.idTokenB.v}, poolPair.reserveA);
+        cache.SetLoanTokenLiquidityPerBlock(
+            {static_cast<uint32_t>(pindex->nHeight), poolPair.idTokenB.v, poolPair.idTokenA.v}, poolPair.reserveB);
+
```
kuegi, line 905: @Bushstar: I just realized: we no longer need to remove dusd from the tokens (and even get the Id).
We now save loanTokenLiquidity in both directions, so no more need to check "token" vs "dusd". just check "both sides loantoken (inkl dusd)?"

so keep dusd in tokens and do
```
if(!tokens.count(poolPair.idTokenA) || !tokens.count(poolPair.idTokenB)) {
  return true;
}
// its a loanToken-loanToken pool -> save the liquidity of both sides
cache.SetLoan....
```

Bushstar: I'd prefer to explicitly check that one pair is DUSD in case we add loan tokens pool pairs that do not include DUSD as these will not be used in future swaps.

kuegi: valid point. would not hurt (cause it would be just unnecessary data) to store it, but also doesn't hurt to check it here.

prasannavl:
The cost of doing these calc at every block concerns me. This is going to stall the chain esp. on sync catch-up at this block range, which cannot be fixed after the fact as well. 

We might likely want to just to do the calc at periodic intervals. I don't think accuracy is necessary here.

kuegi:
> The cost of doing these calc at every block concerns me. This is going to stall the chain esp. on sync catch-up at this block range, which cannot be fixed after the fact as well.
> 
> We might likely want to just to do the calc at periodic intervals. I don't think accuracy is necessary here.

valid point. how about doing it every price block (so every 120 blocks)?

prasannavl:
Indeed, that would be perfect. We'll add var that makes the time period changeable.

PR #2850, opened 2024-03-07
Support token split on EVM

Bushstar:
## Summary

- Adds support to split tokens on EVM. There are three options available.
  - Split tokens via a new precompile.
  - Use the DFI Intrinsics V2 contract which makes use of the new precompile for the user.
  - Use Transfer Domain to move the tokens out to DVM.
- For DFI Intrinsic usage please check the test in this PR. For precompile usage please check DFIIntrinsicsV2.sol.

## Implications

- Storage
  - [x] None

- Consensus
  - [ ] None


test/functional/feature_evm_token_split.py
```diff
@@ -0,0 +1,768 @@
...
+                    "singlekeycheck": False,
+                }
+            ]
+        )
+        self.nodes[0].generate(1)
+
+        # Check META balance
+        assert_equal(
+            self.nodes[0].getaccount(destination_address)[0], "40.00000000@META"
+        )
```
kuegi, line 372: i would also add a check that there was no META/v1 added on DVM side. for safety sake.

also check please check that the gov-economy values for TD (transfered to EVM and transfered back) are updated. should have the META/v1 reduced now.

test/functional/feature_evm_token_split.py
```diff
@@ -0,0 +1,768 @@
...
+            self.contract_address_metav1,
+            self.contract_address_metav2,
+            amount,
+            split_multiplier,
+        )
+
+        # Check minted balance
+        assert_equal(
+            self.nodes[0].gettoken("META")[f"{self.meta_final_id}"]["minted"],
+            Decimal(1960.00000000) + (amount * split_multiplier),
```
kuegi, line 524: also check that economy values about the TD are updated correctly. Since its changed from META/v1 to META, the according amounts in the economy need an update. (META/v1 dvm->evm reduces and META dvm->evm increased)

lib/Cargo.toml
```diff
@@ -88,7 +90,7 @@ spin = "0.9.8"
 ### eth
 
 evm = { version = "0.39", default-features = false, features = ["with-serde", "tracing"] }
-evm-runtime = "0.39.0"
+evm-runtime = { version  = "0.39.0" }
```
sieniven, line 93: We can just specify the version if no other requirements are needed

lib/ain-cpp-imports/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 use std::error::Error;
 
 #[cfg(not(test))]
-mod bridge;
+pub mod bridge;
```
sieniven, line 4: We should keep the bridge module private, but expose the defined TokenAmount struct in the cpp-imports module instead

lib/ain-evm/src/precompiles/token_split.rs
```diff
@@ -0,0 +1,220 @@
+use std::collections::BTreeMap;
+
+use ain_contracts::{dst20_address_from_token_id, validate_split_tokens_input, TokenSplitParams};
+use ain_cpp_imports::{bridge::ffi::TokenAmount, split_tokens_from_evm};
```
sieniven, line 4:
```suggestion
use ain_cpp_imports::{TokenAmount, split_tokens_from_evm};
```

sieniven (approved):

sieniven (dismissed):

sieniven (approved):

lib/Cargo.toml
sieniven: Resolved in [7340314](https://github.com/DeFiCh/ain/pull/2850/commits/7340314660205f1a7c8d0b85a863d404d7975243)

lib/ain-cpp-imports/src/lib.rs
sieniven: Resolved in [7340314](https://github.com/DeFiCh/ain/pull/2850/commits/7340314660205f1a7c8d0b85a863d404d7975243)

lib/ain-evm/src/precompiles/token_split.rs
sieniven: Resolved in [7340314](https://github.com/DeFiCh/ain/pull/2850/commits/7340314660205f1a7c8d0b85a863d404d7975243)

prasannavl:
Looks great. 

One detail to resolve before merging in would be - the economy variables. They are currently updated transparently when gone through TD. However, when gone through the contract call, TDs economy vars becomes challenging to account for as they are not  updated. This will create a minor divergence in the accounting system based on how the splits are executed. 

Options:
 
1. Leave it as is.
  - Pros: Now, transferdomain economy vars will only indicate actual transfer domain calls.
  - Cons: Have to account across all transfer domain v1s and v2s to check sanity. Quickly makes the whole thing complex.
2. Update transfer domain economy vars as a part of split token EVM intrinsic call so that, when a migration happens, they are reduced out on the old and added to the new.
  - Pros: This will preserve easier accounting. Once split, the indicators will move to the latest version. 
  - Cons: will no longer hold actual transfer domain call indicators.
  
2 could be a saner route -- basically an auto transferdomain out and in on migration conceptually.

prasannavl (approved):

prasannavl (approved):

prasannavl:
Jellyfish failures still due to https://github.com/DeFiCh/ain/pull/2821 (pre-jellyfish runner fixes). Merging to resolve them directly on master.

PR #2865, opened 2024-03-19
Run Jellyfish Tests in parallel

canonbrother:
## Summary

- speed up jellyfish ci by running with 10 instances, from (approximately) 2 hrs to 40 (30+15) mins

### DIFF
--- a/.github/workflows/tests-jellyfish.yml
+++ b/.github/workflows/tests-jellyfish.yml
@@ -43,4 +43,4 @@ jobs:
         run: |
           npm ci
           npm run compile
-          DEFICHAIN_DOCKER_IMAGE=test-build-container npm run ci:test -- --testTimeout=600000
+          DEFICHAIN_DOCKER_IMAGE=test-build-container npm run ci:test -- --testTimeout=1800000
--- a/.github/workflows/tests-jellyfish.yml
+++ b/.github/workflows/tests-jellyfish.yml
@@ -11,12 +11,35 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
 
 env:
-  BUILD_VERSION: latest # Computed
+  DOCKER_HUB_USER: defi
+  DOCKER_IMAGE: defi/defichain:jellyfish-ci-${GITHUB_SHA:0:10}
 
 jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Populate environment
+        run: ./make.sh ci-export-vars
+
+      - name: Build docker image
+        run: docker build -t ${{ env.DOCKER_IMAGE }} -f ./contrib/dockerfiles/x86_64-pc-linux-gnu.dockerfile .
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ env.DOCKER_HUB_USER }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+
+      - name: Push to Docker Hub
+        run: docker push ${{ env.DOCKER_IMAGE }}
+
   test:
     name: Test
     runs-on: ubuntu-latest
+    needs: [docker-build]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


sieniven (commented):
Am I missing something - I dont exactly see how running 10 instances of the same jellyfish test will speed up the CI runtime?

canonbrother:
> Am I missing something - I dont exactly see how running 10 instances of the same jellyfish test will speed up the CI runtime?

here for your reference about to compare before/after the total duration
**before** https://github.com/DeFiCh/ain/actions/runs/8337022104
**after** https://github.com/DeFiCh/ain/actions/runs/8338085780/job/22817831027?pr=2865

Jouzo:
`jest.sequencer.js`

```

class CustomSequencer extends Sequencer {
  sort (tests) {
    // In CI, GitHub Actions tests are parallelized across multiple instance
    if (process.env.GH_INSTANCE_TOTAL) {
      const total = parseInt(process.env.GH_INSTANCE_TOTAL, 10)
      const index = parseInt(process.env.GH_INSTANCE_INDEX, 10)

      return ShuffleSeed.shuffle(tests, 'deterministic')
        .filter((_, i) => i % total === index)
    }

    return tests
  }
}
```
That's where the magic happens

Jouzo (dismissed):

sieniven (commented):
If this is the case, we should then separate the docker build stage and then run the tests in 10 instances instead. Does not make sense to block 10 runners while building the same image on every instance.

canonbrother:
sigh.. the `workflow_run` is not supported on PR

Jouzo:
Great improvements, thanks @canonbrother !

Jouzo (dismissed):

.github/workflows/tests-jellyfish.yml
```diff
@@ -11,12 +11,42 @@ concurrency:
...
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ env.DOCKER_HUB_USER }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+
+      - name: Push to Docker Hub
+        run: docker push ${{ env.DOCKER_IMAGE }}
+
```
sieniven, line 38:
```suggestion
  build:
    runs-on: ubuntu-latest
    container: defi/ain-builder:latest
    env:
      CARGO_INCREMENTAL: 0

    steps:
    - uses: actions/checkout@v4
    - run: git config --global --add safe.directory '*'

    - name: Populate environment
      run: ./make.sh ci-export-vars

    - name: Setup dependencies
      run: ./make.sh ci-setup-deps

    - name: Setup user dependencies
      run: ./make.sh ci-setup-user-deps

    - name: Restore cpp build cache
      id: cpp-cache-restore
      uses: actions/cache/restore@v3
      with:
        path: |
          ./build/depends
          ./build/src
          ~/.ccache
        key: cpp-${{ env.TARGET }}-${{ env.BUILD_TYPE }}

    - name: Rust build cache
      uses: Swatinem/rust-cache@v2
      id: rust-cache-restore
      with:
        workspaces: lib -> ../build/lib/target
        save-if: ${{ github.ref == 'refs/heads/master' }}
        shared-key: ${{ env.TARGET }}

    - name: Build binaries
      run: ./make.sh build

    - name: Upload binaries
      uses: actions/upload-artifact@v3
      with:
        name: defi-bins
        path: |
          build/src/defid
          build/src/defi-cli
```

sieniven: There shouldnt be a need to push images to the registry since the test workflow is local. In addition, using docker build for this workflow will not be ideal since we cant leverage on the incremental builds on our cpp and rust side, as we cache already our builds to speed up our workflows. Using local builds for the jellyfish test will speed up the CI workflow even more

.github/workflows/tests-jellyfish.yml
```diff
@@ -32,15 +62,15 @@ jobs:
...
+      - name: Docker pull
+        run: |
+          docker pull ${{ env.DOCKER_IMAGE }}
 
       - name: Run tests
         run: |
           npm ci
           npm run compile
-          DEFICHAIN_DOCKER_IMAGE=test-build-container npm run ci:test -- --testTimeout=600000
+          DEFICHAIN_DOCKER_IMAGE=${{ env.DOCKER_IMAGE }} npm run ci:test -- --testTimeout=600000
```
sieniven, line 73: Might have to change passing defid as a docker image to instead running defid locally - dependent on the jellyfishsdk side of things

.github/workflows/tests-jellyfish.yml
sieniven: If jellyfishsdk is dependent on defid to be running in a container, then perhaps we can use defi.dockerfile to build our defid image instead. We can re-use the same pipeline to build the image, then upload the built image as an artifact.
 Ref: https://github.com/DeFiCh/ain/blob/cd853b625f41d8769eece8d725204b1e842689df/.github/workflows/build-dev.yaml#L101-L123

prasannavl (approved):

PR #2885, opened 2024-04-08
Fix: Loan token missing after creating pool pair

canonbrother:
## Summary

- Not a "bug".. perhaps we're missing the name checking on token-loanToken poolpair
- Slot a token-loanToken poolpair rename test


canonbrother:
close due to not a bug but test error

PR #2892, opened 2024-04-15
ocean indexing active price

nagarajm22:
## Summary

- This repo contains ocean active price indexing 
  
  ✓ should list (16 ms)
  ✓ should get ticker (3 ms)
  ✓ should get feeds (4 ms)
  ✓ should get oracles (4 ms)

## Storage

- active price indexing 

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None


lib/ain-ocean/src/bin/cli.rs
```diff
@@ -64,7 +64,7 @@ async fn main() -> Result<()> {
             .by_height
             .get_highest()?
             .map_or(0, |b| b.height);
-        let new_height = highest_block + 1;
+        let new_height: u32 = highest_block + 1;
```
Jouzo, line 67: Why not type infer here ?

lib/ain-ocean/src/indexer/loan_token.rs
```diff
@@ -0,0 +1,171 @@
...
+        services.oracle_price_active.by_id.delete(&ticker_id)?;
+        services
+            .oracle_price_active
+            .by_key
+            .delete(&(ticker_id.0, ticker_id.1))?;
+        Ok(())
+    }
+}
+pub fn aggregated_validate(aggrigated_price: OraclePriceAggregated, context: &Context) -> bool {
+    let minimum_live_oracles: i32 = 2;
```
Jouzo, line 142: Same as above

lib/ain-ocean/src/indexer/loan_token.rs
```diff
@@ -0,0 +1,171 @@
...
+        return false;
+    }
+
+    true
+}
+
+pub fn is_live(active: OraclePriceActiveActive, next: OraclePriceActiveNext) -> bool {
+    let deviation_threshold = Decimal::from_f64(0.5).unwrap_or_default();
+    let active_price = Decimal::from_str_exact(&active.amount).unwrap_or_default();
+    let next_price = Decimal::from_str_exact(&next.amount).unwrap_or_default();
```
Jouzo, line 160: Let's return a Result<bool> and fail instead of setting to default

lib/ain-ocean/src/indexer/oracle.rs
```diff
@@ -240,13 +240,26 @@ impl Index for UpdateOracle {
...
-                        match services.oracle_token_currency.by_id.delete(&deletion_key) {
+                        match services.oracle_token_currency.by_id.delete(&deletion_id) {
+                            Ok(_) => {
+                                // Successfully deleted
+                            }
+                            Err(err) => {
+                                let error_message = format!("Error:update oracle: {:?}", err);
+                                eprintln!("{}", error_message);
+                                return Err(Error::NotFound(NotFoundKind::Oracle));
+                            }
```
Jouzo, line 256: Why converting to a NotFound error ? The error could be db related as well

lib/ain-ocean/src/indexer/loan_token.rs
```diff
@@ -0,0 +1,189 @@
...
+        let abs_diff = diff.abs();
+        let threshold = active_price * deviation_threshold;
+        if abs_diff >= threshold {
+            return false;
+        }
+
+        true
+    } else {
+        false
+    }
```
Jouzo, line 188: We can return a Result<bool> here instead.

```suggestion
    let (Some(active), Some(next)) = (active, next) else { return Ok(false) };
    
    let active_price = Decimal::from_str_exact(&active.amount)?;
    let next_price = Decimal::from_str_exact(&next.amount)?;

    if active_price <= Decimal::zero() || next_price <= Decimal::zero() {
        return Ok(false);
    }

    let deviation_threshold = Decimal::new(5, 1); // This represents 0.5

    let diff = next_price - active_price;
    let abs_diff = diff.abs();
    let threshold = active_price * deviation_threshold;
    if abs_diff >= threshold {
        return Ok(false);
    }

    Ok(true)
```

We don't want to silence the error. An error here doesn't mean that the price is not live but that there was an error parsing the amounts

Jouzo (approved):

PR #2903, opened 2024-04-22
EVM: Add support for token split via DST20 v2 contract

Jouzo:
## Summary

- Introduces new DST20 v2 contract which updates the EVM token split mechanism such as :
   - Removes the need for intrinsic v2
   - Upgrades the DST20 contract to implement `ITokenSplit` interface which contains both `SplitResult` event and `split` method
   - Injects the new amount on the old contract instead of the message sender, which lets us send the token from the old contract to the message sender.
   - Goes through the `DST20(newContract).transfer` method to trigger a transfer event. This should now be correctly picked up by blockscout as a token transfer from the old contract to the message sender of the new amount of the new token

- Deploy new DST20 version on fork height df23 by updating the implementation slot
- Clean up of DST20 contract free functions in dedicated file and introduce `contract` directory


Jouzo:
Home tab:
Screenshot 2024-04-23 at 14 55 53

TX detail tab:
Screenshot 2024-04-23 at 14 55 58

Log tab:
Screenshot 2024-04-23 at 14 56 09

Function call tab:
Screenshot 2024-04-23 at 14 56 03

State changes tab:
Screenshot 2024-04-23 at 14 56 13

Issue #2909, opened 2024-05-03
EVM: token transfers on stock split (upgradeToken) misses inflow


kuegi:
on 4.1.0-beta3. 

when doing a stock split on EVM, the tokentransfers contain the creation of the new token, but miss the removal of the old token.
example: https://testnet-dmc.mydefichain.com:8444/tx/0xedb21367809270dd94b61c0320d0cac213c57499a4c051f14869669cc92367b1
This converts 1 TWTR/v1 (0xff0..0a) to 10 TWTR (0xff0...01c). but tokentransfers only include the transfer of 10 TWTR.

So any tool would miss the change in balance of TWTR/v1

Jouzo:
Good catch, the internal burn was overlooked. Should be fixed in #2911

PR #2911, opened 2024-05-06
EVM: DST20 v2 burn of sent amount

Jouzo:
## Summary

- Remove original contract storage update  from precompile
- Use internal DST20 _burn method instead
- Fix total supplies issues
- Add tests for burn event
- Fixes https://github.com/DeFiCh/ain/issues/2909

Screenshot 2024-05-06 at 11 30 03
Screenshot 2024-05-06 at 11 30 13

## Implications

- Consensus
  - [x] changi rollback

### DIFF
--- a/lib/ain-contracts/dst20_v2/DST20V2.sol
+++ b/lib/ain-contracts/dst20_v2/DST20V2.sol
@@ -588,6 +588,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata, IDST20Upgradeable {
 
         // Upgrade available
         if (newAmount != amount) {
+            _burn(msg.sender, amount);
             IERC20(newTokenContractAddress).transfer(msg.sender, newAmount);
         }
 
--- a/lib/ain-evm/src/precompiles/token_split.rs
+++ b/lib/ain-evm/src/precompiles/token_split.rs
@@ -121,28 +121,9 @@ impl DVMStatePrecompile for TokenSplit {
             reset_storage: false,
         };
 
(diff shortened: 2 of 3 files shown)


Bushstar:
If we are no longer going to zero the old token amount then we should not deduct that amount from the economy Gov vars as it still exists inside EVM on the burn address.

Jouzo:
We do zero it through calling the internal _burn method which handles deducting the total supply. You can see on the first screenshot that the META/v1 total supply is now correctly zeroed out

Jouzo:
```
function _burn(address account, uint256 amount) internal virtual {
        require(account != address(0), "ERC20: burn from the zero address");

        _beforeTokenTransfer(account, address(0), amount);

        uint256 accountBalance = _balances[account];
        require(accountBalance >= amount, "ERC20: burn amount exceeds balance");
        unchecked {
            _balances[account] = accountBalance - amount;
            // Overflow not possible: amount <= accountBalance <= totalSupply.
            _totalSupply -= amount;
        }

        emit Transfer(account, address(0), amount);

        _afterTokenTransfer(account, address(0), amount);
    }
 ```
 
 Our implementation consider the burn as a reduction of `totalSupply`. 
 It's our appreciation how sending to the burn address should be counted for the gov vars economy. I would say that deducting it sounds right since it won't be accessible. 
 
 `require(from != address(0), "ERC20: transfer from the zero address");` checks are scatted in the other methods and prevent the burn address from sending any amount

PR #2918, opened 2024-05-20
Ocean: list pool swap aggregates

canonbrother:
## Summary

Ocean: List pool swap aggregates

```
PASS apps/whale-api/src/module.api/__defid__/poolpairs.defid.ts (92.776 s)
  ✓ should show aggregated swaps for 24h and 30d (9740 ms)
```


src/init.cpp
```diff
@@ -2484,8 +2484,18 @@ bool AppInitMain(InitInterfaces& interfaces)
...
+                const auto token = pcustomcsview->GetToken(id);
+                if (token) {
+                    pools.push_back({id.v, pool.creationHeight});
+                };
+                return true;
+            }, {0});
+
         CrossBoundaryResult result;
-        ocean_index_block(result, b.write());
+        ocean_index_block(result, b.write(), pools);
```
canonbrother, line 2498: ideally inject json rpc client but `// Error: Syn(Error("async function is not directly supported yet, but see https://cxx.rs/async.html for a working approach,
// and https://github.com/pcwalton/cxx-async for some helpers; eventually what you wrote will work but it isn't integrated into the cxx::bridge macro yet"))`

Jouzo (approved):

PR #2924, opened 2024-05-28
Static pool reward

Bushstar:
## Summary

The CalculateOwnerRewards function is currently the most performance intensive part of DeFiChain, this function calculates the coinbase, loan, commission and custom pool rewards for a liquidity owner from when their address was last used to the current usage which triggers this call. When calculating rewards, every single block since between the last usage and now is checked with the coinbase, loan, commission and custom pool rewards calculated for the address' liquidity against the total liquidity. This is heavy, especially if the address has not been used for a long time and has liquidity in multiple pools.

This PR greatly simplifies this process by avoiding working through each block a time. On each block we workout the coinbase, loan, commission and custom pool reward for each liquidity share in the pool and save this against a running total stored with the current height. When an address is used we get this total from the last time the address was used and the current total, deducting the last usage total from the current we can see the reward per liquidity share for this period and use that to reward the address for however many liquidity shares it has.

Rather than work through each block and run calculations for each reward type, we get the two totals from the last usage block to the current block for each reward type and calculate the rewards. There is additional work storing these totals on each block, but this can be considered negligible and a reasonable trade off.

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None

### DIFF
--- a/src/dfi/poolpairs.cpp
+++ b/src/dfi/poolpairs.cpp
@@ -253,8 +253,10 @@ void CPoolPairView::CalculateStaticPoolRewards(std::function<CAmount()> onLiquid
     // Get start and end reward per share
     TotalRewardPerShareKey key{beginHeight, poolID};
     const auto startSharePerReward = GetTotalRewardPerShare(key);
+    const auto startSharePerLoanReward = GetTotalLoanRewardPerShare(key);
     key.height = endHeight - 1;
     const auto endSharePerReward = GetTotalRewardPerShare(key);
+    const auto endSharePerLoanReward = GetTotalLoanRewardPerShare(key);
 
     // Get owner's liquidity
     const auto liquidity = onLiquidity();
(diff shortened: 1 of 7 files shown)


src/dfi/poolpairs.cpp
```diff
@@ -241,6 +241,50 @@ auto InitPoolVars(CPoolPairView &view, PoolHeightKey poolKey, uint32_t end) {
...
+        if (const auto rewardPerShare = end - start; rewardPerShare > 0) {
+            // Calculate reward
+            const auto reward = (liquidity * rewardPerShare / HIGH_PRECISION_SCALER).GetLow64();
+            // Pay reward to the owner
+            onReward(type, {DCT_ID{id}, static_cast<CAmount>(reward)}, endHeight);
+        }
+    };
+
+    calcReward(RewardType::Coinbase, startCoinbase, endCoinbase, {});
+    calcReward(RewardType::LoanTokenDEXReward, startLoan, endLoan, {});
```
kuegi, line 279: better to explicitly give DFI_ID(0) than {} ? IMHO such defaults make it harder to spot errors.

src/dfi/poolpairs.cpp
```diff
@@ -241,6 +241,50 @@ auto InitPoolVars(CPoolPairView &view, PoolHeightKey poolKey, uint32_t end) {
...
+    calcReward(RewardType::Commission, startCommission.commissionA, endCommission.commissionA, endCommission.tokenA);
+    calcReward(RewardType::Commission, startCommission.commissionB, endCommission.commissionB, endCommission.tokenB);
+
+    for (const auto &[id, endCustom] : endCustom) {
```
kuegi, line 283: better not shadow endCustom directly. makes it really hard to read and easy to get errors in.

kuegi (commented):
really elegant solution. I like it!

prasannavl (approved):

PR #2926, opened 2024-06-04
Ocean list pool swap verbose api

canonbrother:
## Summary

- check swap type
- enable composite swap indexer
- fix list_pool_swap response convert from_amount base unit to logical unit (8 decimal)

🚫 ~blocked by https://github.com/DeFiCh/rust-bitcoin/pull/6~

## Test coverage

```
/poolpair.swap.controller.defid.ts
/poolpair.swap.verbose.controller.defid.ts
poolswap buy-sell indicator
    ✓ should get pool swap details (50892 ms)
    ✓ should get composite pool swap for 2 jumps (50138 ms)
    ✓ should get composite pool swap for 2 jumps scenario 2 (50080 ms)
    ✓ should get composite pool swap for 3 jumps (49514 ms)
    ✕ should get direct pool swap for composite swap (49471 ms) // <-- require manually get the poolId
```
Regarding `should get direct pool swap for composite swap`..
manually calculate the direct pool id is required
https://github.com/BirthdayResearch/jellyfishsdk/blob/cb1c46722c9495dcfffd1e9f8c6d9a684d4dc7c9/apps/whale-api/src/module.indexer/model/dftx/composite.swap.ts#L25

no json rpc client inject in indexer.. thinking to reuse `graph`.. fix it at another pr
https://github.com/BirthdayResearch/jellyfishsdk/blob/cb1c46722c9495dcfffd1e9f8c6d9a684d4dc7c9/apps/whale-api/src/module.indexer/model/dftx/pool.pair.path.mapping.ts#L29

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -20,8 +20,8 @@ vsdbsled = { git = "https://github.com/defich/vsdbsled.git" }
 ethereum = { git = "https://github.com/defich/ethereum.git" }
 evm = { git = "https://github.com/defich/evm.git" }
 evm-runtime = { git = "https://github.com/defich/evm.git" }
-bitcoin = { git = "https://github.com/defich/rust-bitcoin.git", branch = "canonbrother/defichain-network" }
-bitcoin-io = { git = "https://github.com/defich/rust-bitcoin.git", branch = "canonbrother/defichain-network" }
+bitcoin = { git = "https://github.com/defich/rust-bitcoin.git" }
+bitcoin-io = { git = "https://github.com/defich/rust-bitcoin.git" }
 
 [workspace.dependencies]
 
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -20,8 +20,8 @@ vsdbsled = { git = "https://github.com/defich/vsdbsled.git" }
 ethereum = { git = "https://github.com/defich/ethereum.git" }
 evm = { git = "https://github.com/defich/evm.git" }
 evm-runtime = { git = "https://github.com/defich/evm.git" }
-bitcoin = { git = "https://github.com/defich/rust-bitcoin.git" }
-bitcoin-io = { git = "https://github.com/defich/rust-bitcoin.git" }
+bitcoin = { git = "https://github.com/defich/rust-bitcoin.git", branch = "canonbrother/defichain-network" }
+bitcoin-io = { git = "https://github.com/defich/rust-bitcoin.git", branch = "canonbrother/defichain-network" }
 
 [workspace.dependencies]
 
--- a/lib/ain-ocean/src/api/pool_pair/mod.rs
+++ b/lib/ain-ocean/src/api/pool_pair/mod.rs
@@ -1,11 +1,13 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
+    str::FromStr,
 };
 
 use ain_macros::ocean_endpoint;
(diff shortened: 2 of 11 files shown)


lib/Cargo.toml
```diff
@@ -20,8 +20,8 @@ vsdbsled = { git = "https://github.com/defich/vsdbsled.git" }
 ethereum = { git = "https://github.com/defich/ethereum.git" }
 evm = { git = "https://github.com/defich/evm.git" }
 evm-runtime = { git = "https://github.com/defich/evm.git" }
-bitcoin = { git = "https://github.com/defich/rust-bitcoin.git" }
-bitcoin-io = { git = "https://github.com/defich/rust-bitcoin.git" }
+bitcoin = { git = "https://github.com/defich/rust-bitcoin.git", branch = "canonbrother/defichain-network" }
+bitcoin-io = { git = "https://github.com/defich/rust-bitcoin.git", branch = "canonbrother/defichain-network" }
```
canonbrother, line 24: revert to target main after https://github.com/DeFiCh/rust-bitcoin/pull/6 merged

lib/ain-ocean/src/api/pool_pair/mod.rs
```diff
@@ -468,11 +452,20 @@ async fn list_pool_swaps_verbose(
             Ok((k, _)) => k.0 == id,
             _ => true,
         })
-        .map(|item| {
+        .map(|item| async {
             let (_, swap) = item?;
-            Ok(PoolSwapVerboseResponse::from(swap))
+            let from_to =
+                find_swap_from_to(&ctx, swap.block.height, swap.txid, swap.txno.try_into()?)
```
Jouzo, line 458: This is expensive and should only be done if we haven't indexed the `TxResult::PoolSwap`.
Should make the `TxResult::PoolSwap` optional in the PoolSwap `Index` implementation

canonbrother: the poolswap index impl is interesting.. checking...

canonbrother: if its not mistaken.. only need to execute the `TxResult::PoolSwap` with ocean enable to reduce op cost
[38149e8](https://github.com/DeFiCh/ain/pull/2926/commits/38149e8b20e0a745dfa31a8c9c6b9be71b71d0c5)

canonbrother: https://github.com/DeFiCh/ain/blob/38149e8b20e0a745dfa31a8c9c6b9be71b71d0c5/lib/ain-ocean/src/indexer/pool.rs#L39-L41
another blocker for deployment.. fixing it at another PR

src/dfi/mn_checks.cpp
```diff
@@ -1155,7 +1155,8 @@ Res CPoolSwap::ExecuteSwap(CCustomCSView &view,
     result = swapAmountResult.nValue;
 
     // Send final swap amount Rust side for indexer
-    if (txInfo) {
+    bool isOceanEnabled = gArgs.GetBoolArg("-oceanarchive", false);
```
canonbrother, line 1158: due to circular dep caused.. hardcoded the default as false as similar with `static constexpr bool DEFAULT_OCEAN_ARCHIVE_ENABLED = false`

Jouzo: Sounds good to me, we'll resolve it correctly later

PR #2927, opened 2024-06-06
Add -ports=auto option to auto set P2P, RPC, Eth RPC and Websocket ports

Bushstar:
## Summary

- Remove gRPC options
- Print actual ports used to debug.log
- Add `-ports=auto` option to auto set ports
- Store used ports in ports.lock
- defi-cli automatically uses ports.lock file is present

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


src/logging.cpp
```diff
@@ -322,3 +324,69 @@ void BCLog::Logger::ShrinkDebugFile()
...
+        uint16_t port;
+        if (std::getline(iss, key, '=') && (iss >> port) && key == portTypeStr) {
+            return port;
+        }
+    }
+
+    return 0;
+}
+
+void PrintPortUsage(const AutoPort portType, const uint16_t portNumber)
```
prasannavl, line 367: Name this `SetPortToLockFile`?

src/logging.cpp
```diff
@@ -322,3 +324,69 @@ void BCLog::Logger::ShrinkDebugFile()
     else if (file != nullptr)
         fclose(file);
 }
+
+static std::string PrintAutoPort(const AutoPort type)
```
prasannavl, line 328: `GetAutoPortString` might be better as it doesn't actually print anything

PR #2937, opened 2024-06-12
[breaking] Simplify UpdateTip log entry

Bushstar:
## Summary

- Simplify the overly verbose UpdateTip log entry to the following.

```
UpdateTip: height=4062963 hash=ed6ddcc70f13e798afa14a4529162a077bc56c740980f2b1d92cbc71d3bdf792 progress=1.000000 tx=32104696 date='2024-06-12T09:50:51Z'
```

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -3708,18 +3708,14 @@ void static UpdateTip(const CBlockIndex *pindexNew, const CChainParams &chainPar
     auto currentTime = GetSystemTimeInSeconds();
     if (!warningMessages.empty() || !::ChainstateActive().IsInitialBlockDownload() || lastTipTime < currentTime - 20) {
         lastTipTime = currentTime;
-        LogPrintf(
-            "%s: new best=%s height=%d version=0x%08x log2_work=%.8g tx=%lu date='%s' progress=%f cache=%.1fMiB(%utxo)",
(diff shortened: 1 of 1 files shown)


Jouzo:
Should we use this PR to add EVM block height as well ?

Bushstar:
> Should we use this PR to add EVM block height as well ?

Sure. Feel free to use this PR.

prasannavl:
Note that this is small enough to reduce noise and makes logs more useful at the same time not break compatibility too much that's it's harder for using parsers related to bitcoin to adapt.    

Leaving the rationale for the breaking change: 
- Remove longstanding noise, that's not very useful, but has been inherited from BTC 
  - version: Not very useful for us to print on each line. 
  - cache: Not accurate as it isn't an accurate repr of any actual usage in our chain as we have several caches that's spread across many things. 
 - Moving height to the first parameter, as that's the most useful info that's currently sandwiched in between many things. 
 - Removing superfluous words and labeling hash.  
 
 Could use significant changes, but anything else would break log parsers completely.

prasannavl (approved):

prasannavl:
Further: https://github.com/DeFiCh/ain/pull/2939

PR #2938, opened 2024-06-12
Ocean: add error field in ApiError

canonbrother:
## Summary

- as per title

Fix:
```
should throw error while getting non-existent poolpair
should throw error when fromToken === toToken
should throw error while getting non-existent collateral token id
should throw error while getting non-existent scheme
should throw error while getting non-existent loan token id
should throw error while getting non-existent vault
should throw error while getting non-existent token
```
Note for NotFound error missing id
should throw error for invalid tokenId


lib/ain-ocean/src/error.rs
```diff
@@ -121,8 +124,10 @@ impl ApiError {
 impl IntoResponse for ApiError {
     fn into_response(self) -> Response {
         let status = self.status;
-        let reason = Json(self);
-        (status, reason).into_response()
+        let body = Json(json!({
+            "error": self.error
```
canonbrother, line 128: add error field
ref:
https://ocean.defichain.com/v0/mainnet/poolpairs/1

Jouzo (approved):

PR #2940, opened 2024-06-13
Ocean: fix `PoolCreationHeight` sorting & add `CompositeSwap` Aggregated

canonbrother:
## Summary

- re-fix `should show aggregated swaps for 24h and 30d` by sorted poolpair creation height
- add missing index composite swap aggregated


lib/ain-ocean/src/indexer/mod.rs
```diff
@@ -63,6 +63,8 @@ fn index_block_start(services: &Arc<Services>, block: &Block<Transaction>) -> Re
         })
         .collect::<Result<Vec<_>>>()?;
 
+    pool_pairs.sort_by(|a, b| b.creation_height.cmp(&a.creation_height));
```
Jouzo, line 66: Can use `SortOrder:Descending above instead

Jouzo (approved):

PR #2947, opened 2024-06-26
updated list_gov_proposals API

nagarajm22:
## Summary

- Governance API 

## RPCs

- `API` implementation

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None

### DIFF
--- a/lib/ain-ocean/src/api/governance.rs
+++ b/lib/ain-ocean/src/api/governance.rs
@@ -47,8 +47,8 @@ async fn list_gov_proposals(
         r#type: query.r#type,
         cycle: query.cycle,
     };
-    let proposals = ctx.client.list_gov_proposals(Some(opts)).await?;
-
+    let mut proposals = ctx.client.list_gov_proposals(Some(opts)).await?;
+    proposals.sort_by(|a, b| a.creation_height.cmp(&b.creation_height));
     Ok(ApiPagedResponse::of(proposals, size, |proposal| {
         proposal.proposal_id.to_string()
     }))
--- a/lib/ain-ocean/src/api/pool_pair/mod.rs
+++ b/lib/ain-ocean/src/api/pool_pair/mod.rs
@@ -19,6 +19,7 @@ use path::{
     SwapPathsResponse,
 };
 use petgraph::graphmap::UnGraphMap;
+use price::DexPriceResponse;
 use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
@@ -43,8 +44,6 @@ use crate::{
     storage::SortOrder,
     Result, TokenIdentifier,
 };
-
-use price::DexPriceResponse;
 
 pub mod path;
 pub mod price;
--- a/lib/ain-ocean/src/api/pool_pair/price.rs
+++ b/lib/ain-ocean/src/api/pool_pair/price.rs
@@ -1,10 +1,9 @@
-use serde::Serialize;
 use std::{collections::HashMap, sync::Arc};
 
 use defichain_rpc::json::token::TokenInfo;
+use serde::Serialize;
 
 use super::{path::get_best_path, AppContext};
-
 use crate::{
     api::{
         cache::{get_token_cached, list_token_cached},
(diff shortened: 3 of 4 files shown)


lib/ain-ocean/src/model/governance.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub proposal_id: String,
+    pub title: String,
+    pub context: String,
+    pub context_hash: String,
+    pub r#type: ProposalType,
+    pub status: ProposalStatus,
+    #[serde(
+        serialize_with = "serialize_amount",
+        skip_serializing_if = "should_skip"
+    )]
```
canonbrother, line 18: we don't need this.. return null will do

lib/ain-ocean/src/model/mod.rs
```diff
@@ -21,7 +22,9 @@ mod transaction_vin;
 mod transaction_vout;
 mod tx_result;
 mod vault_auction_batch_history;
+// use ain_dftx::governance;
```
canonbrother, line 25:
```suggestion
```

lib/ain-ocean/src/api/governance.rs
```diff
@@ -41,30 +42,36 @@ async fn list_gov_proposals(
...
         cycle: query.cycle,
     };
     let proposals = ctx.client.list_gov_proposals(Some(opts)).await?;
-
-    Ok(ApiPagedResponse::of(proposals, size, |proposal| {
-        proposal.proposal_id.to_string()
-    }))
+    let mut proposals_with_string_amount: Vec<ApiProposalInfo> =
+        proposals.into_iter().map(ApiProposalInfo::from).collect();
+    // proposals.sort_by(|a, b| a.creation_height.cmp(&b.creation_height));
```
canonbrother, line 55:
```suggestion
```

lib/ain-ocean/src/api/governance.rs
```diff
@@ -41,30 +42,36 @@ async fn list_gov_proposals(
...
+    proposals_with_string_amount.sort_by(|a, b| a.creation_height.cmp(&b.creation_height));
```
canonbrother, line 56: hmm.. is sort needed here??

canonbrother:
> ✕ should listProposalVotes with all masternodes and cycle  (query issue)
what does this mean??

lib/ain-ocean/src/model/governance.rs
nagarajm22: if I dont use `should_skip `, the amount params does not appear, so we need have `should_skip `.

lib/ain-ocean/src/api/governance.rs
nagarajm22: I ll remove it :)

lib/ain-ocean/src/api/governance.rs
nagarajm22: if I dont use sort, test case is fail bcoz, default from `rpc` is descending.

lib/ain-ocean/src/model/mod.rs
nagarajm22: I cannot use above `governance `, due to amount 1.23 to 1.230000 conversion and some params are skipped as mentioned above, I need to serialisation.

canonbrother:
> Solution is change `cycle:Option<u64>` to `cycle:Option<I64>`

yep.. this is the correct way.. can we update the type??

lib/ain-ocean/src/model/mod.rs
canonbrother: hmm.. i don't get it.. i think i just asked for removing unuse 🤔

lib/ain-ocean/src/model/governance.rs
canonbrother: fixed [a286aa1](https://github.com/DeFiCh/ain/pull/2947/commits/a286aa1abbbe83ee58c34792cc4f5d16a654e78a)

canonbrother: @nagarajm22 my fixed got erased......

nagarajm22: I took the latest and did changes in above serialisation :) can you undo it ?

canonbrother: fixed [b7879e7](https://github.com/DeFiCh/ain/pull/2947/commits/b7879e717a42b2069ad0c7ea9230d2c5c3199a83)

canonbrother (approved):

PR #2951, opened 2024-06-27
tests: Add TokenAmount abstraction

prasannavl:
## Summary

- Make token amount easier to use in tests and simplify the amount of noisy concatenated strings in tests that results in harder to follow tests and add higher impedance to tests

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None


Bushstar:
TokenAmount should also have a `def __eq__(self, other)` method.

prasannavl:
Done

PR #2981, opened 2024-07-24
Ocean: `checked_div` default value ONE

canonbrother:
## Summary

- follow BigNumber `div` and `dividedBy` default value as ONE
- COIN refactor
- mild refine


lib/ain-ocean/src/indexer/oracle.rs
```diff
@@ -472,11 +471,10 @@ impl Index for SetOracleData {
                 }
             }
 
-            let result = total
+            let amount = total
                 .checked_div(Decimal::from(weightage))
-                .ok_or_else(|| Error::UnderflowError)?;
+                .unwrap_or(dec!(1));
```
canonbrother, line 476: key changes here

PR #2982, opened 2024-07-25
Add interruptblock RPC

Jouzo:
## Summary

- Introduces `interruptblock` RPC, which takes a block height as only param.
- If a node is running without `-interrupt-block` startup flag, sets `fInterrupt` and `fInterruptBlockHeight`.
- Can be run multiple times in a row and will update `fInterruptBlockHeight` to the latest height passed to `interruptblock` RPC.

This let's a user easily run a command against a target block height.
Example usage :

```
for height in `5000 10000 15000`
   defi-cli interruptblock height
   wait_for_block height
   ... Run any command at  required height.
```

Idea behind it is to ease diff log generation for full sync CI or Ocean A/B testing.
Current node behaviour requires multiple restart (as interrupt-block height can only be set via startup flag).
This also lets one interrupt an already running node.

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -414,6 +414,8 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "eth_getTransactionByBlockNumberAndIndex", 0, "tag"},
     { "eth_getTransactionByBlockNumberAndIndex", 1, "txIndex"},
     { "debug_feeEstimate", 0, "tx"},
+
+    { "interruptblock", 0, "height" },
 };
 
 /**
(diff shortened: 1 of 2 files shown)


prasannavl (approved):

prasannavl:
This region is un-synchronized. Would put this under the lock too. 

```
    fInterruptBlockHeight = height;
    fInterrupt = true;
```

prasannavl:
Actually most of this needs the lock until the last activate step. 

I'd actually just simplify it and take the lock as a whole than attempting to be granular here.

PR #2997, opened 2024-08-13
docs: use `<>` notation for MD email

DemoYeti:


prasannavl:
Hey, thanks for the PR. Even though autolinks are a part of the GFM spec (https://github.github.com/gfm/), it might be unnecessary to constrain compatibility. If there's a larger need, please feel free to reopen. 

Closing this for now, as this is not a necessary change on it's own.

PR #3002, opened 2024-08-17
fix: ConsolidateRewards batch consolidation bugs

### LINKED ISSUE #3001: Concurrent Balance modification in ConsolidateRewards when all rewards consolidated
ConsolidateRewards expects the List to contain each address only once. If it occurs multiple times, the same address might be consolidated multiple times "at once" (on different threads) which can lead to rewards being added multiple times in error.

The method doesn't even need a List of owner:balance as input but should rather receive a set which in itself already guarantees the uniqueness.

This is most likely also the reason for the chain split a few weeks ago where a MN used consolidaterewards and had an address get the rewards added multiple times.

https://github.com/DeFiCh/ain/blob/cd46ef77a88fa7403e0aa5f25e7170a0fea86828/src/init.cpp#L2207-L2215

kuegi:
## Summary

Fix for issue #3001 , ConsolidateRewards now uses set of CScript. It doesn't need/use the CAmount anyway.

If node was started with `-consolidaterewards` before, it needs a reindex or clean snapshot, as this likely lead to wrong balances before.

## Implications

- Storage
  - [x] Database reindex optional
  - [ ] None

- Consensus
  - [X] None


prasannavl:
Neat solution and clean-up. LGTM.

prasannavl (approved):

prasannavl:
One minor suggestion would be limit the scope of the set to `ConsolidateRewards` to avoid multiple sources of truth that can become out of sync if someone extends the code at later point etc.

Create a temp set just before ConsolidateRewards and drop it right after (scoped), or just creating a set out of balancesToMigrate inside ConsolidateRewards without changing the signature, might both accomplish it.

kuegi:
> One minor suggestion would be limit the scope of the set to `ConsolidateRewards` to avoid multiple sources of truth that can become out of sync if someone extends the code at later point etc.
> 
> Create a temp set just before ConsolidateRewards and drop it right after (scoped), or just creating a set out of balancesToMigrate inside ConsolidateRewards without changing the signature, might both accomplish it (both places where size hint is available to preallocate as well)

Can see if a scope is feasible in the few cases where its not single source of truth already. 
Change of signature is on purpose. Wrong signatures (unused argument data, mismatching container for purpose of function...) are a bad code smell cause it leads to exactly such errors.

prasannavl:
Thanks @kuegi. 

> Change of signature is on purpose. Wrong signatures (unused argument data, mismatching container for purpose of function...) are a bad code smell cause it leads to exactly such errors

The C++ idiomatic way to do this would be not to use the concrete container type in the signature at all and instead pass the iterator types begin, end and a size hint, and only use that on ConsolidateRewards, so vector or set doesn't matter. 

At this point here, I think we just want to get this in quickly, so OK with either approach. 

> Can see if a scope is feasible in the few cases where its not single source of truth already

Thanks, works - this is what I had in mind: https://github.com/DeFiCh/ain/compare/pvl/pr-3002?expand=1#diff-872ba0cbdc244b4c2d8d90afb30f4842e386ff9a22ae22c90875a8a4327ec5e9R1727-R1738
  
That said: the reason for the hold up of this PR: it could change ordering. Since a set is used (std::set is RB tree, which keeps the sorted form). Need more testing to ensure there aren't edge case where this changes the ordering. 

A quicker workaround would just be use unsorted set.

prasannavl:
Here's my branch with it switched to unordered set: https://github.com/DeFiCh/ain/commit/9bd2d25d485a9c85f472d69cc509de7890f429b2

In addition, should also fix another bug if it's consolidate rewards is called with both empty args and tokens. 

Verifying full sync. Will merge it back into this PR once ready.

kuegi:
> The C++ idiomatic way to do this would be not to use the concrete container type in the signature at all and instead pass the iterator types begin, end and a size hint, and only use that on ConsolidateRewards, so vector or set doesn't matter.

But that again would loose the uniqueness. It must not be a vector, but a set. If the arguments allow any container, the method internally must ensure uniqueness to prevent the bug.

prasannavl:
> But that again would loose the uniqueness. It must not be a vector, but a set. If the arguments allow any container, the method internally must ensure uniqueness to prevent the bug.

Yes, but a container is just a layout. It's not the responsibility of the consolidate rewards method to ensure it's unique in this case. If it is, then we build the unique list inside `ConsolidateRewards`. 

> or just creating a set out of balancesToMigrate inside ConsolidateRewards

If it's not, then it's the responsibility of the outer method, in which case, ConsolidateRewards just iterates over (this can be a vector that already has the unique list). 

Anyway, unsorted_set should take care of it (which is internally, just a vector with a hash table to ensure O(1) is there check to insert).

prasannavl:
@kuegi could you merge my branch: https://github.com/DeFiCh/ain/compare/pvl/pr-3002 into this if you get a chance? I'm off my desk a bit and don't see an active branch here to push into based off the web api.

kuegi:
> Thanks, works - this is what I had in mind: https://github.com/DeFiCh/ain/compare/pvl/pr-3002?expand=1#diff-872ba0cbdc244b4c2d8d90afb30f4842e386ff9a22ae22c90875a8a4327ec5e9R1727-R1738

Why did you change the scoping compared to my version? Your version needs an extra loop and has totalAccounts outside, which is actually not used anywhere else. Any reason for that?

prasannavl:
> Why did you change the scoping compared to my version? Your version needs an extra loop and has totalAccounts outside, which is actually not used anywhere else. Any reason for that?

Extra loop should have no significance here as it's fully contained in-memory. In fact, I'd expect this to be very slighly micro-faster, since it pre-allocates the memory for the container that's needed due to known size.

prasannavl:
The reason to do it here, is to scope it as tightly as possible. So, it's a container set that's made for consolidaterewards and discarded right now, so it doesn't accidentally leak anywhere else and the source of truth still just remains as balancesToMigrate.

prasannavl:
I might have overlooked the `totalAccounts` part. Let me look through - might be bug that needs to be fixed

prasannavl:
Don't think this changed the behavior of `totalAccounts` in anyway here. Not sure what you mean on it's impact / significance of it in scope. 

That said - I'd probably make a few more changes / cleanup, as this could be changing behavior: 

https://github.com/DeFiCh/ain/pull/3002/files#diff-872ba0cbdc244b4c2d8d90afb30f4842e386ff9a22ae22c90875a8a4327ec5e9R1736-R1741

This sorts things and previous it was done pre-sort (and the set ended up doing the sort again). I have to clean this up a bit more.

prasannavl:
Just leaving a note for ref until fixed: the current state cannot be merged. As it has completely different sorting behavior. set with RB tree will sort based on `CScript` while the previous behavior was based on the balance. 

Fix: 
- Use the unsorted_set *after* the sorting step.

kuegi:
> Don't think this changed the behavior of `totalAccounts` in anyway here. Not sure what you mean on it's impact / significance of it in scope.

I just realized that totalAccounts is only used for ConsolidateRewards which is why I added it to the scope in my scope implementation that I pushed a week ago. You completely ignored my implementation, and I wondered why.

kuegi:
> Just leaving a note for ref until fixed: the current state cannot be merged. As it has completely different sorting behavior. set with RB tree will sort based on `CScript` while the previous behavior was based on the balance.
> 
> Fix:
> 
> * Use the unsorted_set _after_ the sorting step.

You mean in the poolSplit case? IMHO ordering should be completely irrelevant (as its done multithreaded anyway, so ordering is "random" in the execution anyway), but can easily be fixed if important for you.

prasannavl:
> You mean in the poolSplit case? IMHO ordering should be completely irrelevant (as its done multithreaded anyway, so ordering is "random" in the execution anyway), but can easily be fixed if important for you.

Yes, agree, ordering in the "ideal" case should be irrelevant. But trying to stay close to the original implementation here, so that if there were any other bugs that we haven't discovered, it still matches the existing implementation and doesn't deviate from it.

prasannavl:
It's all in random threads as well. So, in theory there should be no chance. Just trying to be extra cautious here and play it safe to reduce the amount of things to be mindful of, since we have a lot of things coming from the current update.

Bushstar (approved):

prasannavl:
Thanks for this @kuegi. Merging.

PR #3003, opened 2024-08-19
use `match` rather than `if`/`else`

DemoYeti:
This PR fixes the styling of the use to use a `match` statement rather than an `if`/`else` statement


prasannavl:
Sorry, closing this. Contributions are welcome when they add value.

Please try not to add commits that increasingly seem to be namesake than actually adding value.  Related: #2997

PR #3039, opened 2024-09-06
fix: revert wrong refactor of GetLoanTokensForFutures

kuegi:
## Summary

During the restart PR, the method `GetLoanTokensForFutures` was copied to `GetLoanTokensForLocks`. This method was later simplified cause the "old way" of getting loan tokens is not needed for the restart case. With this refactor, the GetLoanTokensForFutures was simplified the same way which leads to errors in FutureSwaps before the loanTokens moved to attributes.

This PR fixes this by reverting the change in the GetLoanTokensForFutures method.

nodes with existing data on changi should be fine with the fork. I guess this fix is only needed to make the full reindex work again.

## Implications

- Storage
  - [X] None

- Consensus
  - [X] None


prasannavl:
Merging to set a pre-release for changi candidates to update to in-time. Release to be set after https://github.com/DeFiCh/ain/actions/runs/10747073649 passes.

PR #3050, opened 2024-09-16
Restart: Check pools valid

Bushstar:
## Summary

- Before adding EVM and creation TXs miner side for the dToken restart make sure that all loan and collateral token pools are not disabled.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -216,6 +216,7 @@ static void AddTokenRestartTxs(BlockContext &blockCtx,
     };
 
     std::set<uint32_t> loanTokenIds;
+    std::set<uint32_t> collateralTokenIds;
 
     attributes->ForEach(
         [&](const CDataStructureV0 &attr, const CAttributeValue &) {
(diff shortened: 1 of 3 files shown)


src/miner.cpp
```diff
@@ -238,7 +240,19 @@ static void AddTokenRestartTxs(BlockContext &blockCtx,
 
     const auto tokensLocked = mnview.AreTokensLocked(loanTokenIds);
 
-    if (!tokenPricesValid || tokensLocked) {
+    bool poolDisabled{false};
+    mnview.ForEachPoolPair([&](DCT_ID const &poolId, const CPoolPair &pool) {
+        if (loanTokenIds.count(pool.idTokenA.v) || loanTokenIds.count(pool.idTokenB.v) ||
+            collateralTokenIds.count(pool.idTokenA.v) || collateralTokenIds.count(pool.idTokenB.v)) {
+            if (!pool.status) {
```
kuegi, line 247: does this lead to issue with old dToken/v1-DUSD pools? cause they all include DUSD, so come up in this list, but its fine that they are disabled.
only dToken-DUSD pools, gateways and collateral-DFI pools are "needed"

PR #3062, opened 2024-09-23
feat: unfreeze mn

canonbrother:
## Summary

- Adds in the option to enable unfreeze MN should the DFIP https://defiscan.live/governance/d9bd72090e6d5258ec2372049ae6758dc319deba76c6f49850ec5e80ae3823db pass.
- Note that this doesn't enforce it - just simply adds it so the community can decide and has the option to.


prasannavl:
Thanks, simple and straight forward enough. Merging. 

Community can make the call on if they want to enable or not as needed later.

PR #3063, opened 2024-09-23
Add unsetgovheight/cleargovheights. Remove Gov from setgov/unsetgov.

Bushstar:
## Summary

- Remove Governance access to `setgov` and `unsetgov`.
- Add `unsetgovheight` which will unset a Gov var at the set height. Governance has access to this feature.
- Add a minimum block Gov var that will be used to check that the height on `setgovheight` and `unsetgovheight` is set higher than the current height plus that new minimum block height.
- `cleargovheights` will clear all pending `setgovheight` and `unsetgovheight` changes.

## Implications

- Storage
  - [ ] None

- Consensus
  - [ ] None


src/dfi/consensus/governance.cpp
```diff
@@ -9,8 +9,7 @@
 
 Res CGovernanceConsensus::operator()(const CGovernanceMessage &obj) const {
     // Check foundation auth
-    auto authCheck = AuthManager(blockCtx, txCtx);
-    if (auto res = authCheck.HasGovOrFoundationAuth(); !res) {
+    if (auto res = HasFoundationAuth(); !res) {
```
kuegi, line 12: why is this foundation only, and not govOrFoundation?

src/dfi/consensus/governance.cpp
```diff
@@ -106,7 +101,39 @@ Res CGovernanceConsensus::operator()(const CGovernanceMessage &obj) const {
...
+Res CGovernanceConsensus::operator()(const CGovernanceClearHeightMessage &obj) const {
+    // Check auth
+    auto authCheck = AuthManager(blockCtx, txCtx);
+    if (auto res = authCheck.HasGovOrFoundationAuth(); !res) {
+        return res;
+    }
+
+    auto &mnview = blockCtx.GetView();
+
+    auto heightToClear = [](const auto &collection) {
```
kuegi, line 113: shouldn'T this be "varsToClear"? this has nothing to do with height, right?

src/dfi/consensus/governance.cpp
```diff
@@ -106,7 +101,39 @@ Res CGovernanceConsensus::operator()(const CGovernanceMessage &obj) const {
...
+
+    const auto unsetVars = mnview.GetAllUnsetStoredVariables();
+    for (const auto &height : heightToClear(unsetVars)) {
+        mnview.EraseUnsetStoredVariables(height);
+    }
+
+    return Res::Ok();
+}
+
+Res CGovernanceConsensus::operator()(const CGovernanceUnsetHeightMessage &obj) const {
```
kuegi, line 136: should this have a DF24 check?

src/dfi/consensus/governance.cpp
```diff
@@ -138,6 +176,37 @@ Res CGovernanceConsensus::operator()(const CGovernanceUnsetMessage &obj) const {
         if (!res) {
             return Res::Err("%s: %s", name, res.msg);
         }
+    }
+
+    // Store pending Gov var changes
+    return StoreUnsetGovVars(obj, mnview);
+}
+
+Res CGovernanceConsensus::operator()(const CGovernanceUnsetMessage &obj) const {
```
kuegi, line 185: should have a DF24 check to no longer allow it afterwards, right?

src/dfi/consensus/governance.cpp
prasannavl: https://github.com/DeFiCh/ain/pull/3063/files#diff-2e2980eda773148aea9f732c9f165bb7a9427b4f9df3c49b78b16d9166d58c42R289-R293

src/dfi/consensus/governance.cpp
prasannavl: Same as the other. So, this should be ok

https://github.com/DeFiCh/ain/pull/3063/files#diff-2e2980eda773148aea9f732c9f165bb7a9427b4f9df3c49b78b16d9166d58c42R289-R293

src/dfi/consensus/governance.cpp
kuegi: ok thx. means the gates in other similar methods are actually unnecessary?

prasannavl: yes, if it's only only used for simple gating tied to network upgrades.

Issue #3098, opened 2024-10-16
segmentation fault : rocksdb : Too many open files


skutcher:
## Summary

- After my DB got destroyed by heavy load again, i tried to start with a snapshot as always. But without success. Every time got the same error (see below)

## Steps to Reproduce

- just to have a clean directory (the behavior is the same with the old fully configured .defi folder)
- mv .defi .defi_backup
- mkdir -p .defi/snapshot
- cd  .defi/snapshot 
-  wget https://defi-snapshots-europe.s3.eu-central-1.amazonaws.com/snapshot-mainnet-4417683.zip (i tried the last 3 so far)
- unzip
-  mv indexes evm blocks history burn anchors chainstate enhancedcs ../
-  start defid

## Environment

- Node Version: 4.2.0
- Block height on bug if applicable: 
- TX or TX type on bug if applicable:  
- OS with version: archlinux with 6.11.3-arch1-1 on btrfs

## RUST_BACKTRACE=full
```
2024-10-16T18:59:04Z [ain_grpc] Starting JSON Eth RPC server at 127.0.0.1:8551
2024-10-16T18:59:04Z [ain_grpc] Init rs ws services
2024-10-16T18:59:04Z [ain_grpc] Starting ws server at 127.0.0.1:8553
2024-10-16T18:59:04Z init message: Done loading
2024-10-16T18:59:04Z 0 addresses found from DNS seeds
2024-10-16T18:59:04Z dnsseed thread exit
2024-10-16T18:59:04Z New outbound peer connected: version: 70038, blocks=4426259, peer=0
2024-10-16T18:59:05Z New outbound peer connected: version: 70038, blocks=4426259, peer=1
2024-10-16T18:59:05Z Synchronizing blockheaders, height: 4419682 (~99.85%)
2024-10-16T18:59:05Z New outbound peer connected: version: 70038, blocks=4426259, peer=2
2024-10-16T18:59:06Z Synchronizing blockheaders, height: 4421682 (~99.90%)
2024-10-16T18:59:06Z [vsdb_core::common::engines::rocks_db] rocksdb write_buffer_size per column family = 512MB
thread '<unnamed>' panicked at /github/home/.cargo/git/checkouts/vsdb-80d7017a79ae04c0/020aaf6/core/src/common/engines/rocks_db.rs:31:67:
[...]

PR #3104, opened 2024-10-25
Update rocksdb default options

Jouzo:


prasannavl:
Should also resolve https://github.com/DeFiCh/ain/issues/3098

PR #3105, opened 2024-10-28
Ocean: Catchup to tip on startup

Jouzo:
## Summary

- Ocean tries to catch up to tip height on startup if ocean height is below tip height.
- Let's one sync ocean from snapshot


src/ocean.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+#include <chrono>
+#include <cstdint>
+#include <string>
+
+bool OceanIndex(const UniValue b, uint32_t blockHeight) {
+    CrossBoundaryResult result;
+    ocean_index_block(result, b.write());
+    if (!result.ok) {
+        LogPrintf("Error indexing ocean block %d: %s\n", result.reason, blockHeight);
```
Bushstar, line 23: `blockHeight` and `result.reason` are the wrong way round.

src/ocean.cpp
```diff
@@ -0,0 +1,139 @@
...
+        ocean_invalidate_block(result, b.write());
+        if (!result.ok) {
+            LogPrintf("Error invalidating ocean %d block: %s\n", result.reason, blockHeight);
```
Bushstar, line 26: Same here with print args being the wrong way round.

Bushstar (approved):

PR #3109, opened 2024-11-04
Log DVM hash without undos

Bushstar:
## Summary

- Hashes DVM DB without undos which were changed from the 4.1.0 release onwards. This PR then allows us to compare nodes using the hash without undos which should match regardless of whether they were synced with 4.1.0+ or not.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/dfi/rpc_accounts.cpp
+++ b/src/dfi/rpc_accounts.cpp
@@ -3700,8 +3700,9 @@ UniValue logdbhashes(const JSONRPCRequest &request) {
     // Create a CDBIterator
     auto pcursor = db->NewIterator(leveldb::ReadOptions());
 
-    // Create a SHA256 hasher
+    // Create SHA256 hashers
     CSHA256 hasher;
+    CSHA256 hasherNoUndo;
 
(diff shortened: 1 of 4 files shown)


src/dfi/rpc_accounts.cpp
```diff
@@ -3712,9 +3713,16 @@ UniValue logdbhashes(const JSONRPCRequest &request) {
         auto keySlice = pcursor->GetKey();
         auto valueSlice = pcursor->GetValue();
 
-        // Feed the key and value into the hasher
-        hasher.Write((const unsigned char *)keySlice.data(), keySlice.size());
-        hasher.Write((const unsigned char *)valueSlice.data(), valueSlice.size());
+        const auto key = std::string(keySlice.data(), keySlice.size());
+        if (!key.empty() && key[0] == 'u') {
```
prasannavl, line 3717: The second branch can be turned branchless if the condition is flipped 

`key[0] != 'u'` { hashNoUndo.write }

prasannavl (approved):

PR #3110, opened 2024-11-05
Fix skipping of static rewards

Bushstar:
## Summary

- Fix skipping of static rewards when consolidating rewards.
- Flush DB after consolidating rewards.
- Fix consolidate rewards test.
- Always set skip static when consolidating rewards, we never want them.
- Remove passing of workers as this is always from RewardConsolidationWorkersCount() which is called internally anyway.

## Implications

- Storage
  - [x] None

- Consensus
  - [x] None

### DIFF
--- a/src/validation.h
+++ b/src/validation.h
@@ -936,7 +936,6 @@ void ConsolidateRewards(CCustomCSView &view,
                         int height,
                         const std::unordered_set<CScript, CScriptHasher> &owners,
                         bool interruptOnShutdown,
-                        int numWorkers = 0,
                         bool skipStatic = false);
 
 extern std::map<CScript, CBalances> mapBurnAmounts;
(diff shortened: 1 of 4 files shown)


prasannavl (approved):

src/validation.h
```diff
@@ -936,7 +936,6 @@ void ConsolidateRewards(CCustomCSView &view,
                         int height,
                         const std::unordered_set<CScript, CScriptHasher> &owners,
                         bool interruptOnShutdown,
-                        int numWorkers = 0,
                         bool skipStatic = false);
```
kuegi, line 939: to prevent further default-value bugs and increase clarity: why not remove the default and have an explicit false on the calls?

prasannavl:
Missed the comment above. Agree that default values are best avoided - usually not worth the cost of difficult to catch errors.

